PR #4, opened 2011-03-08
PGP/Mime support in K-9

mpasch:
Hey guys,

I'm working on PGP/Mime support in K-9 based on APG for the tarent company.

Since we'd like to see the code upstream, I wanted to ask for your opinion about this "patch".
You probably don't want to accept this request at this point, because of bugs 1) and 2) (see below).
I still would like to know what you think, cause I won't have much time to spend on this later.

I use the current version of APG in the android market. I'm working with an HTC Desire Z. (non-root android 2.2.1)

What's working (both send and receive)
- encrypted mails, encrypted+signed mails (attachments are encrypted, too)

What's not working:
- send/receive signed-only mails (you can still read them, but the signature is not checked)

Known Bugs:

1) there is no warning that sending signed-only mails does not work (APG fails)

2) because the whole email including attachments is sent to APG as an byte array, sending attachments >~500Kb fails (without error message). This needs to be replaced with passing it over using an Uri (content://), I would like to use the K-9 content-provider here, do you think that might work? (Is it saving the data as tempfile?, is that a security issue?)

3) receiving attachments works with common attachment sizes you would except on a smartphone (so a 3-5MB file is no problem, my 16MB-testfile caused an endless loop)

4) the PGP/Mime specification requires the following header:

Content-Type: multipart/encrypted; boundary="...."; protocol="application/pgp-encrypted"

due to the way K-9 stores messages in the local store the protocol value gets lost (I think this is a known problem in k-9)
Thunderbird, K-9, and Kontact do not care about the missing attribute, I haven't tested any other clients, so I hope this won't be a problem. (I think fixing this would require massive changes in K-9)

I think 5) and 6) are not related to my patch, if they are please let me know.

5) if you're sending an attachment using OIFileManager, the full path is used as filename

6) PGP/Inline sending does not work if you have HTML enabled, I think this is tracked in Issue 3051 

Thank you very much for your thoughts!

I hope we can get PGP/Mime to work in K-9!!

Mattis Pasch


Diapolo:
What can be done to get this forward? We really need PGP/SMIME support for K9 to become perfectly usable!

dschuermann:
@Diapolo work on https://github.com/k9mail/k-9/milestones/PGP/MIME

PR #530, opened 2015-01-12
Pgp/mime

hagenduk:
current merge


hagenduk:
this schould be intern, my github just didnt change the path correctly, sry about that

PR #533, opened 2015-01-23
Fix a typo in Polish translation.

karolba:
`settings_export_account` is used in the menu that can be activated by long-pressing one account on the main screen. Without this commit, there are two actions named `Ustawienia konta` (`Account settings`), but one of them is actually used to export the settings.


cketti:
We use Transifex for translations. Please fix the bug there: https://www.transifex.com/projects/p/k9mail/language/pl/

karolba:
@cketti 
Today I have received a K9 update on F-Droid from 5.002 to 5.004. I have fixed this bug on Transifex when you told me to do so, but the it's still present in the new version. Can you change it manually? 

This is what I'm talking about

PR #538, opened 2015-01-29
Improve UI for switching on/off PGP signing and encryption

ligi:
The Idea is to use the same design language as is used when displaying messages so users get to know what these icons mean.
Also we will add a settings thingy next to it as the next step.
Also cleaned up stuff around legacy PgpData - should be removed completely as @dschuermann suggested - but let's do it in small steps

output_optimized


ligi:
as agreed with @cketti will make it with different layout in new PR

PR #556, opened 2015-02-17
PgpData unit tests

vcaulier:
Beginning the crypto package tests.
And starting with PgpData tests (seen as uncovered with Jenkins).

However, I don't know how to integrate the test to the gradle build,
will do it next.

### DIFF
--- a/k9mail/src/androidTest/java/com/fsck/k9/crypto/PgpDataTest.java
+++ b/k9mail/src/androidTest/java/com/fsck/k9/crypto/PgpDataTest.java
@@ -1,0 +1,166 @@
+package com.fsck.k9.crypto;
+
+import android.support.test.runner.AndroidJUnit4;
+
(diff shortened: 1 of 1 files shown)


artbristol:
PgpData is just a DTO. IMHO, a unit "test" is just overhead.

The only method that is vaguely worth testing is hasEncryptionKeys

dschuermann:
Sry that you made this effort, but PgpData will be removed in future versions. It already is partially gone in the pgp preparations branch.

vcaulier:
Ok, but at least, its a way to better understand the code.
Do you have any advice on which classes should be tested ?

dschuermann:
If you want to create tests for PGP, please contact cketti on the mailinglist and look into https://github.com/k9mail/k-9/tree/pgp_mime_preparations

vcaulier:
Ok, thank you; I will look for that.

PR #558, opened 2015-02-18
IMAP Keywords (rebased)

igoralmeida:
This is a rebase of https://github.com/k9mail/k-9/pull/146 on top of current master, also adding the ability to list the tags in the message view.
I have no intention of merging this as-is, but hopefully this is a good starting point to discuss the functionality.

For example: the actual listing should use a mapping so that the user sees _tags_, not the IMAP keywords themselves ($label1, $label2, some_keyword, ...), but I don't have enough android experience to do this yet.

Please comment.

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
@@ -13,7 +13,7 @@ package com.fsck.k9.mail;
  * attached internally. When using the flags with external servers,..
  * one should use the realName() method.
  */
-public class Flag {
+public final class Flag {
 
     /*
      * IMPORTANT WARNING!!
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
@@ -16,6 +16,12 @@ package com.fsck.k9.mail;
 public class Flag {
 	
 	/*
+	 * IMPORTANT!!
+	 * 
+	 * INTERNAL NAME MUST EQUAL THE NAME OF THE FIELD!!!
+	 */
+	
+	/*
 	 * IMAP Spec flags.
 	 */
     public static final Flag DELETED = new Flag("DELETED", "\\Deleted");
@@ -76,49 +82,62 @@ public class Flag {
      */
     private static final String USER_PREFIX = "USER_";
     
-    private final String name;
-    private final String internal_name;
+    private final String name;				// for use towards third party 	ex. "\\Deleted"
+    private final String internal_name;		// for internal use in database,...   ex. "DELETED"
     protected boolean bCustom;
     
     /**
      * When a Flag is created dynamically we know it's a custom flag.
-     * @param name Name of the flag.
+     * 
+     * @param name Internal name of the flag.
      * @return Newly created Flag object.
      */
-    public static Flag CreateFlag(String name) {
-    	Flag tmpFlag = new Flag(USER_PREFIX+name, name);
+    public static Flag CreateFlag(String internal_name) {
+    	Flag tmpFlag = new Flag(USER_PREFIX+internal_name, internal_name);
     	tmpFlag.bCustom = true;
     	return tmpFlag;
 	}
     
-    private Flag(String name){
-    	this(name, name);
-    }
-    
-    private Flag(String internal_name, String name){
-    	this.name = name;
-    	this.bCustom = true;
-    	this.internal_name = internal_name;
+    private Flag(String internal_name){
+    	this(internal_name, internal_name);
     }
     
     /**
+     * Create a new Flag. This doesn't create a custom flag, it's used to define
+     * the predefined flags.
+     * 
+     * @param internal_name Name for internal use in database,...   ex. "DELETED"
+     * @param name Name for use towards third party ( ex. "\\Deleted" )
+     */
+    private Flag(String internal_name, String name){
+    	this.name = name;
+    	this.bCustom = false;
+    	this.internal_name = internal_name;
+    }
+    
+    /** 
      * Returns the predefined static flag object if any. Otherwise a new
      * custom flag is created and returned.
      * 
-     * TODO We might not want to create a custom flag by default and just
-     * throw an exception. Only 10 uses of this method so refactoring possible.
-     * Catching the exception should then proceed with a CreateFlag(name) call.
+     * When the name starts with the USER_PREFIX we don't add it again to
+     * create a new one. This is the case for example when this is called with
+     * strings retrieved from the database.
+     * 
+     * IMPORTANT remember the name of the field of predefined flags must equal the
+     * internal name!
      * 
      * @param name Name of Flag wanted.
      * @return	Predefined Flag object if any otherwise new custom Flag.
      * @throws IllegalArgumentException Thrown when the field is not accessible.
      */
-    public static Flag valueOf(String name) throws IllegalArgumentException{
+    public static Flag valueOf(String internal_name) throws IllegalArgumentException{
     	try {
-			return (Flag)(Flag.class.getField(name).get(null));
+			return (Flag)(Flag.class.getField(internal_name).get(null));
 		} catch (NoSuchFieldException e) {
 			// not a predefined flag
-			return Flag.CreateFlag(name);
+			if(internal_name.startsWith(USER_PREFIX))
+				return Flag.CreateFlag(internal_name.substring(USER_PREFIX.length()));
+			else return Flag.CreateFlag(internal_name);
 		} catch (IllegalAccessException e) {
 			throw new IllegalArgumentException(e);
 		}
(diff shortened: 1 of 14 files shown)


k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
```diff
@@ -19,40 +41,183 @@
...
+     * internal name!
+     *
+     * @param mName Name of Flag wanted.
+     * @return  Predefined Flag object if any otherwise new custom Flag.
+     * @throws IllegalArgumentException Thrown when the field is not accessible.
+     */
+    public static Flag valueOf(String internalName) throws IllegalArgumentException {
+        try {
+            // the argument being 'null' implies we look for a static field
+            return (Flag)(Flag.class.getField(internalName).get(null));
```
artbristol: Rather than reflection here, I would extract the internal flags listed in valueOfByRealName into a method returning a set of Strings and see if it contains internalName.

You can ensure the new method is kept in sync with the actual fields in a unit test.The unit test can use reflection to enumerate Flag's fields without concern about performance.

k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
```diff
@@ -3,15 +3,37 @@
...
  */
-public enum Flag {
-    DELETED,
-    SEEN,
-    ANSWERED,
-    FLAGGED,
-    DRAFT,
-    RECENT,
-    FORWARDED,
+public class Flag {
```
artbristol: I guess this class should be final to avoid equals/hashcode problems

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapStore.java
```diff
@@ -1979,19 +1966,9 @@ public void expunge() throws MessagingException {
...
-                    flagNames.add("\\Answered");
-                } else if (flag == Flag.FLAGGED) {
-                    flagNames.add("\\Flagged");
-                } else if (flag == Flag.FORWARDED
-                        && (mCanCreateKeywords || mPermanentFlagsIndex.contains(Flag.FORWARDED))) {
-                    flagNames.add("$Forwarded");
-                }
-
+            	// client's can't add the RECENT flag!
+            	if( !flag.equals(Flag.RECENT) )
```
artbristol: curly brackets avoid goto fail :-)

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapStore.java
```diff
@@ -2113,66 +2090,15 @@ protected String getLogId() {
                 public List<ImapResponse> search() throws IOException, MessagingException {
                     String imapQuery = "UID SEARCH ";
                     if (requiredFlags != null) {
+                        /* TODO not sure of this: should probably use Flag.mName instead */
```
artbristol: I agree!

k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
jberkel: i think this class could actually stay an enum, it's just a value object w/o any behaviour

jberkel: ah ok, didn't notice that you actually create new instances.

k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
igoralmeida: Something like k9mail/src/androidTest/java/com/fsck/k9/mailstore/LocalMessageTest.java ?
How would it account for new flags showing up during the imap session, though?

igoralmeida:
Guys, thanks for the feedback.

Could you also give me some pointers regarding the tag<->keyword mapping I mentioned? This is how alpine does, for example.
I'm guessing K-9 would have to query some sort of local database to find what tag to show when Flag.mName=='$label1'. How should this work here?

jberkel:
another important thing (IMHO) is naming. this PR mentions `flag`, `keyword`, `label`, etc... it's quite confusing. is there a good way to make the concepts more obvious?

igoralmeida:
Here's my take:

"flags" are hard-coded IMAP message attributes (SEEN, ANSWERED, RECENT, etc.)

"[user-defined] keywords" live on the IMAP server as well, but can be created by the IMAP client

"tags" are user-facing text that relate to IMAP keywords. I'm thinking a dictionary-like mapping {'$label1':'TODO', 'some_keyw':'Some Keyword'}

"labels" should be avoided here due to Gmail's implementation and UI, which uses IMAP folders instead of keywords. Also, Thunderbird will unfortunately use '$label1...5' for some keywords, but I believe that's understandable given the context.

Thus, the UI should use Tags consistently, while showing the term keyword only on some account configuration activity that maps Keyword to Tag + (possibly) Color.

johnpatcher:
One of the last features that I miss personally. A feature like this is mostly about the UI and how to present tags to the user in a useful way. Does anyone know of mobile mail clients that already support this? I like the way Thunderbird handles this (color and GUI "labels"), but I'm not sure whether this is applicable on mobile devices.

k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
artbristol: @igoralmeida  I've added an example of how you could do it here

https://github.com/artbristol/k-9/tree/art/keywords-swap-reflection-to-unit-test

igoralmeida:
Latest push: merged @artbristol 's consistency/test code + answer comments here.
I'm still 'losing' keywords, i.e. sometimes the Tags field won't show up because the USER_* keywords get lost from LocalMessage.mFlags. This happens somewhere between MessagingController.syncFlags() and MessageHeader.populate().

igoralmeida:
I could not reproduce the problem fully, but I'm almost sure it was due to removing some X_.... flags. If I understand correctly, those are supposed to be local to k-9 only.

Anyway, I believe it's fixed now. Whenever messages have user-defined keywords, they will show up in the header in the message view activity.

igoralmeida:
Some ramblings...

I've been looking into the keyword->tag mapping functionality and how things would look like in StoreSchemaDefinition and the fetching code.
Initially, I was considering adding three more tables (`imap_keywords`, `ui_tags` and `keyword_tag_map`), so that any keyword could be mapped to any tag, consistently. However, ensuring the mappings were consistent in the DB would require changing the `messages`.flags column as well, so I put that aside.

The next idea would be a single `keyword_and_tag` table, "(keyword TEXT UNIQUE, tag TEXT)", later possibly with a 'color' field too. In this case, I'd add the USER_xxxx keywords in lowercase as they came up during sync, and eventually the user could edit the mappings in a new activity.

Regardless of the design, `messages`.flags being a multivalued field makes it a bit cumbersome to simply extend the query and have the tags magically appear, so I'd probably have to query keyword_and_tag. So what am I looking for, something inside EmailProvider?

And, finally, another approach that occurred to me was adding the tags to LocalMessage directly, probably in LocalMessage.populateFromGetMessageCursor. In this case, I think I need to create some kind of "TagStore" similar to LocalStore so I can discover the tags as the keywords appear.

If anyone can explain the current architecture and/or comment on the approaches above, I'd be happy to read.

igoralmeida:
This "works" (read-only). Kind of ugly. Contributions, especially to the UI, are very welcome.

igoralmeida:
How it works: from the message header, tap "Tags" (or tap any of them):
message_header

The mapping activity shows up:
tagmapping

igoralmeida:
I think adventurous people can start poking now, despite the annoying usability problems:

1) Click the "Tags" label or any of the listed tags in the message header (if the message happens to have one) to open the mapping activity.
2) Click an item in the mapping activity to be able to edit it. Mind the cursor focus (I have no f'ing idea how this works).
... then click some other item to commit the change, because the back button will cancel it. Use a blank tagname to hide the tag.
3) Pressing back after editing the items will take you back to the message view, but the tags won't update right away. Reopen the message to see the new tagname (need a new hook here?).
4) When opening a message (say, after killing k-9 and restarting) you might see old tagnames instead of the ones you committed. Clicking "check mail" should fix it. That happens because the in-memory mapping is populated (from the keywords themselves) before the database can overwrite the defaults assumed by the UI.

I'm trying to get the most basic functionality working (reading keywords from the server and customizing the tags locally) first. I'd love ideas and patches to fix the most pressing usability problems, but I'm much more interested in making the internals good enough for a merge, possibly with a big "EXPERIMENTAL" sign next to some preference to enable the UI bits.

PanderMusubi:
Coming from Thunderbird, I would say you need:
- tag value (in the message) 
- mapping from tag value to a tag label (display in gui) 
- tag ordering (unique integer per value-label mapping to unambiguously sort messages on tags) (in Thunderbird, this is also they shortkey to toggle a tag on message(s) and is mapped to keyboad 1, 2, .., 9, 0) 
- tag color (font color for tag label)

equaeghe:
> ... in Thunderbird, this is ...

Please take care to not make the implementation so Thunderbird-specific as to conflict with other MUAs' interfaces.

Also, there are a number of registered keywords whose treatment should be special-cased: http://www.iana.org/assignments/imap-keywords/imap-keywords.xhtml

(There are also a lot of non-registered $keywords used by various servers and clients; I'd think special-casing them would not be advised, as their use is not standardized.)

Furthermore: not all IMAP servers support arbitrary keywords (Exchange...), so giving people the option of setting them there must be disabled.

igoralmeida:
On 11/13/2015 08:04 AM, Erik Quaeghebeur wrote:
>     ... in Thunderbird, this is ...
> 
> Please take care to not make the implementation so Thunderbird-specific
> as to conflict with other MUAs' interfaces.
> 
> Also, there are a number of registered keywords whose treatment should
> be special-cased:
> http://www.iana.org/assignments/imap-keywords/imap-keywords.xhtml
> 
> (There are also a lot of non-registered $keywords used by various
> servers and clients; I'd think special-casing them would not be advised,
> as their use is not standardized.)
> 
So far all keywords are equal, and the user decides what tag to assign
to each keyword.

> Furthermore: not all IMAP servers support arbitrary keywords
> (Exchange...), so giving people the option of setting them there must be
> disabled.
> 
Yes. Probably at SELECT time. However, for now, this is read-only and I
hope to get some help with the UI part so this can be merged. We can
then think about how to implement tagging in K-9.

> —
> Reply to this email directly or view it on GitHub
> <https://github.com/k9mail/k-9/pull/558#issuecomment-156398572>.
> 

-- 
Igor Almeida

PanderMusubi:
Would it be appreciated if we start a BountySource reward on this bug and bug 769?

richardboegli:
> Would it be appreciated if we start a BountySource reward on this bug and bug 769?
@PanderMusubi Yes.
Would it get pushed up into Kaiten though? I use Kaiten because I need rich text composing.

PanderMusubi:
I noticed that bounties are not often collected even though there are a lot out there, see https://www.bountysource.com/teams/k9mail

By the way, since this issue is closed, I cannot add to a bounty on this issue...

philipwhiuk:
Closing - needs further work, rebase and discussion. See #2260

PR #719, opened 2015-07-23
Don't use StringReader in HtmlConverter as calls to read have unneces…

dougsparling:
While profiling K-9 Mail on some older devices, I noticed that the performance of `HtmlConverter` was unexpectedly poor, regardless of whether it took the regular or simple conversion path. It turns out that `StringReader#read` acquires a lock internally, which is both unnecessary for the conversion, and adds quite a bit of overhead as it occurs once per character.

I replaced it with an for loop and index that essentially does the same thing as `StringReader`, minus locking.


jberkel:
thanks. don't know what the original intention to use a reader was here, it is indeed pointless.

PR #739, opened 2015-08-07
Fix an issue when passphrase is cancelled

GuillaumeSmaha:
Cancel the window asking passphrase crash the app due a missing error.

### DIFF
--- a/k9mail/src/main/res/values/strings.xml
+++ b/k9mail/src/main/res/values/strings.xml
@@ -1122,7 +1122,7 @@ Please submit bug reports, contribute new features and ask questions at
     <string name="openpgp_error">OpenPGP Error: %s</string>
     <string name="openpgp_unknown_error">Unknown OpenPGP Error</string>
     <string name="openpgp_user_id">User Id</string>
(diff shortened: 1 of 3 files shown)


k9mail/src/main/res/values/strings.xml
```diff
@@ -1122,6 +1122,7 @@ Please submit bug reports, contribute new features and ask questions at
     <string name="openpgp_error">OpenPGP Error: %s</string>
     <string name="openpgp_unknown_error">Unknown OpenPGP Error</string>
     <string name="openpgp_user_id">User Id</string>
+    <string name="openpgp_cancel_by_user">Cancel by user</string>
```
maniac103: Grammar issue here: this should be 'Cancelled by user'.

GuillaumeSmaha: Fixed

GuillaumeSmaha: Should I rename "openpgp_cancel_by_user" too ?

Diagoras:
+1

PR #778, opened 2015-09-05
Send client certificate with plain password

gaionim:
Changes seems harmless to me except userInfo pack and unpack.

 - widget for client certificate is always visible

- all auth type  (  I've tested only plain) sends cliantCertificateAlias
 for me works both with certificate ( my server) 
 and without cert ( gmail)

- userInfo pack and unpack seems weak, but I lack skills to fix

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,32 @@
+## Submitting issues
+
+If the app is not behaving like it should, it's not necessarily a bug. Please consult the following resources before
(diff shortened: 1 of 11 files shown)


gaionim:
be understanding, I am not prepared to contribute to a project, meant only to make a proposal.

I lack skills on git, so pushed my own comment, ( not thinking they will be public and some mess with hard reset.
The  diff in the code seems ok.

PR #875, opened 2015-10-30
fixes #575 : Display *Encrypted* for encrypted messages in message preview

### LINKED ISSUE #575: Display *Encrypted* for encrypted messages in message preview

alexandre75:
Hi, I wish I could read PGPMIME encrypted emails on k9 so I am attempting to help. Here is an attempt to fixes #575 .
This is my first time contributing to any opensource project so feel free to tell me if I forgot to fill something else somewhere before this pull request.

Description : 
- Added column encryption_type in message table
- all previously stored message have their encryption_type set to 0. So previously stored email will still display their previously computed preview. Sounds a bit too much work and computing to migrate the database properly.
- On reception, encryption_type is set to a value > 0 when emails are encrypted
- Display "*encrypted*" instead of column preview when encryption_type > 0

-Alexandre

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/EncryptionType.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/EncryptionType.java
@@ -1,0 +1,15 @@
+package com.fsck.k9.mail;
+
+/**
+ * Enumeration of the different possible encryption protocol that can be used.
+ */
+public enum EncryptionType {
+
+    NONE,
+
+    INLINE,
+
+    PGP_MIME,
+
+    S_MIME
+}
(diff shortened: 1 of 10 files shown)


clawoflight:
Is there any reason why you use an int for encryption_type? It sounds like you only use 0 and > 0.
If that is the case, a boolean would be more appropriate.

alexandre75:
A boolean would be fine to fixe #575. Thought, I chose to store the protocol used to encrypt within an enum because the information is available and it could be useful for further use. Let's say that we decide to display the type of encryption used instead of *encrypted*, it could be done without any further database migration.
Moreover sqlite doesn't support boolean type, so it would be stored as an INTEGER anyway.
But if you think this enum could be an issue I can still replace it be a boolean.

clawoflight:
Ah, okay. So you save an `int` in the database, but you use an `enum` in Java?
Makes sense now, looks good to me :)

alexandre75:
support for PGP/MIME #884

PR #884, opened 2015-11-03
support for PGP/MIME

alexandre75:
Hi,
This pull request was at first dedicated to #572. I am now piling up issues from PGP/MIME milestone because it simplifies testing and integration of linked issues scattered on multiple branches. I'll start a new branch once this one will be merged on mainstream unless said otherwise.

## Done
- [x] fixes #571 K9Mail no longer crash when opening a S/MIME signed message ;
- [x] fixes #572 Add support for PGP/MIME composing (Preference Account->Cryptography->Use PGP/MIME) ;
- [x] fixes #575 Display *Encrypted* in message preview ;
- [x] partially fixes #576 Composing of RFC-1847 message ;
- [x] fixes #588 Crash when cancelling passphrase input.

## What doesn't work/not tested/TODO
- [ ] #576 RFC-1847 messages are shown as "Not Signed"
- [ ] #575 Initialize `message.encryption_type` on db migration. I am not sure how messages were stored in the previous schema, so can't parse them. Any document describing the old schema?

## Design notes
1. `MessageCompose.onSend` attempts a dummy encrypt/sign in order to ensure further crypto operation (on the WorkerThread) will succeed. If interactions with PGP provider are required it launches them. If no further interaction were required,  it invokes `performSend` which continue on a worker thread and close the window. 
2. `MessageCompose.onActivityResult` fetches working Intents and invokes `performSend` if activity interactions were successful
3. `MessageCompose.createMessage` creates a *regular* message but then process the `MimeMessage` according to PGP/MIME specifications if cryptography was required

I'm waiting for your questions/advices. Don't hesitate.

Regards,
Alexandre

PR #897, opened 2015-11-13
fixes #588 crash when canceling decryption

### LINKED ISSUE #588: Crash when canceling passphrase input when decrypting messages
K-9 crashes when the passphrase input of OpenKeychain is cancelled.

alexandre75:
fixes #588
- Decryption error was detected but not properly handled.
- I also changed the error message in a separate commit from "OpenPGP error" to "Decryption failed" as it better describe what is going on in my opinion.
Thanks for examining this pull request,
Alexandre


alexandre75:
Merged in support for PGP/MIME #884

PR #928, opened 2015-12-02
If account isn't setup properly, show email address instead of account description (fix for #922)

s3b-:


s3b-:
I should recreate the fork and use a branch. Sorry!

PR #1198, opened 2016-03-21
Add PGP/INLINE "compatibility" mode

Valodim:
Another WIP PR. This adds pgp "compatibility" mode, i.e., pgp/inline.

Not happy with the UI at all, still thinking about how to do this better. It *might* be acceptable with this approach if we use a material style switch and not this ugly old thing, but not sure about that.

I'm thinking about adding another icon next to the crypto one while compat mode is enabled, and move the "enable compat mode" to the options. Ideally, use of pgp/inline should be an exceptional state, so it's not a huge problem with screen estate, and if the user clicks it they can be provided with an explanation of why it was enabled automatically (or disable it).


philipwhiuk:
Do you think compatability means more to most people than inline? I'm not sure which we should be using both internally and in the UI.

k9mail/src/main/java/com/fsck/k9/activity/compose/ComposeCryptoStatus.java
```diff
@@ -102,6 +103,9 @@ public boolean isSigningEnabled() {
         return cryptoMode != CryptoMode.DISABLE && signingKeyId != null;
     }
 
+    public boolean isCompatModeEnabled() {
+        return enableCompat;
+    }
```
philipwhiuk: ComposeCryptoStatus is somewhat technique agnostic, where as this is very much an OpenPGP setting. Just wondering whether you've considered SMIME while developing this?

k9mail/src/main/java/com/fsck/k9/message/PgpMessageBuilder.java
```diff
@@ -33,6 +33,7 @@
 
     public static final int REQUEST_SIGN_INTERACTION = 1;
     public static final int REQUEST_ENCRYPT_INTERACTION = 2;
+    public static final int REQUEST_INLINE_INTERACTION = 3;
```
philipwhiuk: INLINE refers to inline sign or inline verification? The German Wikipedia page suggests it can do both: https://de.wikipedia.org/wiki/PGP/INLINE

k9mail/src/main/java/com/fsck/k9/activity/compose/CryptoSettingsDialog.java
```diff
@@ -104,21 +118,41 @@ public void onCryptoStatusSelected(int status) {
...
+        }
+
+        activity.onCryptoCompatChanged(switchState);
+    }
+
+    @Nullable
+    private OnCryptoModeChangedListener getCryptoModeChangeListenerActivity() {
         Activity activity = getActivity();
         if (activity == null) {
             // is this supposed to happen?
```
philipwhiuk: So - is it supposed to happen? Is a null activity an expected scenario? Should it be considered an error case in each of these checks?

k9mail/src/main/java/com/fsck/k9/message/PgpMessageBuilder.java
```diff
@@ -221,7 +301,9 @@ public void writeTo(OutputStream os) throws IOException {
                 if (currentState == State.OPENPGP_SIGN) {
                     mimeBuildSignedMessage(bodyPart, result);
                 } else if (currentState == State.OPENPGP_ENCRYPT) {
-                    mimeBuildEncryptedMessage(encryptedTempBody, result);
+                    mimeBuildEncryptedMessage(pgpResultTempBody, result);
+                } else if (currentState == State.OPENPGP_INLINE) {
+                    mimeBuildInlineMessage(pgpResultTempBody, result);
                 } else {
```
philipwhiuk: Does inline always mean encrypt then? From my limited reading, inline can including signing instead?

k9mail/src/main/java/com/fsck/k9/activity/compose/CryptoSettingsDialog.java
Valodim: both times this is called, it is triggered by a user interaction. That can only happen when the activity is available, so this *should* never happen. Android activity lifecycle does wacky things though, so I wouldn't throw an illegal state here...

k9mail/src/main/java/com/fsck/k9/activity/compose/ComposeCryptoStatus.java
Valodim: I have not considered s/mime here. do you have any other way to handle this in mind?

about the other option, I don't think that one needs to be exposed to the user. I'm not even sure we need to support it...

k9mail/src/main/java/com/fsck/k9/message/PgpMessageBuilder.java
Valodim: inline can be either or both. in both cases though, it is a single pgp operation replacing the plain text of the mail with its pgp clearsigned/encrypted version, so it doesn't need two passes like sign/encrypt.

JasperWeiss:
+1 For material design. I'm unable to decrypt PGP/Inline in the latest alpha, is support for inline pgp dropped?

Valodim:
Worked on this one a bit. I moved the setting into the overflow menu, added a dialog with an explanation, and also an indicator next to the crypto status icon. Not quite happy with the icon itself I think, I used the only remotely suitable one for "compatibility" I could find ([source](https://thenounproject.com/search/?q=compatibility&i=14437)). I also added a flag so the mode is preserved for drafts.

Have a screenshot:



Still missing:
- display explanation only once
- enable automatically based on replied to message
- possibly: flag contact when message is sent and enable automatically for this contact in the future?
- refuse to add attachments while enabled

I'm also still on the fence about the "compatibility mode" vs "pgp inline" deal.

Valodim:
Alternative icon suggestion:

philipwhiuk:
Text improvement: "The e-mail will be sent..."

I'm ambivalent about the icon - both are fine.

The extra bits sound good - I guess you have to refuse attachments because they would be sent in plain-text?

JasperWeiss:
I like the icon, but I'm not sure if it's obvious to the user that bandage represents compatibility

cketti:
UI consistency issues:
* We use "email" instead of "e-mail" everywhere else
* In the overflow menu it's "Inline-PGP", in the dialog "Inline/PGP"
* The compatibility icon is a different shade of grey than the action icons and the crypto indicator
* The header of the dialog is sufficiently different from the other dialogs that it bothers me (a little)

dschuermann:
Just two smaller comments: I like the medical Patch Icon, it shows that PGP/INLINE is an "ugly patch", which it is :) 
I would prefer that it is always called " PGP/INLINE".

Valodim:
* email it is
* see below
* crypto indicator is a "disabled"-grey, so that's sort of supposed to be different.
* the headline is material dialog style, with an icon to reinforce the icon association. It's sort of a general issue I have that designing things holo now feels like a waste of time (and looks bad), but I can't make them properly material either...

@dschuermann  I dislike the typographic impression of "PGP/INLINE" in a text, particularly in the overflow menu it just looks bad. Also it's an acronym, followed by a slash, followed by a capitalized word? Is that last bit also an acronym? How do you pronounce a slash? Something like "Inline-PGP" is still recognizable to everyone who knows what "PGP/INLINE" is, but doesn't have that weird overly technical typographic vibe.

That said, I'm still not sure how to handle this everywhere. I definitely don't want PGP/INLINE in the overflow menu, in the text it's sort of fine but then also inconsistent. At the moment I use the notion of "compatibility mode" in some places, like the dialog title, which seems like a good layman's term and two-word-title for what PGP/INLINE really is.

dschuermann:
@Valodim I still stand by my opinion that I would not use two notations for the same technical term, at least use "Inline-PGP" at both places.

I also like to see something like "compatibility mode" or "backward compatibility mode" in menu and dialog title.

cketti:
"Compatibility mode" is not descriptive enough. As a user I'd have no idea what it refers to. So it'd have to be something like "PGP compatibility mode".

dschuermann:
@cketti you are right. Unfortunately, "PGP compatibility mode" makes it sound like it needs to be enabled to make emails compatible to PGP :(

What about "PGP backward compatibility"?

Valodim:
Apart from wording issues and such, there is the question of when to show the explanatory dialog to the user, and with what buttons. There are three instances where the dialog may or may not appear:

* when the user clicks the pgp/inline icon. this is a rather clear case: the dialog must be shown, and the standard action is "disable".
* when the user enables pgp/inline from the overflow menu. this could be handled without showing the dialog, however a first-time explanation might be useful. the natural flow from there is "enable" (or possibly just one button, "got it")
* when pgp/inline is enabled automatically, e.g. when replying to a pgp/inline encrypted e-mail. this definitely shouldn't happen every time, so maybe do the first-time thing there as well?

k9mail/src/main/java/com/fsck/k9/activity/compose/OpenPgpInlineDialog.java
```diff
@@ -0,0 +1,81 @@
...
+
+        @SuppressLint("InflateParams")
+        View view = LayoutInflater.from(activity).inflate(R.layout.openpgp_inline_dialog, null);
+
+        AlertDialog.Builder builder = new AlertDialog.Builder(activity);
+//        builder.setTitle("Compatibility Mode");
+        builder.setView(view);
+
+        if (getArguments().getInt(ARG_FIRST_TIME) != 0) {
+            builder.setPositiveButton("Got it!", new OnClickListener() {
```
cketti: Hardcoded text

k9mail/src/main/java/com/fsck/k9/activity/compose/OpenPgpInlineDialog.java
```diff
@@ -0,0 +1,81 @@
...
+                @Override
+                public void onClick(DialogInterface dialog, int which) {
+                    dialog.dismiss();
+                }
+            });
+        } else {
+            builder.setPositiveButton("Disable", new OnClickListener() {
```
cketti: Hardcoded text

k9mail/src/main/java/com/fsck/k9/activity/compose/OpenPgpInlineDialog.java
```diff
@@ -0,0 +1,81 @@
...
+                    Activity activity = getActivity();
+                    if (activity == null) {
+                        return;
+                    }
+
+                    ((OnOpenPgpInlineChangeListener) activity).onOpenPgpInlineChange(false);
+                    dialog.dismiss();
+                }
+            });
+            builder.setNegativeButton("Keep Enabled", new OnClickListener() {
```
cketti: Hardcoded text

k9mail/src/main/res/values/strings.xml
```diff
@@ -1152,5 +1152,13 @@ Please submit bug reports, contribute new features and ask questions at
     <string name="crypto_mode_private">Always Sign, Always Encrypt.</string>
     <string name="error_crypto_provider_connect">Cannot connect to crypto provider, check your settings or click crypto icon to retry!</string>
     <string name="error_crypto_provider_ui_required">Crypto provider access denied, click crypto icon to retry!</string>
+    <string name="error_crypto_inline_attach">PGP/INLINE mode does not support attachments!</string>
+    <string name="enable_inline_pgp">Enable Inline-PGP</string>
+    <string name="disable_inline_pgp">Disable Inline-PGP</string>
+    <string name="openpgp_inline_title">Compatibility Mode</string>
+    <string name="openpgp_inline_text">The e-mail will be sent in PGP/INLINE format.\nThis should only be used for compatibility:</string>
```
cketti: Use "email" instead of "e-mail".

k9mail/src/main/res/values/strings.xml
```diff
@@ -1152,5 +1152,13 @@ Please submit bug reports, contribute new features and ask questions at
     <string name="crypto_mode_private">Always Sign, Always Encrypt.</string>
     <string name="error_crypto_provider_connect">Cannot connect to crypto provider, check your settings or click crypto icon to retry!</string>
     <string name="error_crypto_provider_ui_required">Crypto provider access denied, click crypto icon to retry!</string>
+    <string name="error_crypto_inline_attach">PGP/INLINE mode does not support attachments!</string>
+    <string name="enable_inline_pgp">Enable Inline-PGP</string>
```
cketti: Let's consistently use PGP/INLINE.

Valodim:
rebased, incorporated wording changes. this works acceptably now

cketti:
When sending a PGP/Inline-encrypted email the decrypted message is quoted-printable encoded.
Also, the message doesn't seem to be signed.

Valodim:
about missing signature, I couldn't reproduce in a quick test. please test again with the compose-combined branch. it simplifies a lot of things in PgpMessageBuilder, which might very well fix this problem.

Valodim:
maintaining two PRs which depend on each other is quite annoying. closing this one, please see #1323 which builds on this branch but makes PgpMessageBuilder a lot simpler, adds tests, and fixes bugs from this one.

Issue #1224, opened 2016-03-28
Crypto icon grey on reply


dschuermann:
When replying to an encrypted email, the crypto Icon in the top right corner is grey, but should instead be orange/green. After clicking on it and closing the dialog it is correct.

philipwhiuk:
Closing as the pull request was merged in.

PR #1229, opened 2016-03-29
ditch crypto status icon animation

Valodim:
The animation of the crypto status icon is not used anymore because the visibility of the icon is static depending on the availability of a crypto provider. It also led problems when the animation was cancelled (#1224). So this PR simply ditches that animation to fix that.

(I also added my ToolableViewAnimator for easier previewing)


Valodim:
@cketti ping

PR #1303, opened 2016-04-18
Add support for encapsulated PGP/MIME format (WIP)

Valodim:
Fix for #576, i.e. https://tools.ietf.org/html/rfc3156#section-6.1

I split the crypto processing into two passes, where the first only handles encrypted parts, the second only signed parts. We can probably simplify this even more because we don't want to handle multiple encrypted/signed parts at all anymore, but I tried to touch as little as possible to get this working.

Still need to profile this, it feels really sluggish. Might just be subjective...

(Note PR is not based on master but on #1299)


philipwhiuk:
There was that issue for progress bars for decrypting - my device isn't slow enough to really test that issue - we can always add it in if it's noticeable with a multiphase process (not that I'm in favour of simultaneously removing features and slowing the app down).

k9mail/src/main/java/com/fsck/k9/ui/crypto/MessageCryptoAnnotations.java
```diff
@@ -25,4 +25,13 @@ public CryptoResultAnnotation get(Part part) {
     public boolean has(Part part) {
         return annotations.containsKey(part);
     }
+
+    public Part findKeyForAnnotationWithReplacementPart(Part part) {
```
ligi: this would be such a nice one-liner with kotlin ;-)

Valodim: it would probably be a nice one-liner even with java 8 ;)

Valodim:
closing this one for now, I'll get back to it later

PR #1323, opened 2016-04-22
switch from encapsulated to combined pgp/mime format

Valodim:
...what the title says.

PgpMessageBuilder got a hundred lines lighter in the process and lost the state machine, the logic should now be a lot easier to read. Pretty happy with that.


philipwhiuk:
Would be nice to see tests covering the general behavior.

Valodim:
that's what I'm working on at the moment ;)

Valodim:
Pushed what I have. The async call and openpgp api structure does not exactly yield nice and clean testing code, but at least some of the behavior is pinned down now.

PR #1526, opened 2016-07-29
FancyPart (WIP)

Valodim:
This PR moves all methods which return metadata based on the headers of a part out of the `Part` class, and into `FancyPart`. Yes this is not a good name, please suggest a better one. Something like `PartMetadataExtractor`? or just `PartMetadata`?


Valodim:
rebase nearly impossible at this point

PR #1540, opened 2016-08-03
use vector drawables for all openpgp icons

Valodim:
does what it says on the tin. I only tested this on my android 6 device, but checked the .apk file and there are .png files in `drawable-*dpi-v4` folders and the vector drawables themselves in `drawable-anydpi-v21`, as one would expect.


Valodim:
(also removed a couple of unused icons)

PR #1725, opened 2016-10-18
[WIP] Support for viewing iCal event invitations in K-9

philipwhiuk:
Work in progress solution for #930 Needs unit tests writing.

Features:
- Viewing basic info on attached iCal event
- Retains opening and saving

Testing:
- Tested against ICS files from Microsoft Outlook
- Ongoing daily client usage tests


philipwhiuk:
I've seen that this doesn't cover the reply case, only the initial invite.

I'll update this PR with work on that soon-ish.

PR #1747, opened 2016-10-24
Back-end changes for Google XOAUTH2

philipwhiuk:
Splitting out from #1295

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/oauth/OAuth2TokenProvider.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/oauth/OAuth2TokenProvider.java
@@ -1,61 +1,60 @@
 package com.fsck.k9.mail.oauth;
 
+
+import java.util.List;
+
 import android.app.Activity;
 
 import com.fsck.k9.mail.AuthenticationFailedException;
 
-import java.util.List;
 
 public interface OAuth2TokenProvider {
-
     /**
      * A default timeout value to use when fetching tokens.
      */
-    public static final int OAUTH2_TIMEOUT = 30000;
+    int OAUTH2_TIMEOUT = 30000;
 
+    
     /**
      * @return Accounts suitable for OAuth 2.0 token provision.
      */
     List<String> getAccounts();
 
     /**
-     * Provides an asynchronous response to an
-     * {@link OAuth2TokenProvider#authorizeAPI(String, Activity, OAuth2TokenProviderAuthCallback)} request
-     */
-    interface OAuth2TokenProviderAuthCallback {
-
-        void success();
-
-        void failure(AuthorizationException e);
-    }
-
-    /**
      * Request API authorization. This is a foreground action that may produce a dialog to interact with.
-     * @param username Username
-     * @param activity The responsible activity
-     * @param callback A callback to process the asynchronous response
+     *
+     * @param username
+     *         Username
+     * @param activity
+     *         The responsible activity
+     * @param callback
+     *         A callback to process the asynchronous response
      */
-    void authorizeAPI(String username, Activity activity,
-                        OAuth2TokenProviderAuthCallback callback);
+    void authorizeApi(String username, Activity activity, OAuth2TokenProviderAuthCallback callback);
 
     /**
      * Fetch a token. No guarantees are provided for validity.
-     * @param username Username
-     * @return Token string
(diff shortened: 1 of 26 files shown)


k9mail-library/src/main/java/com/fsck/k9/mail/transport/SmtpTransport.java
```diff
@@ -638,30 +656,90 @@ private void checkLine(String line) throws MessagingException {
...
+        private final String message;
+
+        public CommandResponse(int replyCode, String message) {
+            this.replyCode = replyCode;
+            this.message = message;
+        }
+    }
+
+    private CommandResponse executeSimpleCommandWithResponse(String command, boolean sensitive) throws IOException, MessagingException {
+        List<String> results = new ArrayList<String>();
```
Valodim: diamond operator~

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
```diff
@@ -356,6 +371,75 @@ private void authenticate() throws MessagingException, IOException {
...
+        );
+    }
+
+    private void handleXOAuthUntaggedResponse(ImapResponse response) throws IOException {
+        if (response.isString(0)) {
+            retryXoauth2WithNewToken = XOAuth2ChallengeParser.shouldRetry(
+                    response.getString(0), settings.getHost());
+        }
+
+        if(response.isContinuationRequested()) {
```
philipwhiuk: If it was me I'd have a commit that fixed all the errors and made the build fail but nobody's a fan of that :(

I'll fix it.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapResponseParser.java
```diff
@@ -109,7 +109,7 @@ private void readTaggedResponse(ImapResponseCallback callback) throws IOExceptio
                 continue;
             }
 
-            if (untaggedHandler != null) {
+            if (response.getTag() == null && untaggedHandler != null) {
```
philipwhiuk: I think it only didn't fail a test because the test was badly written. I wrote some better tests for it and I think it should have failed. To me it looked like faulty logic and I browsed the spec a bit.

I'll keep using it on my device though.

k9mail-library/src/main/java/com/fsck/k9/mail/oauth/OAuth2TokenProvider.java
```diff
@@ -0,0 +1,61 @@
+package com.fsck.k9.mail.oauth;
+
+import android.app.Activity;
+
+import com.fsck.k9.mail.AuthenticationFailedException;
+
+import java.util.List;
+
+public interface OAuth2TokenProvider {
```
jberkel: where is this interface actually implemented? don't see it in this PR.

philipwhiuk: It's not. This is just the back-end changes. That's an interface for the front-end to implement.

jberkel: Ah OK, makes sense, was confused because OAuth2TokenProvider sounds like back-end functionality. Does only `authorizeAPI()` need user interaction?

philipwhiuk: That's the expectation, yes. 

An (imperfect) implementation that supports GMail only is here: https://github.com/philipwhiuk/k-9/commit/7bd0f52c6d24ba1696620c4109c30d1f06aa770f#diff-4e75c10b7bac153fec6888b448a738ea

HLFH:
Thanks for that great work!

I hope we'll get a release soon to try this great feature.

sivaraam:
I came here after some time of searching for open issues about Gmail OAuth logins. It's been around 1.5 years and as far as I could see I've not found OAuth support for logging into Gmail in the app. What's the current progress of this?

BerndErnst:
@sivaraam Don't worry. Nothing has changed since your last message until now.

wiktor-k:
Actually there is a PR #3804 that enables Gmail OAuth login.

PR #1808, opened 2016-11-22
Testing for settings import and export

philipwhiuk:
Hijacked from work I did while developing XOAUTH2's abandoned front-end.

Would also be nice to get a test or two that validates export, delete, import produces an identical account (less passwords) (possibly using some constrained non-deterministic data).


Valodim:
y u no newline in tests :hurtrealbad:

PR #1844, opened 2016-12-09
Added config for fairnatics.net, a soon upcoming service.

supersn0b:
Hi,

we are about to startup a new service focusing on fair and private principles.
One key component will be k9, so we would like to have our mail configuration supported.

Thanks!

### DIFF
--- a/k9mail/src/main/res/xml/providers.xml
+++ b/k9mail/src/main/res/xml/providers.xml
@@ -624,7 +624,7 @@
 
     <!-- EU wide -->
     <provider id="fairnatics.net" label="Fairnatics" domain="fairnatics.net">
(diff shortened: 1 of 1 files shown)


k9mail/src/main/res/xml/providers.xml
```diff
@@ -622,6 +622,12 @@
         <outgoing uri="smtp+tls+://smtp.ziggo.nl" username="$email" />
     </provider>
 
+    <!-- EU wide -->
+    <provider id="fairnatics.net" label="Fairnatics" domain="fairnatics.net">
+        <incoming uri="imap+ssl+://mail.fairnatics.net:993" username="$email" />
```
vt0r: No need for `:993` here, as this is the default port for `imap+ssl+`

PR #1894, opened 2016-12-31
Use theme-able icon for CC/BCC expander

### LINKED ISSUE #1893: Cc/Bcc expander barely visible in dark theme
### Expected behavior
K-9 Mail version 5.115 used to provide the option of choosing a TO, CC or BCC recipient in the selection field at the very top of the compose an email window. This is no longer available in version 5.2. There should also be the option to choose a To, CC or BCC recipient in the pull down list 

### Actual behavior
When composing an email one can't select a CC nor BCC recipient unless one has checked "Always show Cc/Bcc" in the settings for each email account. 

### Steps to reproduce
1. Compose an email 
2. Can't select Cc nor Bcc recipients for the email 

### Environment
K-9 Mail version:
5.2
Android version:
6.0.1
POP.

philipwhiuk:
Fixes #1893

### DIFF
--- a/k9mail/src/main/res/layout/message_compose_recipients.xml
+++ b/k9mail/src/main/res/layout/message_compose_recipients.xml
@@ -291,7 +291,7 @@
                 android:layout_height="match_parent"
                 android:padding="6dp"
                 android:scaleType="centerCrop"
-                android:src="@drawable/ic_action_expand_light"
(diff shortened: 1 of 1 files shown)


Valodim:
Heh, so we actually had the icon around but just didn't use it. Simple fix :+1:

PR #1987, opened 2017-01-05
Widget Size Fix

### LINKED ISSUE #1981: K9 Unread Widget Size is 1x2 or 2x2 instead of 1x1
I first noticed this on my wifes klte with CM12 - for some reason (unexplicable to mee), the unread widget has the size 2x2 instead of 1x1 for at least a year now (uninstall and reinstall of k9 didn't help).

Now, since I updated to K9 5.2, it also happened on my klte with CM13.0, but the size is now 1x2 (as you can see in the attached screenshot). The widget is not resizable after added to a homescreen.
k9widgetsize

### Expected behavior
The widget should always be 1x1 in size, as it was before.

### Actual behavior
On some devices, the size is off - 1x2 or 2x2 on two of my devices.

### Steps to reproduce
Can't reproduce - can get rid of, either, by uninstalling/reinstalling. Also did a wipe of K9 data once on my wifes phone, to no avail.

### Environment
K-9 Mail version: 5.201 currently, was already for some months before on one device.

Android version: 5.0/6.0 (CM 12/13)

GoneUp:
It could fix #1981. I could not reproduce it with my phone, but the minValues did not match google's recommendation. 
https://developer.android.com/guide/practices/ui_guidelines/widget_design.html#anatomy


philipwhiuk:
This makes sense to me and explains why it's only sporadically reproducible. Up to @cketti whether we merge it into 5.2-MAINT but I'm happy to see it go in.

PR #1995, opened 2017-01-06
Allow pgp/inline with sign-only again

### LINKED ISSUE #1974: Inline PGP Signatures on Non-encrypted mail
### Expected behavior
I'd like to be able to send and receive non-encrypted emails that use inline PGP signatures. Basically, using OpenKeychain, I don't want to send an attachment called `signature.asc`, I want a message that will look like: 

```
-----BEGIN PGP SIGNATURE-----
my message

-----END PGP SIGNATURE------
<signature>
```

### Actual behavior
You cannot use inline mode with sign-only mode. Not sure why.

### Steps to reproduce
1. Compose a new email
2. Click on the hamburger menu in the top right
3. Enable PGP/INLINE mode
4. Try to enable PGP Sign Only mode as well
5. It is unsupported

### Environment
K-9 Mail version: 5.201

Android version: 6

Account type (IMAP, POP3, WebDAV/Exchange): N/A

Why was the decision made to limit the way signatures are added to the message? Is this something OpenKeychain does not support? Because as far as I can tell, K9 can decode _both_ types of email, and show that the signatures are verified, but you cannot compose mail with the above settings.

Valodim:
While making sign-only a mode of its own I disabled clearsign, to avoid having two "modes" at the same time. Clearsign is a legitimate use case though, so this patch readds it.

I'm not *super happy* with the icon as it is now, but then again both sign-only and pgp/inline are out of most users' way so far that I don't really care... :)

Fixes #1974.

\\ edit

Screenshot for reference:

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/activity/compose/ComposeCryptoStatus.java
+++ b/k9mail/src/main/java/com/fsck/k9/activity/compose/ComposeCryptoStatus.java
@@ -93,12 +93,16 @@ public class ComposeCryptoStatus {
             return CryptoSpecialModeDisplayType.NONE;
         }
 
-        if (isPgpInlineModeEnabled()) {
-            return CryptoSpecialModeDisplayType.PGP_INLINE;
+        if (isSignOnly() && isPgpInlineModeEnabled()) {
+            return CryptoSpecialModeDisplayType.SIGN_ONLY_PGP_INLINE;
         }
 
         if (isSignOnly()) {
             return CryptoSpecialModeDisplayType.SIGN_ONLY;
+        }
+
+        if (isPgpInlineModeEnabled()) {
+            return CryptoSpecialModeDisplayType.PGP_INLINE;
         }
 
         return CryptoSpecialModeDisplayType.NONE;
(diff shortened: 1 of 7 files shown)


philipwhiuk:
I'm confused - isn't PGP/MIME vs INLINE PGP merely a transport layer issue?

What makes clear-sign valid and sign-only PGP/MIME invalid?

This all seems rather messy in terms of what we support  :(

Valodim:
We support pgp/inline and pgp/mime as encodings, however before this patch pgp/inline wasn't available for sign-only.

philipwhiuk:
Oh I didn't realise that that we'd ever stopped supporting reading any PGP-based mail, merely sending it.

Valodim:
I was indeed talking about sending :)

philipwhiuk:
So **why** are we supporting INLINE sign only sending when we don't support MIME sign only sending. (I thought this was the question I asked above).

ArchangeGabriel:
> So *why* are we supporting INLINE sign only sending when we don't
> support MIME sign only sending. (I thought this was the question I
> asked above).
>

It’s supported but walled behind an option in crypto settings, right?
But as I understand it, this option only makes it work with PGP/MIME,
not with inline, which this PR intend to fix.

PR #2005, opened 2017-01-07
Delete data from Full-Text-Search table

### LINKED ISSUE #1917: Database size doesn't decrease
Before the update when I would empty the trash the mailbox size would reduce now after the update this is not happening. Even after using the option to clear local Messages the size of the mailbox was only reduced by a very small amount . 

 Before the update the size of my Nina email box after clearing all trash was approximately 92 KB to 127 KB now after deleting all local messages it is still over 1 MB.

 I am using canine mail app for four different email accounts sure the size of the mailboxes is quite excessive .

K-9 Mail version:  5.201

Android version:  6.0.1

Account type (IMAP, POP3, WebDAV/Exchange): POP3 and IMAP

Galaxy Note 5

philipwhiuk:
Should fix #1917 

I'm guessing there's other places where we might not be deleting from full-text search (i.e. on deleting messages from the store) but this fixes the problem in the bug report.

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/mailstore/LocalMessage.java
+++ b/k9mail/src/main/java/com/fsck/k9/mailstore/LocalMessage.java
@@ -322,6 +322,8 @@ public class LocalMessage extends MimeMessage {
                         throw new WrappedException(e);
                     }
 
+                    deleteFulltextIndexEntry(db, mId);
+
                     return null;
                 }
             });
--- a/k9mail/src/main/java/com/fsck/k9/mailstore/LocalStore.java
+++ b/k9mail/src/main/java/com/fsck/k9/mailstore/LocalStore.java
@@ -153,7 +153,7 @@ public class LocalStore extends Store implements Serializable {
      */
     private static final int THREAD_FLAG_UPDATE_BATCH_SIZE = 500;
 
-    public static final int DB_VERSION = 55;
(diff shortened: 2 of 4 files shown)


Valodim:
can't we simply add a trigger ON DELETE messages to the db? That should handle all of this much more consistently than trying to do the bookkeeping ourselves everywhere.

philipwhiuk:
That wouldn't work because we don't actually delete deleted messages IDs I think. There's a comment about leaving the deleted messages IDs so we don't resync them. 

Also I'm not a fan of triggers - devolving logic to the database always feels wrong to me.

In any case, even if we added triggers, I'd still want to clear up here.

Valodim:
I'm a big fan of database integrity though, and if this was doable with foreign keys (ON DELETE CASCADE) we would definitely have those. There is no case where deletion of a message (or its content /root part, at least) should leave dependent rows around, so I'd say it's the way to go.

philipwhiuk:
I'm happy with your changes to it @cketti

korelstar:
I've seen, that there are relatively many open issues in the milestone v5.203. Therefore, I would like to kindly ask if you could release v5.203 at the current state (this fix included) as soon as possible, since my mail storage is growing and growing ...  :blush:
This would be very cool, thanks! :+1:

PR #2131, opened 2017-01-24
Settings classes cleanup

Valodim:
The .preferences.* classes have not been touched in a long time and had a lot of warnings, particularly use of the diamond operator, unnecessarily broad visibility, typecasts that can be handled with generics, and unnecessary unboxing.

I also broke down the `upgrade` method into smaller methods because it took me way too long to figure out what was going on.

There should be no semantic changes in this entire PR, besides a checked assertion in `Settings.upgrade`.

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/preferences/Settings.java
+++ b/k9mail/src/main/java/com/fsck/k9/preferences/Settings.java
@@ -142,11 +142,14 @@ public class Settings {
                 }
 
                 SettingsDescription setting = versionedSettings.get(toVersion);
+                if (setting == null) {
+                    throw new AssertionError("First version of a setting must be non-null!");
+                }
                 upgradeSettingInsertDefault(validatedSettingsMutable, settingName, setting);
             }
 
             Integer highestVersion = versionedSettings.lastKey();
-            boolean isRemovedSetting = highestVersion == toVersion && versionedSettings.get(highestVersion) == null;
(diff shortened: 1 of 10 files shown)


k9mail/src/main/java/com/fsck/k9/preferences/Settings.java
```diff
@@ -97,77 +98,87 @@
...
 
         for (int toVersion = version + 1; toVersion <= VERSION; toVersion++) {
-
-            // Check if there's an SettingsUpgrader for that version
-            SettingsUpgrader upgrader = upgraders.get(toVersion);
-            if (upgrader != null) {
-                deletedSettings = upgrader.upgrade(upgradedSettings);
+            if (customUpgraders.containsKey(toVersion)) {
+                SettingsUpgrader upgrader = customUpgraders.get(toVersion);
+                deletedSettings = upgrader.upgrade(validatedSettingsMutable);
```
Valodim: I didn't change semantics here, but it seems odd to me: is keeping only the set of `deletedSettings`  of the last upgrader really what we want to do here?

k9mail/src/main/java/com/fsck/k9/preferences/Settings.java
```diff
@@ -274,13 +283,13 @@
      * negligible.
      * </p>
      */
-    public static abstract class SettingsDescription {
+    static abstract class SettingsDescription<A> {
```
Valodim: `PseudoEnumSetting` already had a parameter A, which influenced me there. I agree it should be T :)

PR #2138, opened 2017-01-25
Move recipient tokens via drag and drop (sort of WIP)

Valodim:
I dabbled around with android's drag and drop mechanism a little yesterday. Seems like it's not too difficult to get this to work. But it feels like a lot of poking in the dark to get to a nice user experience and I don't have the time for that.

It would be a nice feature though for moving recipients around between To/Cc/Bcc, so I'll just leave this branch here. Maybe someone else wants to fiddle around with it some more. If not, feel free to close in a while.


rhari991:
I'd like to work on this.

Valodim:
Cool! I suppose the goal right now is to make it "acceptably useful", but it's tough to say what that means exactly :)

PR #2141, opened 2017-01-26
MessageCompose cleanup

Valodim:
Cleaning up these classes while we don't have a lot of MessageCompose changes in the queue. This mostly just fixes android studio warnings, cleans up the order of fields and gets rid of the m prefix.


Valodim:
(whoops, missed a newline there in hub :hurtrealbad:)

PR #2151, opened 2017-01-29
Address file clean-up

philipwhiuk:
Lint check fixes.

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/Address.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/Address.java
@@ -84,7 +84,7 @@ public class Address implements Serializable {
             return null;
         }
 
-        return mAddress.substring(hostIdx+1);
+        return mAddress.substring(hostIdx + 1);
     }
 
     public void setAddress(String address) {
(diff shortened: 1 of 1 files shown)


k9mail-library/src/main/java/com/fsck/k9/mail/Address.java
```diff
@@ -95,7 +95,8 @@ public String getPersonal() {
         return mPersonal;
     }
 
-    public void setPersonal(String personal) {
+    public void setPersonal(String newPersonal) {
+        String personal = newPersonal;
```
Valodim: What's the point of this?

Related note, this method is doing a check for empty string and *then* trims, so it can still result in an empty string being assigned.

philipwhiuk: There's a checkstyle warning for assigning parameters.

        <module name="ParameterAssignment">
            <property name="severity" value="warning"/>
        </module>

I've been slowly trying to get rid of the lint, checkstyle and findBugs warnings. But this one is a bit lame so we could just turn it off. 

Good catch about the trim point.

PR #2176, opened 2017-02-04
MessageProvider related fixes and testing

philipwhiuk:
Mainly a PR to add tests to MessageProvider plus some tiny tweaks I found while testing.


k9mail/src/main/java/com/fsck/k9/Account.java
```diff
@@ -1840,7 +1840,7 @@ public void excludeUnwantedFolders(LocalSearch search) {
     }
 
     private void excludeSpecialFolder(LocalSearch search, String folderName) {
-        if (!K9.FOLDER_NONE.equals(folderName)) {
+        if (folderName != null && !K9.FOLDER_NONE.equals(folderName)) {
```
Valodim: This is unnecessary, `equals` will return false on a null value (since it doesn't equal a string)

PR #2190, opened 2017-02-05
Don't build messages without a format

### LINKED ISSUE #2188: K-9 sometimes does not send a complete MIME header
### Expected behavior
K-9 should send a correct MIME email message.

### Actual behavior
The message contains the MIME stuff (boundaries, headers, etc) in the mail content, but _does not_ send the "MIME-Version" header nor the "Content-Type" header in the email's header part.

### Steps to reproduce
1. Click on the "New message" icon
2. Choose a receiver (yourself if you're testing)
3. Paste a text from a web page into the _Subject_ line of the mail that includes formatting (for example, a text from a &lt;h1&gt;...&lt;/h1&gt;); the text will look formatted in the input box (which I doubt it should, subjects cannot contain formatting)
4. Type an arbitrary body text
5. Send the message

### Environment
K-9 Mail version: 5.203

Android version: 5.1.1

### Sent email (source view)
```
Return-Path: <sven@berkvens.net>
X-Original-To: sven@berkvens.net
Delivered-To: sven@localhost
Received: from SvensNote3.dynamic.berkvens.net (whiskers.berkvens.net [80.100.221.128])
	(using TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384 (256/256 bits))
	(No client certificate requested)
	by mail.csaw.nl (Postfix) with ESMTPSA id 6C702340136;
	Sun,  5 Feb 2017 14:51:37 +0100 (CET)
Date: Sun, 05 Feb 2017 14:51:35 +0100
Subject: Some formatted text
To: Sven Berkvens-Matthijsse <sven@berkvens.net>
From: Sven Berkvens-Matthijsse <sven@berkvens.net>
Message-ID: <5D3A2FF3-A1C6-45CA-80F6-732C87E04424@berkvens.net>
X-SpamProbe: GOOD 0.0351478 7a76b3a682e10d01f73efc258f2dc661

------2LNBJIVCUUGCF2F3DBHG6IYX800DH3
Content-Type: text/plain;
 charset=utf-8

Test
------2LNBJIVCUUGCF2F3DBHG6IYX800DH3
Content-Type: text/html;
 charset=utf-8

Test
------2LNBJIVCUUGCF2F3DBHG6IYX800DH3--
```

### Extra information
As you can see above, the email is lacking certain headers that it should have had, as discussed above.

Not sure whether this is related, and I could investigate further if anyone wants me  to, but I use MessagEase as a keyboard. This keyboard allows pasting directly from the keyboard itself. It doesn't matter whether I paste using that function or simply by long-pressing the _Subject_ line and then choosing _PASTE_ from the context menu.

I copied the text from this site: http://www.marktplaats.nl/a/huis-en-inrichting/bureaus-en-bureaustoelen/a1009985199-bureaustoel-bock-burostoel-zwart-verstelbare-armleggers.html . For example, copy the _Bureaustoel Bock_ part of the page.

philipwhiuk:
This doesn't fix #2188 but it will prevent bad messages from being sent and should give a stack trace for why we are not setting the MIME Version (which we do by calling `MimeMessageHelper.setBody(message, body);`

(I'm currently looking at why exactly it happens).

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/message/MessageBuilder.java
+++ b/k9mail/src/main/java/com/fsck/k9/message/MessageBuilder.java
@@ -183,6 +183,8 @@ public abstract class MessageBuilder {
                 // No attachments to include, just stick the text body in the message and call it good.
                 MimeMessageHelper.setBody(message, body);
             }
+        } else {
+            throw new MessagingException("Unexpected message fomat: " + messageFormat);
         }
 
         // If this is a draft, add metadata for thawing.
--- a/k9mail/src/test/java/com/fsck/k9/message/MessageBuilderTest.java
+++ b/k9mail/src/test/java/com/fsck/k9/message/MessageBuilderTest.java
@@ -258,6 +258,14 @@ public class MessageBuilderTest {
         verifyNoMoreInteractions(anotherCallback);
     }
 
+    @Test(expected = MessagingException.class)
+    public void buildWithNullMessageFormat_shouldThrow() throws MessagingException {
+        MessageBuilder messageBuilder = createSimpleMessageBuilder();
+        messageBuilder.setMessageFormat(null);
+
+        messageBuilder.build();
+    }
+
     private MimeMessage getMessageFromCallback() {
         ArgumentCaptor<MimeMessage> mimeMessageCaptor = ArgumentCaptor.forClass(MimeMessage.class);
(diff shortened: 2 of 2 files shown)


philipwhiuk:
I don't think this is actually possible to hit and it didn't catch the issue.

PR #2192, opened 2017-02-05
Add encrypted.asc filename to PGP/MIME emails

dschuermann:
Following what Thunderbird and GPGTools does, I am naming the encrypted part "encrypted.asc".
Fixes https://github.com/k9mail/k-9/issues/2031


philipwhiuk:
This looks like what I got from reading the spec so LGTM 👍

PR #2206, opened 2017-02-08
Fix consecutive retrying of crypto operations

Valodim:
This PR fixes two small bugs with RetainFragment. Most importantly, a fragment could be returned if cleared and retrieved before FragmentManager had a chance to execute pending actions, leading to inconsistent state.


philipwhiuk:
Does this fix #1932?

Valodim:
Nope. I have another idea though, see over there

PR #2220, opened 2017-02-09
Handle unconfigured OpenPGP provider in MessageView

Valodim:
Building on top of the global-crypto-provider branch (#2130), this PR adds a dialog into message view when an encrypted message is encountered but no crypto provider is configured. This should make configuration seamless enough to mitigate lack of a migration from the account setting to the global setting.

Screenie:
shot_noprovider

What's notable is that this PR introduces a dialog activity for choosing a crypto provider that changes this setting while sidestepping the usual `Prefs` dialog. My first attempt was an extra for the Prefs activity to only show the dialog, but that turned out super messy - DialogPreferences just aren't made for the case.

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/activity/setup/OpenPgpAppSelectDialog.java
+++ b/k9mail/src/main/java/com/fsck/k9/activity/setup/OpenPgpAppSelectDialog.java
@@ -1,0 +1,276 @@
+package com.fsck.k9.activity.setup;
+
+
+import java.util.ArrayList;
+import java.util.List;
+
+import android.app.Activity;
+import android.app.AlertDialog;
+import android.app.Dialog;
+import android.app.DialogFragment;
+import android.content.Context;
+import android.content.DialogInterface;
+import android.content.Intent;
+import android.content.pm.ResolveInfo;
+import android.graphics.drawable.Drawable;
+import android.net.Uri;
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.ArrayAdapter;
+import android.widget.ListAdapter;
+import android.widget.TextView;
+
+import com.fsck.k9.K9;
+import com.fsck.k9.Preferences;
+import com.fsck.k9.R;
+import com.fsck.k9.preferences.StorageEditor;
+import com.fsck.k9.ui.dialog.ApgDeprecationWarningDialog;
+import org.openintents.openpgp.util.OpenPgpApi;
+import org.openintents.openpgp.util.OpenPgpAppPreference;
+
+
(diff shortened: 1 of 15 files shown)


k9mail/src/main/java/com/fsck/k9/activity/setup/OpenPgpAppSelectDialog.java
```diff
@@ -0,0 +1,294 @@
...
+import com.fsck.k9.preferences.StorageEditor;
+import com.fsck.k9.ui.dialog.ApgDeprecationWarningDialog;
+import org.openintents.openpgp.util.OpenPgpApi;
+import org.openintents.openpgp.util.OpenPgpAppPreference;
+
+
+/**
+ * Does not extend ListPreference, but is very similar to it!
+ * http://grepcode.com/file_/repository.grepcode.com/java/ext/com.google.android/android/4.4_r1/android/preference/ListPreference.java/?v=source
+ */
```
dschuermann: this comment is no longer true for this file.

k9mail/src/main/java/com/fsck/k9/activity/setup/OpenPgpAppSelectDialog.java
```diff
@@ -0,0 +1,294 @@
...
+        OpenPgpAppSelectFragment fragment = new OpenPgpAppSelectFragment();
+        fragment.show(getFragmentManager(), "openpgp_select");
+    }
+
+    private void showApgDeprecationDialogFragment() {
+        ApgDeprecationDialogFragment fragment = new ApgDeprecationDialogFragment();
+        fragment.show(getFragmentManager(), "apg_deprecation");
+    }
+
+    public static class OpenPgpAppSelectFragment extends DialogFragment {
```
dschuermann: Hmmmm, a lot of code duplication here with ``OpenPgpAppPreference``.
Can we make ``OpenPgpAppSelectFragment`` a single file and then let this activity and ``OpenPgpAppPreference`` use it? This should be possible. If required you can increase the minSdk of the library and move ``OpenPgpAppSelectFragment`` into it.

dschuermann: example: https://github.com/attenzione/android-ColorPickerPreference/tree/master/ColorPickerPreference/src/main/java/net/margaritov/preference/colorpicker

Valodim: You're right, it's mostly because this new code lives in k-9 where the old one was in the library. I didn't want to rewrite too much of the lib for no reason at this point, but we could move to using the activity as well from the preference later on.

PR #2228, opened 2017-02-12
Issue#2210 : Unit Tests for SmtpDataStuffing

shonramamurthy:
#2210 
I've noticed that k-9 mail replaces '.' (char 46) with '=2E' (char byte sequence 61, 50, 69). This eliminates the need for dot stuffing SMTP content (this is good in a way as it allows to stop relying on servers to perform dot removal correctly).
Let me know if more test cases need to be added.


shonramamurthy:
Thanks for your time and feedback. I'll get this fixed.

PR #2253, opened 2017-02-17
Dropped sig np2

ployt0:
Seems I have satisfied Cketti's requirement to fix my AS :) ?

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/crypto/MessageDecryptVerifier.java
+++ b/k9mail/src/main/java/com/fsck/k9/crypto/MessageDecryptVerifier.java
@@ -73,7 +73,7 @@ public class MessageDecryptVerifier {
         if (multipart.getCount() == 0) {
             return null;
         }
-
+        
         BodyPart firstBodyPart = multipart.getBodyPart(0);
 
         Part foundPart;
@@ -99,7 +99,7 @@ public class MessageDecryptVerifier {
             if (multipart.getCount() == 0) {
                 return null;
             }
-
+            
             BodyPart firstBodyPart = multipart.getBodyPart(0);
             if (isPartPgpInlineEncryptedOrSigned(firstBodyPart)) {
                 return firstBodyPart;
@@ -195,10 +195,8 @@ public class MessageDecryptVerifier {
     }
 
     /**
-     * @param part the part with signature data.
+     * @param part potentially with signature data.
      * @return may be null if no valid signature data found.
-     * @throws IOException
-     * @throws MessagingException
      */
     public static byte[] getSignatureData(Part part) throws IOException, MessagingException {
         if (isPartMultipartSigned(part)) {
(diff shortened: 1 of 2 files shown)


k9mail/src/main/java/com/fsck/k9/crypto/MessageDecryptVerifier.java
```diff
@@ -194,16 +194,22 @@ private static Part findPrimaryPartInAlternative(Part part) {
...
         if (isPartMultipartSigned(part)) {
             Body body = part.getBody();
             if (body instanceof Multipart) {
                 Multipart multi = (Multipart) body;
-                BodyPart signatureBody = multi.getBodyPart(1);
-                if (isSameMimeType(signatureBody.getMimeType(), APPLICATION_PGP_SIGNATURE)) {
-                    ByteArrayOutputStream bos = new ByteArrayOutputStream();
-                    signatureBody.getBody().writeTo(bos);
-                    return bos.toByteArray();
+                if (multi.getCount() > 1) {//Bug #2225 mailing list strips part out
```
Valodim: this comment isn't really going to be helpful some time from now, please drop it

k9mail/src/main/java/com/fsck/k9/ui/crypto/MessageCryptoHelper.java
```diff
@@ -344,18 +340,29 @@ public void onProgress(int current, int max) {
         });
     }
 
+    /**
+     * Common operations preceding {@link #onCryptoOperationReturned(MimeBodyPart)}
+     * @param result
+     * @param decryptedPart
+     */
+    private void onCryptoOpReturnedPreops(Intent result, MimeBodyPart decryptedPart) {
```
Valodim: `onCryptoOperationReturned` is never called otherwise, might as well move the contents of this method in there

k9mail/src/main/java/com/fsck/k9/ui/crypto/MessageCryptoHelper.java
```diff
@@ -344,18 +340,29 @@ public void onProgress(int current, int max) {
...
+        onCryptoOperationReturned(decryptedPart);
+    }
+
     private void callAsyncDetachedVerify(Intent intent) throws IOException, MessagingException {
         OpenPgpDataSource dataSource = getDataSourceForSignedData(currentCryptoPart.part);
 
         byte[] signatureData = MessageDecryptVerifier.getSignatureData(currentCryptoPart.part);
+        if (signatureData != null)
+            intent.putExtra(OpenPgpApi.EXTRA_DETACHED_SIGNATURE, signatureData);
         intent.putExtra(OpenPgpApi.EXTRA_DETACHED_SIGNATURE, signatureData);
```
Valodim: This line and the previous make no sense. Also, did you check what actually happens when this parameter is null?

Valodim:
Looking at this again, it doesn't solve the actual problem. The null value handed to OpenKeychain will probably crashes there or return an error at best. We should be able to figure out the format isn't right earlier and not even attempt this operation, adding an error annotation instead.

PR #2273, opened 2017-02-22
Playground draggable

rhari991:
Should work now :)

drag_and_drop


rhari991:
@Valodim please review :)

rhari991:
@Valodim added a gif

Valodim:
Tested this, I found two problems: firstly, the CC and BCC fields should expand when a drag starts, so that they are available as targets. secondly, clicking a token opens the dropdown now as before but then immediately removes the token. this latter problem should have been obvious during testing, does this only happen for me?

k9mail/src/main/java/com/fsck/k9/view/RecipientSelectView.java
```diff
@@ -731,4 +740,8 @@ private void readObject(ObjectInputStream ois) throws ClassNotFoundException, IO
             }
         }
     }
+
+    public interface DragListener {
+        void beforeDragStart();
```
Valodim: should have on* prefix

k9mail/src/main/java/com/fsck/k9/activity/compose/RecipientMvpView.java
```diff
@@ -149,6 +155,12 @@ public void onTokenChanged(Recipient recipient) {
         });
     }
 
+    @Override
+    public void beforeDragStart() {
+        View recipientExpander = activity.findViewById(R.id.recipient_expander);
+        recipientExpander.performClick();
```
Valodim: this logic should be in RecipientPresenter. see `setPresenter` for how this is handled in other cases, it should be simple to forward the onDrag events to the presenter, and handle visibility changes from there.

rhari991: @Valodim done

Valodim:
*almost* nice now! I noticed two things:
1) Moving the token over the subject field collapses CC/BCC, it shouldn't
2) Dropping a token outside of the recipient views probably shouldn't delete the token. For example, if a user drags and drops a token from the recipient view into the message body, they probably don't want to delete the token but just copy the address there.

rhari991:
@Valodim Made the required changes.

Valodim:
Moving a token from one recipient field to another now copies it, moving it over the Subject or Body still collapses the Cc/Bcc. Did you test this?

Valodim:
I rebased this with some changes into https://github.com/k9mail/k-9/tree/playground-draggable, could you open a new PR from there? thanks!

PR #2274, opened 2017-02-23
Allow disable of default encryption

Valodim:
Picking up #2087. This is rebased on current master, and saves the default mode in an enum to allow for different settings in the future without requiring migration.

*However* I might want to do this with a different mechanism, please don't merge this for now.


philipwhiuk:
What happened to this?

Valodim:
Well, we don't encrypt by default anymore, so there is little point to a setting that disables encryption by default. The "different mechanism" I mentioned above turned out to be Autocrypt :)

PR #2299, opened 2017-02-28
Added up button to MessageCompose Activity

saurabhkumar13:
_Added a back button to Action bar which acts as an Up button and open Accounts list Activity._
#2238 
I was conflicted between unified inbox activity and accounts, but current order as now is also confusing, Inbox sends to Folders, Folder sends to Accounts List.

As it was stated all Up buttons should go to the same activity, going to Accounts activity made more sense to me as it acts like the root activity.


philipwhiuk:
We want it to go to the Message List per @cketti's comments. In the future the 'Accounts List' will probably not be an activity in of itself.

I think it should probably be the Message List for the account from which the email is being sent.

PR #2306, opened 2017-03-01
Added 20 best pseudonyms for, and the provider itself, mail.com

ployt0:


cketti:
Is there a reason the mail.com entry is not grouped with the others?

ployt0:
Only because it is not an alias for Mail.com, that is the domain which serves all of its aliases.

cketti:
That doesn't sound like an important distinction :)

ployt0:
I missed that pattern, way better than grouping it with USA

cketti:
Thanks.

ployt0:
Er, okay, is this the fabled limbo where a PR awaits merging with master or is there something wrong still?

cketti:
I squashed the commits and merged the changes from this pull request. You can see that the commit b99b539c20328ca7e90d0f260607518585f06952 closed this issue.

github_pull_request_closed_via_commit

Valodim:
Just for the record, changes like this preserve ownership of the original commits, your changes are now in 7af7d8a3d781f8a9e4863cb40b26ebafd632a83b

PR #2312, opened 2017-03-02
Add folder selection in unread widget

rhari991:
For #1813

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/Account.java
+++ b/k9mail/src/main/java/com/fsck/k9/Account.java
@@ -821,6 +821,50 @@ public class Account implements BaseAccount, StoreConfig {
         return stats;
     }
 
+    public int getFolderUnreadCount(Context context, String folderName) throws MessagingException {
+        if (!isAvailable(context)) {
+            return 0;
+        }
+
+        int unreadMessageCount = 0;
+
+        Cursor cursor = loadUnreadCountForFolder(context, folderName);
+        try {
+            if (cursor != null && cursor.moveToFirst()) {
+                unreadMessageCount = cursor.getInt(0);
+            }
+        } finally {
+            Utility.closeQuietly(cursor);
+        }
+
+        return unreadMessageCount;
+    }
+
+    private Cursor loadUnreadCountForFolder(Context context, String folderName) {
+        ContentResolver cr = context.getContentResolver();
+
+        Uri uri = Uri.withAppendedPath(EmailProvider.CONTENT_URI,
+                "account/" + getUuid() + "/stats");
+
+        String[] projection = {
+                StatsColumns.UNREAD_COUNT,
+        };
+
+        LocalSearch search = new LocalSearch();
+        search.addAllowedFolder(folderName);
+
+        // Use the LocalSearch instance to create a WHERE clause to query the content provider
+        StringBuilder query = new StringBuilder();
+        List<String> queryArgs = new ArrayList<>();
+        ConditionsTreeNode conditions = search.getConditions();
+        SqlQueryBuilder.buildWhereClause(this, conditions, query, queryArgs);
+
+        String selection = query.toString();
+        String[] selectionArgs = queryArgs.toArray(new String[queryArgs.size()]);
+
+        return cr.query(uri, projection, selection, selectionArgs, null);
+    }
+
 
     public synchronized void setChipColor(int color) {
         chipColor = color;
--- a/k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
+++ b/k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
@@ -126,7 +126,13 @@ public class UnreadWidgetConfiguration extends K9PreferenceActivity {
     }
 
     private void handleChooseAccount(String accountUuid) {
+        boolean userSelectedSameAccount = accountUuid.equals(selectedAccountUuid);
+        if (userSelectedSameAccount) {
+            return;
+        }
+
         selectedAccountUuid = accountUuid;
+        selectedFolderName = null;
         unreadFolder.setSummary(getString(R.string.unread_widget_folder_summary));
         if (SearchAccount.UNIFIED_INBOX.equals(selectedAccountUuid) ||
                 SearchAccount.ALL_MESSAGES.equals(selectedAccountUuid)) {
--- a/k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
+++ b/k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
@@ -6,32 +6,55 @@ import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.SharedPreferences.Editor;
 import android.os.Bundle;
+import android.preference.CheckBoxPreference;
+import android.preference.Preference;
+import android.view.Menu;
+import android.view.MenuItem;
+import android.widget.Toast;
 
-import com.fsck.k9.BaseAccount;
+import com.fsck.k9.Account;
+import com.fsck.k9.Preferences;
 import com.fsck.k9.R;
+import com.fsck.k9.helper.UnreadWidgetProperties;
 import com.fsck.k9.provider.UnreadWidgetProvider;
+import com.fsck.k9.search.SearchAccount;
+
+import timber.log.Timber;
 
 
 /**
  * Activity to select an account for the unread widget.
  */
-public class UnreadWidgetConfiguration extends AccountList {
+public class UnreadWidgetConfiguration extends K9PreferenceActivity {
     /**
      * Name of the preference file to store the widget configuration.
      */
     private static final String PREFS_NAME = "unread_widget_configuration.xml";
 
     /**
-     * Prefix for the preference keys.
+     * Prefixes for the preference keys
      */
     private static final String PREF_PREFIX_KEY = "unread_widget.";
+    private static final String PREF_FOLDER_NAME_SUFFIX_KEY = ".folder_name";
 
+    private static final String PREFERENCE_UNREAD_ACCOUNT = "unread_account";
+    private static final String PREFERENCE_UNREAD_FOLDER_ENABLED = "unread_folder_enabled";
+    private static final String PREFERENCE_UNREAD_FOLDER = "unread_folder";
 
     /**
      * The ID of the widget we are configuring.
      */
-    private int mAppWidgetId = AppWidgetManager.INVALID_APPWIDGET_ID;
+    private int appWidgetId = AppWidgetManager.INVALID_APPWIDGET_ID;
 
+    private static final int REQUEST_CHOOSE_ACCOUNT = 1;
+    private static final int REQUEST_CHOOSE_FOLDER = 2;
+
+    private Preference unreadAccount;
+    private CheckBoxPreference unreadFolderEnabled;
+    private Preference unreadFolder;
+
+    private String selectedAccountUuid;
+    private String selectedFolderName;
 
     @Override
     public void onCreate(Bundle icicle) {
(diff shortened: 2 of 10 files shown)


rhari991:
I have updated the UI to a preference screen.
unread_widget_folder_option

k9mail/src/main/java/com/fsck/k9/Account.java
```diff
@@ -820,6 +820,45 @@ public AccountStats getStats(Context context) throws MessagingException {
...
+        ContentResolver cr = context.getContentResolver();
+
+        Uri uri = Uri.withAppendedPath(EmailProvider.CONTENT_URI,
+                "account/" + getUuid() + "/stats");
+
+        String[] projection = {
+                StatsColumns.UNREAD_COUNT,
+        };
+
+        LocalSearch search = new LocalSearch();
```
philipwhiuk: replace this whole block with a function that explains intent:

`buildUnreadQueryForFolder(folderName)`

k9mail/src/main/java/com/fsck/k9/Account.java
```diff
@@ -820,6 +820,45 @@ public AccountStats getStats(Context context) throws MessagingException {
...
+        search.addAllowedFolder(folderName);
+
+        // Use the LocalSearch instance to create a WHERE clause to query the content provider
+        StringBuilder query = new StringBuilder();
+        List<String> queryArgs = new ArrayList<>();
+        ConditionsTreeNode conditions = search.getConditions();
+        SqlQueryBuilder.buildWhereClause(this, conditions, query, queryArgs);
+
+        String selection = query.toString();
+        String[] selectionArgs = queryArgs.toArray(new String[0]);
```
philipwhiuk: `new String[queryArgs.size()]` prevents Java having to create a pointless array.

k9mail/src/main/java/com/fsck/k9/provider/UnreadWidgetProvider.java
```diff
@@ -47,7 +47,7 @@ public static void updateUnreadCount(Context context) {
     }
 
     public static void updateWidget(Context context, AppWidgetManager appWidgetManager,
-            int appWidgetId, String accountUuid) {
+            int appWidgetId, String accountUuid, String folderName) {
 
         RemoteViews remoteViews = new RemoteViews(context.getPackageName(),
```
philipwhiuk: This method is getting longer and longer and it's difficult to see how it could be split.

Maybe we have to accept some small code duplication and split it into:

```
updateWidgetWithSearchAccount()
updateWidgetWithAccount()
updateWidgetWithFolder()
```

Either that or build a structs or something that can be modified by methods and mean we don't have to pass round lots of properties. Maybe:

`class UnreadWidgetProperties {
   unreadCount;
   title;
   clickIntent;
}`

Just some thoughts. Let me know what you think.

k9mail/src/main/java/com/fsck/k9/provider/UnreadWidgetProvider.java
```diff
@@ -115,7 +124,11 @@ public static void updateWidget(Context context, AppWidgetManager appWidgetManag
             remoteViews.setTextViewText(R.id.unread_count, displayCount);
         }
 
-        remoteViews.setTextViewText(R.id.account_name, accountName);
+        if (folderName != null) {
+            remoteViews.setTextViewText(R.id.account_name, accountName + "/" + folderName);
```
philipwhiuk: 1. This should use a format string: getString(R.string.unread_widget_fmt, accountName, folderName);
2. `R.id.accout_name` is becoming an increasingly stupid name for this property. I think it should probably just be title. It can be:
* The app name
* An account name
* The account name followed by the folder name.

If you just call it title you can move this if-clause back into the folder handling stuff above.

k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
```diff
@@ -41,58 +61,153 @@ public void onCreate(Bundle icicle) {
...
         if (extras != null) {
-            mAppWidgetId = extras.getInt(AppWidgetManager.EXTRA_APPWIDGET_ID,
+            appWidgetId = extras.getInt(AppWidgetManager.EXTRA_APPWIDGET_ID,
                     AppWidgetManager.INVALID_APPWIDGET_ID);
         }
 
         // If they gave us an intent without the widget ID, just bail.
-        if (mAppWidgetId == AppWidgetManager.INVALID_APPWIDGET_ID) {
+        if (appWidgetId == AppWidgetManager.INVALID_APPWIDGET_ID) {
             finish();
```
philipwhiuk: We might want to log this case.

k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
```diff
@@ -41,58 +61,153 @@ public void onCreate(Bundle icicle) {
...
         setTitle(R.string.unread_widget_select_account);
     }
 
     @Override
-    protected boolean displaySpecialAccounts() {
+    public void onActivityResult(int requestCode, int resultCode, Intent data) {
+        if (resultCode == RESULT_OK) {
+            switch (requestCode) {
+                case REQUEST_CHOOSE_ACCOUNT:
+                    selectedAccountUuid = data.getStringExtra(ChooseAccount.EXTRA_ACCOUNT_UUID);
```
philipwhiuk: move this stuff into a private helper function - it's getting kind of long

k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
```diff
@@ -41,58 +61,153 @@ public void onCreate(Bundle icicle) {
...
+        switch (item.getItemId()) {
+            case R.id.done:
+                checkToCreateWidget();
+                return true;
+            default:
+                return super.onOptionsItemSelected(item);
+        }
+    }
+
+    private void checkToCreateWidget() {
```
philipwhiuk: `validateThenUpdateWidgetThenExitIfPossible()` would be more accurate - you're doing more than just checking here.

Alternatively, make this method return a value and then do


```
if(validateWidget()) {
   updateWidgetAndExit();
}
```

in `onOptionsItemSelected`

k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
```diff
@@ -41,58 +61,153 @@ public void onCreate(Bundle icicle) {
...
         setResult(RESULT_OK, resultValue);
         finish();
     }
 
     private static void saveAccountUuid(Context context, int appWidgetId, String accountUuid) {
         SharedPreferences.Editor editor =
                 context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE).edit();
-        editor.putString(PREF_PREFIX_KEY + appWidgetId, accountUuid);
+        editor.putString(PREF_ACCOUNT_UUID_PREFIX_KEY + appWidgetId, accountUuid);
+        editor.commit();
```
philipwhiuk: Prefer apply()

k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
```diff
@@ -41,58 +61,153 @@ public void onCreate(Bundle icicle) {
...
+    }
+
+    private static void saveFolderName(Context context, int appWidgetId, String folderName) {
+        SharedPreferences.Editor editor =
+                context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE).edit();
+        editor.putString(PREF_FOLDER_NAME_PREFIX_KEY + appWidgetId, folderName);
         editor.commit();
```
philipwhiuk: Switch this to apply() too.

k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
```diff
@@ -41,58 +61,153 @@ public void onCreate(Bundle icicle) {
...
+        SharedPreferences prefs = context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE);
+        String folderName = prefs.getString(PREF_FOLDER_NAME_PREFIX_KEY + appWidgetId, null);
+        return folderName;
+    }
+
     public static void deleteWidgetConfiguration(Context context, int appWidgetId) {
         Editor editor = context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE).edit();
-        editor.remove(PREF_PREFIX_KEY + appWidgetId);
+        editor.remove(PREF_ACCOUNT_UUID_PREFIX_KEY + appWidgetId);
+        editor.remove(PREF_FOLDER_NAME_PREFIX_KEY + appWidgetId);
```
philipwhiuk: And this one

rhari991:
@philipwhiuk done ! Sorry for the long wait.

k9mail/src/main/java/com/fsck/k9/provider/UnreadWidgetProvider.java
rhari991: I like the structs idea, so I've implemented that. It contains the logic for calculating the clickIntent, unreadCount and title.

k9mail/src/main/java/com/fsck/k9/provider/UnreadWidgetProvider.java
rhari991: I've changed the id to title, but I'm not sure I understand the rest. Do you mean to allow the user to modify the title of the widget ?

philipwhiuk:
@cketti @Valodim Happy to merge this?

k9mail/src/main/java/com/fsck/k9/activity/UnreadWidgetConfiguration.java
```diff
@@ -41,58 +64,164 @@ public void onCreate(Bundle icicle) {
...
         // Let the caller know that the configuration was successful
         Intent resultValue = new Intent();
-        resultValue.putExtra(AppWidgetManager.EXTRA_APPWIDGET_ID, mAppWidgetId);
+        resultValue.putExtra(AppWidgetManager.EXTRA_APPWIDGET_ID, appWidgetId);
         setResult(RESULT_OK, resultValue);
         finish();
     }
 
-    private static void saveAccountUuid(Context context, int appWidgetId, String accountUuid) {
+    private static void saveWidgetProperties(Context context, UnreadWidgetProperties properties) {
```
cketti: Changing the format of the persisted widget configuration without updating existing data will lead to widgets created with previous app versions to stop working.
Right now we have no convenient place where such migration code could live. I propose avoiding migration by changing the format in a way that doesn't invalidate old configurations:

* `unread_widget.<widgetId>` stores the account UUID like before
* `unread_widget.<widgetId>.folderName` stores the folder name

k9mail/src/main/java/com/fsck/k9/helper/UnreadWidgetProperties.java
```diff
@@ -0,0 +1,140 @@
...
+
+    public String getTitle(Context context) {
+        String accountName = getAccount(context).getDescription();
+        switch (type) {
+            case SEARCH_ACCOUNT:
+            case ACCOUNT:
+                return accountName;
+            case FOLDER:
+                return context.getString(R.string.unread_widget_title, accountName, folderName);
+            default:
```
cketti: No need to change this now.

The `default` case is unnecessary. The way to make the compiler happy, catch missing cases at runtime, and to make it possible to catch missing cases by static analysis, is to change the code to:

```java
switch (type) {
    …
}

throw new IllegalStateException("Missing case: " + type);
```

k9mail/src/main/java/com/fsck/k9/helper/UnreadWidgetProperties.java
```diff
@@ -0,0 +1,140 @@
...
+        if (SearchAccount.UNIFIED_INBOX.equals(accountUuid)) {
+            return SearchAccount.createUnifiedInboxAccount(context);
+        } else if (SearchAccount.ALL_MESSAGES.equals(accountUuid)) {
+            return SearchAccount.createAllMessagesAccount(context);
+        }
+        return Preferences.getPreferences(context).getAccount(accountUuid);
+    }
+
+    private Intent getClickIntentForAccount(Context context) {
+        Account account = (Account) getAccount(context);
```
cketti: At this point we know it's a regular account. Better use `Preferences.getAccount()` directly to avoid the unnecessary work in this class's `getAccount()` and the unsafe type cast. Same in the method below.

rhari991:
@cketti done

k9mail/src/main/res/values/strings.xml
```diff
@@ -1019,6 +1019,8 @@ Please submit bug reports, contribute new features and ask questions at
     <string name="unread_widget_select_account">Show unread count for…</string>
     <string name="unread_widget_account_title">Account</string>
     <string name="unread_widget_account_summary">The account for which the unread count should be displayed</string>
+    <string name="unread_widget_unified_inbox_account_summary">Unified Inbox</string>
```
cketti: What's the reason for not using the existing strings for "Unified Inbox" and "All messages"?

rhari991: They were in different contexts

rhari991: @cketti removed the line.

cketti:
Rebased, squashed, and fixed a tiny UX bug.

cketti:
Thanks @rhari991

PR #2313, opened 2017-03-02
Line wrap fix and unit test

danwain:
Unit test for LineWrapOutputStream.
Also found possible fix where a space causes it to find an earlier word boundary when not needed.

This is issue #2209

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/filter/LineWrapOutputStream.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/filter/LineWrapOutputStream.java
@@ -22,8 +22,9 @@ public class LineWrapOutputStream extends FilterOutputStream {
     public void write(int oneByte) throws IOException {
         // Buffer full?
         if (lineLength == buffer.length) {
+            // Can skip the char if end of word
             // Usable word-boundary found earlier?
-            if (endOfLastWord > 0) {
+            if (endOfLastWord > 0 && oneByte != ' ') {
                 // Yes, so output everything up to that word-boundary
                 out.write(buffer, bufferStart, endOfLastWord - bufferStart);
                 out.write(CRLF);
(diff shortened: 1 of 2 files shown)


danwain:
I found a similar issue with the LineWrapOutputStream where it also trys to find an airier boundary if it receives an end of line character, however this would not have an elegant solution.

ployt0:
What was wrong with the tests I last resubmitted 2 days ago? Have you tested the wider implications, in dependent classes, of utilising the extra space I identified? It is only 1 space in 1000 so hardly worth the effort IMHO.

danwain:
Sorry @ployt0 I was a bit over eager and I didn't see your code, I only did a quick search in the pull requests.

danwain:
Yeah the single space probably isn't worth it but a new line might be as it could possibly change the way text if formatted by putting in an extra line where there shouldn't be.

ployt0:
No worries, I did similar myself, but I have naff internet is my excuse, and I'm new too. I believe your fix is a good one.

danwain:
Your test does look a bit more extensive. Also to deal with the new line issue might mean rewritting the method or another cheap hack which probabily isn't a good idea

danwain:
A possible solution would be to store a few extra characters beyond the end of line then do the split after?

ployt0:
I saw this issue as a request for tests, I didn't want to break what is working well enough. Fair point though, my tests do highlight the inefficiency, perhaps if you would write one to demonstrate the unwanted new line you mentioned the fix would be of greater merit?

danwain:
I've updated my tester to show where this is the case.
Note: it will fail

danwain:
There is a possible solution. It seems to pass all the unit tests but I'm not sure what possible side effects it could have as I'm not familiar with the code base. However I don't see why it would be an issue

ployt0:
You seem better with git than myself but if you could break the test into a minimal proof of concept, with an individual name I am sure it would help. The wrapping appears concerned with the SMTP protocol so I am loathe to touch it without reading the RFC.

danwain:
I don't think this class is needed at all #2332, so I will close this pull request.

PR #2264, opened 2017-02-21
search nicknames (issue 2240)

daquexian:
PR for issue #2240 

According to the [document](https://developer.android.com/guide/topics/providers/contacts-provider.html), nicknames and email addresses are stored separately in different rows. So I query the `CONTACT_ID` at first and query email address according to `CONTACT_ID`

PR #502, opened 2014-09-30
add gravatar support - closes #4876

ligi:
add gravatar support - about the indent changes in there - I was running astyle with the parameters from:
https://github.com/k9mail/k-9/wiki/CodeStyle - somehow this style is not really enforced - I had a lot of reformatted files after running:

```
➜  k-9 git:(master) ✗ astyle --style=java --indent=spaces=4 --brackets=attach --convert-tabs   --unpad-paren --pad-header --pad-oper --suffix=none --recursive "src/com/fsck/k9/*.java"
------------------------------------------------------------
Directory  src/com/fsck/k9/*.java
------------------------------------------------------------
Formatted  Account.java
Unchanged  AccountStats.java
Unchanged  BaseAccount.java
Unchanged  Clock.java
Unchanged  EmailAddressAdapter.java
Unchanged  EmailAddressValidator.java
Unchanged  FontSizes.java
Unchanged  Identity.java
Formatted  K9.java
Unchanged  NotificationSetting.java
Formatted  PRNGFixes.java
Unchanged  Preferences.java
Formatted  Throttle.java
Unchanged  activity/AccountList.java
Formatted  activity/Accounts.java
Formatted  activity/ActivityListener.java
Formatted  activity/ChooseFolder.java
Unchanged  activity/ChooseIdentity.java
Formatted  activity/ColorPickerDialog.java
Unchanged  activity/ConfirmationDialog.java
Unchanged  activity/EditIdentity.java
Unchanged  activity/EmailAddressList.java
Formatted  activity/FolderInfoHolder.java
Formatted  activity/FolderList.java
Unchanged  activity/FolderListFilter.java
Formatted  activity/InsertableHtmlContent.java
Unchanged  activity/K9Activity.java
Formatted  activity/K9ActivityCommon.java
Unchatted  activity/MessageList.java
Unchanged  activity/MessageReference.java
Formatted  activity/NotificationDeleteConfirmation.java
Unchanged  activity/Search.java
Formatted  activity/UnreadWidgetConfiguration.java
Formatted  activity/UpgradeDatabases.java
Unchanged  activity/loader/AttachmentContentLoader.java
Formatted  activity/loader/AttachmentInfoLoader.java
Formatted  activity/misc/Attachment.java
Formatted  activity/misc/ContactPictureLoader.java
Formatted  activity/misc/ExtendedAsyncTask.java
Unchanged  activity/misc/NonConfigurationInstance.java
Formatted  activity/misc/SwipeGestureDetector.java
[...]

PR #2315, opened 2017-03-02
Load all contact pictures by Glide

daquexian:
For issue #2057 

Sorry for my carelessness on my last PR.

Use the way in [this](http://stackoverflow.com/a/31940058) to load bitmap by Glide (and the cache of Glide still works). So the contact's photo and the fallback pictures are all loaded by Glide now.

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/activity/misc/ContactPictureLoader.java
+++ b/k9mail/src/main/java/com/fsck/k9/activity/misc/ContactPictureLoader.java
@@ -1,34 +1,43 @@
 package com.fsck.k9.activity.misc;
 
-import java.io.FileNotFoundException;
+
 import java.io.IOException;
-import java.io.InputStream;
-import java.lang.ref.WeakReference;
 import java.util.Locale;
-import java.util.concurrent.RejectedExecutionException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import android.app.ActivityManager;
-import android.content.ContentResolver;
 import android.content.Context;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
-import android.graphics.BitmapFactory;
 import android.graphics.Canvas;
 import android.graphics.Paint;
+import android.graphics.Paint.Style;
 import android.graphics.Rect;
-import android.graphics.drawable.BitmapDrawable;
-import android.graphics.drawable.Drawable;
 import android.net.Uri;
-import android.os.AsyncTask;
 import android.support.annotation.VisibleForTesting;
-import android.support.v4.util.LruCache;
 import android.text.TextUtils;
 import android.widget.ImageView;
 
+import com.bumptech.glide.Glide;
+import com.bumptech.glide.Priority;
+import com.bumptech.glide.load.ResourceDecoder;
+import com.bumptech.glide.load.data.DataFetcher;
+import com.bumptech.glide.load.engine.DiskCacheStrategy;
+import com.bumptech.glide.load.engine.Resource;
+import com.bumptech.glide.load.engine.bitmap_recycle.BitmapPool;
+import com.bumptech.glide.load.model.ModelLoader;
+import com.bumptech.glide.load.resource.bitmap.BitmapEncoder;
+import com.bumptech.glide.load.resource.bitmap.BitmapResource;
+import com.bumptech.glide.load.resource.bitmap.StreamBitmapDecoder;
+import com.bumptech.glide.load.resource.drawable.GlideDrawable;
+import com.bumptech.glide.load.resource.file.FileToStreamDecoder;
+import com.bumptech.glide.load.resource.transcode.BitmapToGlideDrawableTranscoder;
+import com.bumptech.glide.request.RequestListener;
+import com.bumptech.glide.request.target.Target;
 import com.fsck.k9.helper.Contacts;
 import com.fsck.k9.mail.Address;
+import com.fsck.k9.view.RecipientSelectView.Recipient;
+
 
 public class ContactPictureLoader {
     /**
@@ -47,17 +56,11 @@ public class ContactPictureLoader {
     private static final String FALLBACK_CONTACT_LETTER = "?";
 
 
-    private ContentResolver mContentResolver;
     private Resources mResources;
     private Contacts mContactsHelper;
     private int mPictureSizeInPx;
 
     private int mDefaultBackgroundColor;
-
-    /**
-     * LRU cache of contact pictures.
-     */
-    private final LruCache<Address, Bitmap> mBitmapCache;
 
     /**
      * @see <a href="http://developer.android.com/design/style/color.html">Color palette used</a>
@@ -80,7 +83,6 @@ public class ContactPictureLoader {
         String letter = null;
         String personal = address.getPersonal();
         String str = (personal != null) ? personal : address.getAddress();
-
         Matcher m = EXTRACT_LETTER_PATTERN.matcher(str);
         if (m.find()) {
             letter = m.group(0).toUpperCase(Locale.US);
@@ -101,7 +103,6 @@ public class ContactPictureLoader {
      */
     public ContactPictureLoader(Context context, int defaultBackgroundColor) {
         Context appContext = context.getApplicationContext();
-        mContentResolver = appContext.getContentResolver();
         mResources = appContext.getResources();
         mContactsHelper = Contacts.getInstance(appContext);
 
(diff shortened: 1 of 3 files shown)


daquexian:
Does `recipient.photoThumbnailUri` in [this](https://github.com/k9mail/k-9/pull/2315/files#diff-b6f7f0fc0a12bc28752cc904a2c6fe84R165) represent the same value as `contactsHelper.getPhotoUri(String)` in [this](https://github.com/k9mail/k-9/pull/2315/files#diff-b6f7f0fc0a12bc28752cc904a2c6fe84R117) when calling `loadContactPicture`?(I didn't find enough information about `ContactsContract.Contacts.PHOTO_THUMBNAIL_URI`) If so the code could be simplified.

philipwhiuk:
Quite possibly. I think the core problem is that both com.fsck.k9.helper.Contacts and com.fsck.k9.activity.compose.RecipientLoader query the Contacts system.

RecipientLoader ought to be using `com.fsck.k9.helper.Contacts` (which is designed to interface between K-9 and Android contacts) instead of accessing  `Contacts.lookupContact` directly.

daquexian:
@philipwhiuk Thanks for your suggestion! I will work on it later.

daquexian:
@philipwhiuk @cketti I have made some changes roughly in the commit af27f7b :) It is not complete and I just want to ask for some suggestions about whether I am in the right way.
I created a class `ContactCursor` in `Contacts`. The `ContactCursor` class wraps `Cursor` and exposes some functions for retrieving values and so on.

And since it seems not a minor change to move all operations into `Contacts`, should I do it in another branch, make this PR only related to Glide and clean the git history?

Valodim:
I like the idea of this, and moving database specifics out of RecipientLoader is also a good idea. Does this interfere with #2264 a lot?

daquexian:
@Valodim @philipwhiuk I moved commits about database to another branch and I would like to fix the issues #2057 and #2337 firstly :) I have made more progress on these two issues and force updated this branch. Now I think these two issues are all resolved.

daquexian:
What I want to point out is that the original `getPhotoUri` method does not actually work correctly. For example it usually returns a URI rather than null when the contact doesn't have a photo. The code of this method is from [this](http://stackoverflow.com/a/7738363), which is also said wrong by a comment.

I rewrote this method by using [PHOTO_URI](https://developer.android.com/reference/android/provider/ContactsContract.ContactsColumns.html#PHOTO_URI), which is also the value we adopt for `recipient.photoThumbnailUri`

k9mail/src/main/java/com/fsck/k9/activity/misc/ContactPictureLoader.java
```diff
@@ -110,59 +111,70 @@ public ContactPictureLoader(Context context, int defaultBackgroundColor) {
...
+                }
+            };
+
+            Glide.with(imageView.getContext())
+                    .load(photoUri)
+                    .diskCacheStrategy(DiskCacheStrategy.NONE)
+                    .listener(noPhotoListener)
+                    // for some reason, following 2 lines fix loading issues.
+                    .dontAnimate()
+                    .override(mPictureSizeInPx, mPictureSizeInPx)
```
daquexian: I don't fully understand the original role of `mPictureSizeInPx`, but I think this value could be used directly in it.

k9mail/src/main/java/com/fsck/k9/helper/Contacts.java
```diff
@@ -240,30 +238,13 @@ public Uri getPhotoUri(String address) {
                 if (!c.moveToFirst()) {
                     return null;
                 }
-
-                contactId = c.getLong(CONTACT_ID_INDEX);
+                final String uriString = c.getString(c.getColumnIndex(Photo.PHOTO_URI));
+                return Uri.parse(uriString);
+            } catch (IllegalStateException e) {
```
Valodim: when does this happen, and why do we have to catch it?

k9mail/src/main/java/com/fsck/k9/activity/misc/ContactPictureLoader.java
```diff
@@ -110,59 +111,70 @@ public ContactPictureLoader(Context context, int defaultBackgroundColor) {
...
+                .diskCacheStrategy(DiskCacheStrategy.NONE)
+                .load(new FallbackGlideParams(address))
+                // for some reason, following 2 lines fix loading issues.
+                .dontAnimate()
+                .override(mPictureSizeInPx, mPictureSizeInPx)
+                .into(imageView);
+    }
+
+    public void loadContactPicture(final Recipient recipient, final ImageView imageView) {
+        if (recipient.photoThumbnailUri != null) {
```
Valodim: why not simply use `loadContactPicture(recipient.photoThumbnailUri, imageView)`?

k9mail/src/main/java/com/fsck/k9/activity/misc/ContactPictureLoader.java
```diff
@@ -202,146 +208,73 @@ private Bitmap calculateFallbackBitmap(Address address) {
...
-    private ContactPictureRetrievalTask getContactPictureRetrievalTask(ImageView imageView) {
-        if (imageView != null) {
-           Drawable drawable = imageView.getDrawable();
-           if (drawable instanceof AsyncDrawable) {
-               AsyncDrawable asyncDrawable = (AsyncDrawable) drawable;
-               return asyncDrawable.getContactPictureRetrievalTask();
-           }
+        @Override
+        public String getId() {
+            return "com.fsck.k9";
```
Valodim: something like "fallback-photo" is probably a better fit here

k9mail/src/main/java/com/fsck/k9/activity/misc/ContactPictureLoader.java
```diff
@@ -110,59 +111,70 @@ public ContactPictureLoader(Context context, int defaultBackgroundColor) {
...
-                (ActivityManager) appContext.getSystemService(Context.ACTIVITY_SERVICE);
-        int memClass = activityManager.getMemoryClass();
+    }
 
-        // Use 1/16th of the available memory for this memory cache.
-        final int cacheSize = 1024 * 1024 * memClass / 16;
+    public void loadContactPicture(final Address address, final ImageView imageView) {
+        Uri photoUri = mContactsHelper.getPhotoUri(address.getAddress());
+        if (photoUri != null) {
+            final RequestListener<Uri, GlideDrawable> noPhotoListener = new RequestListener<Uri, GlideDrawable>() {
```
Valodim: no need for this var to be final

k9mail/src/main/java/com/fsck/k9/helper/Contacts.java
daquexian: It will happen when the contact have no photo. Maybe there is no need to catch it explicitly, for it will be caught by [the catch below](https://github.com/k9mail/k-9/pull/2315/files/b13dc31555d20a88b3f560058ee07c3a5c781c2e#diff-2bd769df7e1b99efc5dd476c555801fbL267). However, what I concerned is in the original code the [error log](https://github.com/k9mail/k-9/pull/2315/files/b13dc31555d20a88b3f560058ee07c3a5c781c2e#diff-2bd769df7e1b99efc5dd476c555801fbR249) seems only show when something unexpected happens. So I added this catch to avoid the error log. What do you think of it?

Valodim: sounds good

k9mail/src/main/java/com/fsck/k9/activity/misc/ContactPictureLoader.java
daquexian: Do you mean I could change the signature of this method to `loadContactPicture(Uri, ImageView)`?

daquexian: Could you please explain more detail?

Valodim: this method seems to do roughly the same thing as loadContactPicture called with an Address argument, instead of Recipient. wouldn't it be easier to just call that method here? or is there an important semantic difference between the two?

daquexian: You are right.. The two method almost do the same thing, but I think calling `loadContactPicture(Address, ImageView)` directly will cause an extra query for photo URI and waste time.

Is it better to extract a method `loadContactPicture(Uri, Address, ImageView)` which loads picture from the URI and calls `loadFallbackPicture(Address, imageView)` when the URI is null or so on, and simply call it in both `loadContactPicture(Recipient, ImageView)` and `loadContactPicture(Address, ImageView)`? Like this:

```
public void loadContactPicture(Address address, ImageView imageView) {
    Uri photoUri = mContactsHelper.getPhotoUri(address.getAddress());
    loadContactPicture(photoUri, address, imageView);
}

public void loadContactPicture(Recipient recipient, ImageView imageView) {
    loadContactPicture(recipient.photoThumbnailUri, recipient.address, imageView);
}
```

Valodim: sounds good

daquexian:
@Valodim I have made the required changes. For #502, I implemented gravatar and add the corresponding option in settings, but I think it may take some time to let this part get reviewed and approved. Could you please merge this PR if you approved it, and let me open another PR for gravatar then?

daquexian:
@Valodim Thanks! I have squashed the commits into one.

Valodim:
thanks! :+1:

Issue #1232, opened 2016-03-30
Change title & explanation of 'Dim messages after reading' to remove reference to specific colours/shades


sm4rk0:
### Expected behaviour

Some better name. For example: "Grey read messages' background"
### Actual behaviour

Unread messages have black background and read messages have grey background, so the unread messages are actually dimmer in case of dark theme.
### Steps to reproduce
1. Open Settings -> Global Settings -> Display
2. Set the theme to "Dark"
3. Check the "Dim messages after reading"
### Environment

K-9 Mail version:
5.108
Android version:
5.1
Account type (IMAP, POP3, WebDAV/Exchange):
N/A

philipwhiuk:
Is grey the right colour for them to be in dark mode? Maybe it's better to change the grey background to a darker blue in dark mode. (I don't use dark mode so I'm not sure)

Referencing specific colours in preferences would seem to be wrong from a ease-of-theming perspective.

Maybe just 'Different background for read messages' is the right phrasing - then it's a theme decision as to what to do.

sm4rk0:
I'm sorry - I've got the original bug report wrong (sleep deprivation).

Unread messages are grey and read messages are black. So, the "Dim messages" is basically correct. But, the option explanation is wrong in case of the dark theme: "A gray background will show that a message has been read".

Should I close this issue and open another, or rename this one?

I agree with this:
> Maybe just 'Different background for read messages' is the right phrasing - then it's a theme decision as to what to do.

PokerFace8:
Hi. I am working on this. What should the new summary of the setting say though? Right now it says "A grey background will show that a message has been read". 

Should the new version say nothing or something like "A change in background will show that a message has been read".

PokerFace8:
Working on this.

PokerFace8:
The problem with the title "Different background for read messages" is that the text does not completely show up and is completed with ellipsis. Ofcourse, we could subclass CheckBoxPreference and create the required preference to display 2 lines of title instead of one but it would be inconsistent with the other preferences.

I was thinking the title could be changed to the following : "Change shade after reading" with the summary : " A different background will show that a message has been read".

Midhun07:
I' am a beginner, can i work on this topic. How about "Color read items as grey".

philipwhiuk:
There's a PR already, hence why I added the label. Please pick a different issue.

rhari991:
Fixed by #2453

PR #2320, opened 2017-03-02
Fix issue #1232

PokerFace8:
Removes use of specific colors from the preference string.


philipwhiuk:
Please do PRs from branches in your repository so that each PR only contains the fixes a single issue.

PR #2330, opened 2017-03-03
Add support for the UIDVALIDITY field in IMAP accounts

rhari991:
Related issue : #1074 

I tested this by creating a new folder in my email account, deleting it and recreating it with the same name. All the emails in the folder were reloaded.

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -39,7 +39,7 @@ import static com.fsck.k9.mail.K9MailLib.LOG_TAG;
 import static com.fsck.k9.mail.store.imap.ImapUtility.getLastResponse;
 
 
-class ImapFolder extends Folder<ImapMessage> {
+public class ImapFolder extends Folder<ImapMessage> {
     private static final ThreadLocal<SimpleDateFormat> RFC3501_DATE = new ThreadLocal<SimpleDateFormat>() {
         @Override
         protected SimpleDateFormat initialValue() {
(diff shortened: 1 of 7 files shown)


k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
```diff
@@ -167,6 +167,32 @@ public void open(int mode) throws MessagingException {
         }
     }
 
+    public long getCurrentUidValidity() throws MessagingException {
```
Valodim: can't this be done in an easier way by simply calling `open` if the folder is not yet opened? this already open the folder, and fetches a `SelectOrExamineResponse`, which can be used to parse and store the uidvalidity, so in cases the folder is already opened we don't have to do any server communication at all

k9mail/src/main/java/com/fsck/k9/controller/MessagingController.java
```diff
@@ -1023,6 +1028,32 @@ Upload any local messages that are marked as PENDING_UPLOAD (Drafts, Sent, Trash
...
+    private void handleImapUidValidity(Account account, LocalFolder localFolder,
+                                       ImapFolder remoteImapFolder, String folder,
+                                       MessagingListener listener) throws MessagingException {
+        long previousUidValidity = localFolder.getLastUidValidity(account.getUuid(), remoteImapFolder.getName());
+        long currentUidValidity = remoteImapFolder.getCurrentUidValidity();
+
+        if (previousUidValidity != -1L && currentUidValidity != -1L && currentUidValidity != previousUidValidity) {
+
+            //Delete and reload all messages in the local folder
+            Set<String> localMessageUidSet = localFolder.getAllMessagesAndEffectiveDates().keySet();
```
Valodim: I wonder if it would be more appropriate to set the UIDs of all these messages to null, and leave deleting them to the normal syncing mechanism that relies on uids to do so. @cketti?

rhari991:
@Valodim made the change and added unit tests.

rhari991:
Adding this as part of #2607

PR #2332, opened 2017-03-03
Remove LineWarpOutputStream

danwain:
Wraping the line length to 1000 for SMTP is now not needed as
all emails are quoted-printable

quoted-printable has a maximum line length of 76 [rfc 2045](https://www.ietf.org/rfc/rfc2045.txt).
As this function is only called once to shorted it to 1000 characters length it is not needed.


danwain:
This also means that issue #2209 is redundant

danwain:
A point raised by @Valodim is that the headers can exceed this value. The current solution doesn't seem to work properly for long headers but that should probably be raised by another issue.

ployt0:
Where did is it written the 1000 char limit is no more?
I'm reading Tannenbaum, 2011, far later than that RFC, and he describes it as a requirement still.

PR #2348, opened 2017-03-05
Remove line wrap

danwain:
Wraping the line length to 1000 for SMTP is now not needed as
all emails are quoted-printable

quoted-printable has a maximum line length of 76 [rfc 2045](https://www.ietf.org/rfc/rfc2045.txt).
As this function is only called once to shorted it to 1000 characters length it is not needed.

I think this can now be reopened that foldAndEncode is implemented. This means that both the headers and the content is already less than a 1000 characters.


danwain:
#2336 this is possibly cause

danwain:
folding is described in [rfc 2822](https://tools.ietf.org/html/rfc2822#section-2.2.3)

philipwhiuk:
Would be nice to see a couple of tests for this.

danwain:
That would be a good idea, I'll have a look later today

ployt0:
It would be nice if we could actually _use_ line wrapping to enlarge text. It's a feature I most appreciated in Opera Mini but has rather fallen by the wayside of late. HTML is, IMHO, perfectly suited to the feature.

philipwhiuk:
@ployt0 this issue is about line-wrapping of MIME text received over IMAP. It's unrelated to how we display content.

danwain:
I've added a basic tester and changed the name as no encoding is done.
The encoding is done by MimeHeader

PR #2379, opened 2017-03-11
Adding Mail Scheduling

rhari991:
Related issue :  #688 

I added a menu item to the compose screen which leads to a preference screen to allow the user to select the date and time to send the mail. The mail remains in the Outbox like normal until it is sent.

schedule_mail

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/Message.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/Message.java
@@ -215,6 +215,10 @@ public abstract class Message implements Part, Body {
         return 0;
     }
 
+    public boolean isScheduled() {
+        return getSentDate().getTime() > System.currentTimeMillis();
+    }
+
     /**
      * Copy the contents of this object into another {@code Message} object.
      *
--- a/k9mail/src/main/AndroidManifest.xml
+++ b/k9mail/src/main/AndroidManifest.xml
@@ -248,6 +248,12 @@
             </intent-filter>
         </activity>
 
+        <activity
+            android:name=".preferences.ScheduleMailPrefs"
+            android:configChanges="locale"
+            android:excludeFromRecents="true"
+            android:label="@string/schedule_mail_prefs_title" />
+
         <!-- Search Activity - searchable -->
         <activity
             android:name=".activity.Search"
--- a/k9mail/src/main/java/com/fsck/k9/activity/MessageCompose.java
+++ b/k9mail/src/main/java/com/fsck/k9/activity/MessageCompose.java
@@ -10,6 +10,7 @@ import java.util.regex.Pattern;
 
 import android.annotation.SuppressLint;
 import android.app.ActionBar;
+import android.app.Activity;
 import android.app.AlertDialog;
 import android.app.AlertDialog.Builder;
 import android.app.Dialog;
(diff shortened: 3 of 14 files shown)


Valodim:
This looks like a good start, nice work! However this is a complex feature and I believe if we want to support it, some more thought is required. Things that come to mind:

- We will include android-jobs soon to deal with doze mode, which is probably a better approach to the scheduling aspect.
- Is the RTC_ALARM thing reliable? What happens if I turn off my phone during that time, will the mail be sent at the next possible time?
- Should we have a notification that a mail is being scheduled maybe, so the user can have confidence it's going to be sent, and also offering a "dismiss"?
- Is this feature present in other mailers? If yes, how do they do it?
- The schedule picker UI is on par with most of our other ui, but I think it can be improved. I don't have a full concept in mind and don't know which widgets are available, but the current way seems clumsy.

rhari991:
- I updated the commit so that the messages are rescheduled in the BootReceiver. Now they will be sent after a device reboot.
- I'm not sure if a notification is necessary. We could have an option in the settings or a checkbox in the schedule page that shows it.
- Both MailDroid and Boomerang have this feature. They do it in a similar way - you select the date/time in the compose screen and press send. The message is stored in a particular folder until it is sent. They also allow the user to edit/reschedule it before it is sent - this is a feature I am working on supporting. Maildroid also shows a little area at the bottom with the scheduled time and another button to change the scheduled time 
maildroid_compose

- Regarding the UI, both the other apps also offer quick shortcuts to select a schedule time, like 10 minutes later, one hour later etc. I'm not sure how useful this would be.

- Boomerang has a separate activity to select the date and time.  Selecting the date and time directly schedules the message instead of taking the user back to the compose screen.
boomerang_schedule

- Maildroid opens a radio button list to select the schedule date and time. If you select the custom date/time option, a datepicker/timepicker appears on top of this dialog.
maildroid_schedule


Also, currently the scheduled messages sit in the Outbox till they are sent. An alternative could be to store them in the drafts folder.

@cketti @philipwhiuk Please share your thoughts on this.

philipwhiuk:
I think a separate 'Scheduled Mail' folder makes more sense than either Outbox or Drafts. But I don't know if it's worth a separate folder. Outbox makes more sense than Drafts to me though.

Notification might have been the wrong word - maybe a Toast when you tap send saying 'Message scheduled for x' ? I'm not convinced you need a notification (there should be one anyway when the message is being sent, right?).

The UI for Boomerang's scheduling looks nice. I think having some good defaults is worth adding and makes it much less hassle. I'm assuming people who schedule mail schedule quite a bit - I don't do it.

The general 'alway delay send' feature (e.g Gmail's Undo Send where you specify a cancellation period) could use this functionality - so that's worth thinking about for an implementation stand point.

cbabs:
Is this going to be added?  It would be a welcome feature for me

m4ci3k2:
I'm using a similar add-on in Thunderbird ([Send Later](https://addons.mozilla.org/pl/thunderbird/addon/send-later-3)) and the feature I use the most is sending the message within a specific time window (e.g. during work hours). Yet it's UI is not a thing to mention.

The Boomerang's UI looks better to me, it's easier to select icon from a grid than an element from a long list. Also more information is displayed, for example the time for tomorrow.

CueHD:
Is this accomplished using the SMTP FUTURESEND extension?
See [https://fastmail.blog/2019/08/21/fastmail-now-supports-undo-send/](url).

cketti:
Thanks for working on implementing this feature. However, I believe we should not attempt to schedule sending the message locally. There are too many things that could go wrong.

Instead we should use FUTURESEND ([RFC 4865](https://tools.ietf.org/html/rfc4865)).

moi90:
I think this should have been discussed more thoroughly. Both local and on-server delayed sending have their merits and downsides.

But as the local solution was turned down, maybe someone could implement server-based solution reusing the UI parts of this PR?

Issue #2164, opened 2017-02-01
Drafts deleted when syncronizing with IMAP server


JustAnonymous:
Please search to check for an existing issue (including closed issues, for which the fix may not have yet been released) before opening a new issue: https://github.com/k9mail/k-9/issues?q=is%3Aissue

### Expected behavior
When "Save as Draft" is selected, draft should be remain saved either locally or on the server for future access. 

### Actual behavior
Either draft is automatically completely deleted and no longer available upon server sync, or the draft remains in the "drafts" folder but when opening the compose window, the text is no longer available. It seems this is dependent on the encryption setting; everytime I re-open the draft I have to re-select "Don't encrypt" because the default is always "encrypt if possible".  The last selection should be remembered. If "encypt if possible" or "encrypt" is selected, the draft is deleted upon sync.  If not, I can often no longer access the draft text. This happens regardless of account, and only started after upgrading to the latest K9 version and adding my keys to OpenKeychain (I was using a much older version of K9 prior to this, not sure which).

### Steps to reproduce
1. Create a draft email with text in the body.
2. Select "don't encrypt"
3. Save draft by selecting either "save draft" or by pressing back and "save draft".
4. Text is gone after server sync.

OR

1. Create a draft email with text in the body.
2. Select "encrypt" or "encrypt if possible".
3. Save draft by selecting either "save draft" or by pressing back and "save draft".
4. Draft is gone after server sync. 

### Environment
K-9 Mail version: 5.203

Android version: 5.1.1

Account type (IMAP, POP3, WebDAV/Exchange): IMAP

lbort:
I second this. It is really annoying, since writing an email on my phone, and continue writing on my PC does not work anymore. 

**Another way to reproduce:** 

1. Start writing Email, without recipient or encryption setting
2. Leave app by switching to another, triggering a "save as draft"
3. Draft is saved in local draft-folder, but disappears if draftfolder is synced.

 as far as I remember, it used to work recently, i.e. with version 5.202.

I also have errors in the K9mail-errors folder, which seam to be related. See Attached: 
```
K9-Mail version: 5.205
Device make: FP1
Device model: FP1U
Android version: 4.2.2

javax.net.ssl.SSLException: Read error: ssl=0x5deda610: I/O error during system call, Connection timed out
 at org.apache.harmony.xnet.provider.jsse.NativeCrypto.SSL_read(Native Method)
[...]

PR #2407, opened 2017-03-18
Fix for drafts getting deleted during IMAP sync

Mvishnu9:
This is a possible fix for issue #2164. Specifically, the following scenario mentioned in that issue:
1. Create a draft email with text in the body.
2. Select "encrypt" or "encrypt if possible".
3. Save draft by selecting either "save draft" or by pressing back and "save draft".
4. Draft is gone after server sync.

As per the comment by @Valodim on Jan 7 in issue #1983, the intent, at least until draft encryption is implemented, is to save drafts of secure messages ("encrypt" / "encrypt if possible") locally only and not upload them to the server during IMAP sync. But, as mentioned above, such drafts are getting saved locally initially but then getting deleted during the sync. (As per my testing, the issue of drafts getting deleted during sync seems to have been there for a while. In releases 5.108 to 5.200 (which I tested), drafts of secure messages were getting copied to the server in plain text (issue #711) and normal drafts were getting deleted upon sync (opposite of the #2164 scenario mentioned above)).

Hope my fix addresses the issue adequately. Please let me know if any further changes are required. 
(Note: I am planning to apply to k-9 as part of GSOC)

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/Flag.java
@@ -62,4 +62,10 @@ public enum Flag {
      * This flag is used for drafts where the message should be sent as PGP/INLINE.
      */
     X_DRAFT_OPENPGP_INLINE,
+
+    /**
+     * This flag is used for secure message drafts which are to be stored locally only and not remotely (until draft encryption is implemented).
+     */
+    X_SECURE_MESSAGE_DRAFT,
+
 }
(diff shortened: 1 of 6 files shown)


Mvishnu9:
I will resolve the build issues and update.

philipwhiuk:
Pull requests are automatically updated when you push to the branch. I'll close #2410

Mvishnu9:
Hi,
Actually #2410 is a separate branch (from #2407). I rewrote things using new flag names etc. So it was not possible to simply continue #2407. Please review both and continue the process with whichever implementation is better and close the other. I think the implementation in #2410 is better but It is possible that I misunderstood the behaviour of cryptomode setting and so my implementation of that part (in #2410) may be wrong/unnecessary, in which case #2407 might be better and #2410 can be closed.
Thanks,
Vishnu

----- Original Message -----

cketti:
Fixed via #3174
Fixed via #3174

PR #2328, opened 2017-03-03
Use SystemClock.elapsedRealtime() to measure durations

CodingRmy:
I tried to solve this , guide me if i  done any mistakes.

PR #2415, opened 2017-03-21
Changes done with elapsedRealtime()

Jagruti1010:


Jagruti1010:
Since there are many Changes remained still in PR #2328 I Updated complete. please let me know if something i have done wrong. @cketti @Valodim

PR #2439, opened 2017-03-25
fix: Save As Draft problem Issue #2433

majeedk526:
This is the fix for the issue #2433 

This is my first pull request. So if anything is wrong, pls let me know!

* Added and interface AttachmentListener at k9mail/src/main/java/com/fsck/k9/activity/msc/
* MessageCompose.java implements this Listener
* AttachmentPresenter.java calls this listenre after adding the attachment

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/activity/MessageCompose.java
+++ b/k9mail/src/main/java/com/fsck/k9/activity/MessageCompose.java
@@ -66,6 +66,7 @@ import com.fsck.k9.activity.compose.RecipientPresenter;
 import com.fsck.k9.activity.compose.RecipientPresenter.CryptoMode;
 import com.fsck.k9.activity.compose.SaveMessageTask;
 import com.fsck.k9.activity.misc.Attachment;
+import com.fsck.k9.activity.misc.AttachmentListener;
 import com.fsck.k9.controller.MessagingController;
 import com.fsck.k9.controller.MessagingListener;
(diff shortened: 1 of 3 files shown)


philipwhiuk:
1. We carefully considered what changes make a maintenance release versus what changes make a major release. We determine this by pushing them to master and evaluating them. The main reason for this is that by having them on master people building off master (developers) are able to test it as part of other changes. Typically nobody builds and installed unreleased MAINT branch code.
2. You've taken my code, removed much of the functionality and then put yourself as the author. Quite aside from the fact that we don't put authorship comments (or indeed any JavaDoc) in code (per our CodeStyle), on a personal level it's pretty annoying to have people claim they wrote stuff I wrote.

When I left the issue open I assumed it was because my code had been merged into 5.2 and there was still a bug. As this is not the case the issue will be closed and the milestones updated on the issues. Either we will cherry-pick my change for 5.207 or it will be part of the next major release.

majeedk526:
Well, I think I should clarify.

First of all thank you for pointing my mistakes (Not putting author name, CodeStyle and Not builidng MAINT branch).

Secondly using an Interface pattern is quite common. What makes you think that no other developer other than you could think of that ? And claiming Copying of your Code? If you have read the conversation on the issue carefully you must have noticed my comment stating the issue is already fixed in the Main branch. My commit was only to fix the issue in the MAINT branch  by writing my own code (which I think now I shouldn't have done) and not by copying and reducing the functionality of your code.

PR #2440, opened 2017-03-25
Fix K9RobolectricTestRunner

philipwhiuk:
As seen in testing, the path for the manifest is wrong.


philipwhiuk:
It looks like the fact that it fails in the IDE is an IDE path issue. Closing this...

PR #2453, opened 2017-03-28
Change title and explanation for read message colour change

philipwhiuk:
Fixes issue #1232

Replaces #2323


Valodim:
Hm. This probably isn't worth bikeshedding about, but the new strings just seem to be more vague than the previous ones. Is this really more clear for anyone? How will the color change?

PokerFace8:
Maybe the phrasing could be "Darken messages after reading"? In light mode, grey(read) is darker than white(unread). In dark mode, black(read) is darker than grey(unread).

philipwhiuk:
I thought the whole point is they aren't darker...... Otherwise dim would have been fine...

Issue #705, opened 2015-07-05
Use voice dictation to compose emails


knoy:
It would be very helpful in certain cases if you could compose emails using your voice (i.e: while driving, typing on a keyboard or using a pencil)

nuntius35:
Maybe [Kõnele](https://github.com/Kaljurand/K6nele) can be useful?

philipwhiuk:
This can be done by installing an appropriate keyboard (e.g. enabling Google voice). There's no need for K-9 changes.

PR #2468, opened 2017-03-30
#705 Resolved voice dictation for mail composition

RanchoDbest:
Implemented voice recognition for composing mail.
- Added mic button in message_compose.xml
- Updated MessageCompose.java
- Gave permission in Androidmanifest.xml
- Added xml file named dialog_matches_flag.xml


philipwhiuk:
This feature is better provided by just using a keyboard that supports speech-to-text. If you do that you get it in every field.

PR #2495, opened 2017-04-12
Crypto status dryrun

Valodim:
This PR changes how the crypto status icon is displayed and most importantly, calls out to OpenKeychain to retrieve information about the state of selected recipients.

This comes with a cost in complexity since that call is made asynchronously, and there is a new "weird state" where the status is not yet retrieved and the user presses "send". However this should be so rare that it should never come up in practice, which is why I silently abort the user interaction rather than add even more complexity to handle it.

From OpenKeychain's side, the mechanism works as a "dry run" for sign+encrypt, following the same codepath as a regular sign+encrypt operation but aborting the process once all metadata is gathered, returning with a status. This requires a development version of OpenKeychain that includes this PR: https://github.com/open-keychain/open-keychain/pull/2088. I will push for a beta release with that included to make this easier to test.

Sadly, it's not possible to be compatible with older openkeychain versions here unless I keep the old crypto status mechanism around, which I really don't want to do. The plan is to release this in OpenKeychain earlier than K-9 and just error out for older versions.


dschuermann:
OpenKeychain 4.4-beta1 now includes the API.

Valodim:
For release management reasons, this feature is currently not in the openkeychain beta anymore :)

Valodim:
this PR is superseded by #2608

PR #2505, opened 2017-04-18
Add autoconfig settings for riseup.net

micah:
Add autoconfig settings for riseup.net


pshem:
If you want to update it, why not pull in everything from AOSP master? The file is at https://android.googlesource.com/platform/packages/apps/Email/+/master/res/xml/providers.xml

PR #2511, opened 2017-04-25
Refactor recipient drag and drop code

rhari991:
Contains changes as discussed in PR #2451


rancidfrog:
Is there a way to copy or edit email address?
Currently I am unable to edit or copy email when compose is opened from other app

It would be great to have the ability to edit or copying address, somehow.
When typing address manually clicking backspace allows for editing, but when compose opened directly with email mailto link, unable to do so

cketti:
See https://github.com/k9mail/k-9/pull/2451#issuecomment-572222353

PR #2580, opened 2017-06-11
GSoC Account Setup Branch Review

philipwhiuk:
Just opening this so I can provide early feedback.


k9mail-library/src/main/java/com/fsck/k9/mail/autoconfiguration/AutoConfigurationUtil.java
```diff
@@ -0,0 +1,206 @@
...
+import org.xbill.DNS.Record;
+import org.xbill.DNS.Resolver;
+import org.xbill.DNS.SRVRecord;
+import org.xbill.DNS.SimpleResolver;
+import org.xbill.DNS.TextParseException;
+import org.xbill.DNS.Type;
+
+
+/**
+ * Created by daquexian on 6/9/17.
```
philipwhiuk: We don't do author attribution in the code.

k9mail-library/src/main/java/com/fsck/k9/mail/autoconfiguration/AutoConfigurationUtil.java
```diff
@@ -0,0 +1,206 @@
...
+                final String[] targetParts = target.split("\\.");
+
+                String newDomain = targetParts[targetParts.length - 2] + "." + targetParts[targetParts.length - 1];
+
+                if (!newDomain.equals(domain)) {
+                    providerInfo = findProviderInISPDB(newDomain);
+                }
+            }
+        } catch (Exception e) {
+
```
philipwhiuk: Add logging for this, especially because you're catching `Exception` rather than something specific.

k9mail-library/src/main/java/com/fsck/k9/mail/autoconfiguration/AutoConfigurationUtil.java
```diff
@@ -0,0 +1,206 @@
...
+                case "%EMAILLOCALPART%":
+                    providerInfo.outgoingUsernameTemplate = ProviderInfo.USERNAME_TEMPLATE_USER;
+                    break;
+                default:
+                    break;
+            }
+            return providerInfo;
+
+        } catch (Exception e) {
+            return null;
```
philipwhiuk: Again loggin.

k9mail-library/src/main/java/com/fsck/k9/mail/autoconfiguration/AutoConfigurationUtil.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        return providerInfo;
+    }
+
+    private static ProviderInfo findProviderInISPDB(String domain) {
+        try {
+            ProviderInfo providerInfo = new ProviderInfo();
+
+            Document document = Jsoup.connect("https://autoconfig.thunderbird.net/v1.1/" + domain).get();
```
philipwhiuk: Make this string a constant. I think it probably belongs as a non translatable string resource.

k9mail-library/src/main/java/com/fsck/k9/mail/autoconfiguration/AutoConfigurationUtil.java
```diff
@@ -0,0 +1,206 @@
...
+                        providerInfo.outgoingSocketType = "";
+                        break;
+                }
+                providerInfo.outgoingType = "stmp";
+            } else {
+                return null;
+            }
+
+        } catch (TextParseException | UnknownHostException e) {
+            return null;
```
philipwhiuk: Log it

k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSetupBasics.java
```diff
@@ -501,6 +503,50 @@ private Provider findProviderForDomain(String domain) {
...
+
+                providerInfo = AutoConfigurationUtil.findProviderMozilla(domain);
+                if (providerInfo == null) {
+                    providerInfo = AutoConfigurationUtil.findProviderBySrv(domain);
+                }
+
+                try {
+                    mProvider = Provider.newInstanceFromProviderInfo(providerInfo);
+                } catch (URISyntaxException e) {
+                }
```
philipwhiuk: Is it worth logging this? Should be rare and so interesting, right?

daquexian:
@philipwhiuk Thanks for your review!!

A trouble I met is that according to rfc6186, there is no way to detect what is used as a user identifier when contacting to the mail server. We must use the full email address first, if it fails, then fall back to using the "local-part" extracted from the email address. To achieve this, code in AccountSetupCheckSettings need to be modified. Considering that the whole workflow will be rewritten in the future, I just use email address as user identifier, and plan to implement this fully when the new workflow is ready.

Valodim:
This whole thing is a perfect candidate for test-driven development, since most of the logic can be kept in pure functions: create mock responses for the dns/http queries, write tests for the parsing methods that check the expected output, *then* implement the methods.

The techniques should also be kept in separate classes, implementing a common `findProviderInfo` interface. They can then be tied together by a class that depends on those specific techniques and just runs them in some order.

daquexian:
@Valodim Thanks for your review! What's the advantage of creating separate classes rather than different methods in just one class?

daquexian:
I implemented [autodiscover](https://technet.microsoft.com/en-us/library/cc511507.aspx?f=255&MSPPError=-2147217396). But due to the lack of documents, there may be something incorrect in my implementation. For example I don't know exactly what will happen if `DomainRequired` is "on".

I also write tests for three mechanisms. The idea of test-driven development is really helpful. I found some bugs by tests and fixed them.

daquexian:
rewrite some UI. Still need do more and integrate auto configuration with settings checking

k9mail/src/main/java/com/fsck/k9/activity/K9MaterialActivity.java
```diff
@@ -0,0 +1,37 @@
...
+import com.fsck.k9.activity.misc.SwipeGestureDetector.OnSwipeGestureListener;
+
+
+/**
+ * Created by daquexian on 6/26/17.
+ */
+
+public class K9MaterialActivity extends AppCompatActivity implements K9ActivityMagic {
+
+    private K9ActivityCommon mBase;
```
philipwhiuk: We don't use `m` prefixes in new code.

k9mail/src/main/java/com/fsck/k9/activity/setup/autoconfiguration/AccountSetupAutoConfigurationContract.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.fsck.k9.Account;
+import com.fsck.k9.BasePresenter;
+import com.fsck.k9.BaseView;
+import com.fsck.k9.mail.CertificateValidationException;
+import com.fsck.k9.mail.autoconfiguration.AutoConfigure.ProviderInfo;
+
+
+/**
+ * Created by daquexian on 6/26/17.
```
philipwhiuk: Remove this

k9mail/src/main/java/com/fsck/k9/activity/setup/autoconfiguration/AccountSetupAutoConfigurationPresenter.java
```diff
@@ -0,0 +1,576 @@
...
+                    // we need these for matching
+                    String storeURIHost = (Uri.parse(account.getStoreUri())).getHost();
+                    String transportURIHost = (Uri.parse(account.getTransportUri())).getHost();
+
+                    for (List<?> subjectAlternativeName : subjectAlternativeNames) {
+                        Integer type = (Integer)subjectAlternativeName.get(0);
+                        Object value = subjectAlternativeName.get(1);
+                        String name;
+                        switch (type.intValue()) {
+                            case 0:
```
philipwhiuk: Make these constants.

k9mail/src/main/java/com/fsck/k9/activity/setup/autoconfiguration/AccountSetupAutoConfigurationPresenter.java
```diff
@@ -0,0 +1,576 @@
...
+                                        transportURIHost.endsWith(name.substring(2)))) {
+                            //TODO: localize this string
+                            altNamesText.append("Subject(alt): ").append(name).append(",...\n");
+                        }
+                    }
+                    chainInfo.append(altNamesText);
+                }
+            } catch (Exception e1) {
+                // don't fail just because of subjectAltNames
+                Timber.w(e1, "cannot display SubjectAltNames in dialog");
```
philipwhiuk: What exceptions are we expecting / catching here - this is pretty broad.

k9mail/src/main/java/com/fsck/k9/activity/setup/autoconfiguration/AccountSetupAutoConfigurationPresenter.java
```diff
@@ -0,0 +1,576 @@
...
+            view.showAcceptKeyDialog(msgResId, exMessage, chainInfo.toString(), chain[0]);
+        }
+
+    }
+
+    private void handleCertificateValidationException(CertificateValidationException cve) {
+        Timber.e(cve, "Error while testing settings");
+
+        X509Certificate[] chain = cve.getCertChain();
+        // Avoid NullPointerException in acceptKeyDialog()
```
philipwhiuk: This comment is silly - obviously checking for null avoids an NPE. Doesn't really tell us why it would be null.

k9mail/src/main/java/com/fsck/k9/activity/setup/autoconfiguration/AccountSetupAutoConfigurationPresenter.java
```diff
@@ -0,0 +1,576 @@
...
+        private final Account account;
+
+        private CheckAccountTask(Account account) {
+            this.account = account;
+        }
+
+        @Override
+        protected Void doInBackground(CheckDirection... params) {
+            try {
+                // TODO: 6/27/17 it's hard to handle exception and integrate auto configuration with account checking. I plan to refactor this part using RxJava
```
philipwhiuk: If you get an exception you should probably just bail into manual mode.

k9mail/src/main/java/com/fsck/k9/activity/setup/basics/AccountSetupBasics.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    public void onTextChanged(CharSequence s, int start, int before, int count) {
+    }
+
+    private void onNext() {
+
+        String email = mEmailView.getText().toString();
+        String password = mPasswordView.getText().toString();
+        // String[] emailParts = splitEmail(email);
```
philipwhiuk: Commented out code.

k9mail/src/main/java/com/fsck/k9/activity/setup/basics/AccountSetupBasicsContract.java
```diff
@@ -0,0 +1,20 @@
+package com.fsck.k9.activity.setup.basics;
+
+
+import com.fsck.k9.BasePresenter;
+import com.fsck.k9.BaseView;
+
+
+/**
+ * Created by daquexian on 6/26/17.
```
philipwhiuk: Remove

k9mail/src/main/java/com/fsck/k9/activity/setup/autoconfiguration/AccountSetupAutoConfigurationPresenter.java
daquexian: Thanks for your review! These codes are copied from `AccountSetupCheckSettings` and not written by me. But I will fix these issues.

daquexian:
Just integrated autoconfiguration and the new UI. The Travis CI failed because test about new UI should be updated, will fix it in this branch soon.

daquexian:
@philipwhiuk I'm working on xoauth2 based on your code :) Thanks for your amazing work.

I'm using my own debug.keystore to register on Google API Console, so that everyone except me can't use this feature now. It's inconvenient to review. Do you have any advice on it? :)

philipwhiuk:
Unfortunately not. The actual Google API key thing was something I never got fully working. I think I authorised my device to use it regardless in the end (which I never really worked out how you revoked...) Once we push it it'll have to use a keystore provided by cketti I expect.

daquexian:
@philipwhiuk I have made oauth 2.0 without dependency on Android account system working. It seems keystore does not need to be matched any more.

k9mail-library/src/main/java/com/fsck/k9/mail/autoconfiguration/DNSHelper.java
```diff
@@ -0,0 +1,22 @@
...
+import org.xbill.DNS.TextParseException;
+
+
+public class DNSHelper {
+    public static String getMXDomain(String domain) throws TextParseException, UnknownHostException {
+        DNSOperation dnsOperation = new DNSOperation();
+        MXRecord mxRecord = dnsOperation.mxLookup(domain);
+        if (mxRecord != null) {
+            final String target = mxRecord.getTarget().toString(true);
+            final String[] targetParts = target.split("\\.");
```
Valodim: it's not immediately obvious what this does, please extract it into a method (e.g. `getDomainFromFqdn`)

k9mail/src/main/java/com/fsck/k9/account/OAuth2WebViewClient.java
```diff
@@ -30,9 +31,11 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
         Uri uri = Uri.parse(url);
 
         if (arrivedAtRedirectUri(uri)) {
-            if (uri.getQueryParameter("error") != null) {
-                Timber.i("got oauth error: " + uri.getQueryParameter("error"));
-                requestHandler.onErrorWhenGettingOAuthCode(uri.getQueryParameter("error"));
+            final String error = uri.getQueryParameter("error");
+            if (error != null) {
+                Timber.i("got oauth error: " + error);
```
Valodim: `Timber.e` :)

k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSetupActivity.java
```diff
@@ -1336,33 +1293,30 @@ public void closeAuthDialog() {
...
-        canceled = false;
         presenter.onActivityResult(requestCode, resultCode, data);
     }
 
     @Override
     protected void onResume() {
         super.onResume();
-        canceled = false;
-        destroyed = false;
+        presenter.onResume();
```
Valodim: good changes, but also unrelated to "add guessing" :(

k9mail/src/main/res/values/strings.xml
```diff
@@ -1254,4 +1254,5 @@ Please submit bug reports, contribute new features and ask questions at
     <string name="xoauth2_no_account">No account provided</string>
     <string name="linked_webview_title_gmail">Gmail Authorization</string>
     <string name="linked_webview_title_outlook">Outlook Authorization</string>
+    <string name="account_setup_check_settings_guessing_msg">Guessing server settings...</string>
```
Valodim: We probably shouldn't tell the user that we are "Guessing" anything ;)

k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSetupPresenter.java
```diff
@@ -216,17 +226,44 @@ public void onNegativeClickedInConfirmationDialog() {
...
+                if (providerInfo != null || (incomingReady && outgoingReady)) {
+                    return providerInfo;
+                }
+
+                incomingReady = false;
+                outgoingReady = false;
+                try {
+                    String mxDomain = DNSHelper.getMXDomain(domain);
+
+                    if (mxDomain == null) return providerInfo;
```
Valodim: why not use the domain itself if there is no MX record? also, isn't providerInfo likely null here?

k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSetupPresenter.java
```diff
@@ -243,14 +280,89 @@ protected ProviderInfo doInBackground(Void... params) {
...
                 providerInfo = autodiscover.findProviderInfo(email);
+                if (providerInfo != null) return providerInfo;
 
-                return providerInfo;
+                testDomain(domain);
+
+                return null;
+            }
+
+            private void testDomain(String domain) {
```
Valodim: I feel like there is a lot of logic here that warrants its own class. There are already `Autoconfigure*` classes, why not have an `AutoconfigureHeuristic` one?

k9mail-library/src/main/java/com/fsck/k9/mail/autoconfiguration/DNSHelper.java
```diff
@@ -0,0 +1,22 @@
...
+
+
+import java.net.UnknownHostException;
+
+import org.xbill.DNS.MXRecord;
+import org.xbill.DNS.TextParseException;
+
+
+public class DNSHelper {
+    public static String getMXDomain(String domain) throws TextParseException, UnknownHostException {
```
Valodim: Should be `getMxDomain`. Also, it would be nicer if we didn't expose the TextParseException from the library to the caller here, a broken record might as well return `null` here.

Valodim:
Please do a rebase on current master when convenient :)

Valodim:
the autoconfig currently only tries the mail address as login name. I think it's common to have only the local part of the address (before the @) used as a login name, so that should be tried as well.

Also, when a server is found and authentication fails, that should be correctly handled, telling the user that their password was wrong.

k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSetupPresenter.java
```diff
@@ -243,14 +280,89 @@ protected ProviderInfo doInBackground(Void... params) {
...
+            protected void onProgressUpdate(Integer... values) {
+                super.onProgressUpdate(values);
+                view.setMessage(values[0]);
             }
 
             @Override
             protected void onPostExecute(ProviderInfo providerInfo) {
                 super.onPostExecute(providerInfo);
 
+                if (canceled) {
```
daquexian: Actually I have a question here. In master branch a variable `canceled` is used to avoid unexpected things after activity is paused, stopped and so on, so I followed it here. But why not use `AsyncTask.cancel` to cancel the task directly here?

Valodim: If that's the only thing it does, then yes it can be used directly

k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSetupPresenter.java
daquexian: I found it's not so convenient to make these a class because it should modify the `accountConfig` directly(rather than return a `ProviderInfo` object like other `Autoconfigure*` classes do) and modify some global variables(`incomingReady` and `outgoingReady`), what's more some methods are in module `k9mail` but not `k9mail-library`(though it can be simply moved to `k9mail-library` if we want). I think it may be better to do other things first and then try to spend time on it.

Valodim:
Please fix tests, and rebase this on current master :+1:

philipwhiuk:
I'm going to close this review (it's rather long and will be even more confusing post-rebase) - create a new one for the rebased version.

AaronM04:
@philipwhiuk has anybody created a new version of this pull request? I'd really like to avoid the Google Security Alert errors when K-9 Mail happens to check for email when I'm not on WiFi. Thanks :)

PR #2607, opened 2017-06-27
Add support for CONDSTORE and QRESYNC

rhari991:
This PR does several things 

- Moves the folder synchronization code from `MessagingController` into separate classes depending on the protocol (`ImapSyncInteractor` for IMAP accounts and `LegacySyncInteractor` for POP3 and WebDAV accounts)
- Modifies the `MessagingControllerPushReceiver` so that a normal synchronization is performed when untagged responses are received through IDLE, instead of response-specific actions. This synchronization is done through a separate connection instead of reusing the connection through which IDLE responses are received.
- Adds support for the CONDSTORE and QRESYNC IMAP extensions  as described in [RFC 7162](https://tools.ietf.org/html/rfc7162).

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/K9MailLib.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/K9MailLib.java
@@ -3,7 +3,6 @@ package com.fsck.k9.mail;
 
 public class K9MailLib {
     private static DebugStatus debugStatus = new DefaultDebugStatus();
-    private static ImapExtensionStatus imapExtensionStatus = new DefaultImapExtensionStatus();
 
     private K9MailLib() {
     }
@@ -43,14 +42,6 @@ public class K9MailLib {
         return debugStatus.debugSensitive();
     }
 
-    public static boolean shouldUseCondstore() {
-        return imapExtensionStatus.useCondstore();
-    }
-
-    public static boolean shouldUseQresync() {
-        return imapExtensionStatus.useQresync();
-    }
-
     public static void setDebugSensitive(boolean b) {
         if (debugStatus instanceof WritableDebugStatus) {
             ((WritableDebugStatus) debugStatus).setSensitive(b);
@@ -63,28 +54,10 @@ public class K9MailLib {
         }
     }
 
-    public static void setUseCondstore(boolean useCondstore) {
-        if (imapExtensionStatus instanceof WritableImapExtensionStatus) {
-            ((WritableImapExtensionStatus) imapExtensionStatus).setUseCondstore(useCondstore);
-        }
-    }
-
-    public static void setUseQresync(boolean useQresync) {
-        if (imapExtensionStatus instanceof WritableImapExtensionStatus) {
-            ((WritableImapExtensionStatus) imapExtensionStatus).setUseQresync(useQresync);
-        }
-    }
-
     public interface DebugStatus {
         boolean enabled();
 
         boolean debugSensitive();
-    }
-
-    public interface ImapExtensionStatus {
-        boolean useCondstore();
-
-        boolean useQresync();
     }
 
     public static void setDebugStatus(DebugStatus status) {
@@ -94,23 +67,10 @@ public class K9MailLib {
         debugStatus = status;
     }
 
-    public static void setImapExtensionStatus(ImapExtensionStatus status) {
-        if (status == null) {
-            throw new IllegalArgumentException("status cannot be null");
-        }
-        imapExtensionStatus = status;
-    }
-
     private interface WritableDebugStatus extends DebugStatus {
         void setEnabled(boolean enabled);
 
         void setSensitive(boolean sensitive);
-    }
-
-    private interface WritableImapExtensionStatus extends ImapExtensionStatus {
-        void setUseCondstore(boolean useCondstore);
-
-        void setUseQresync(boolean useQresync);
     }
 
     private static class DefaultDebugStatus implements WritableDebugStatus {
@@ -135,31 +95,6 @@ public class K9MailLib {
         @Override
         public void setSensitive(boolean sensitive) {
             this.sensitive = sensitive;
-        }
-    }
-
-    private static class DefaultImapExtensionStatus implements WritableImapExtensionStatus {
-        private boolean useCondstore = true;
-        private boolean useQresync = true;
-
-        @Override
-        public boolean useCondstore() {
-            return useCondstore;
-        }
-
-        @Override
-        public boolean useQresync() {
-            return useQresync;
-        }
-
-        @Override
-        public void setUseCondstore(boolean useCondstore) {
-            this.useCondstore = useCondstore;
-        }
-
-        @Override
-        public void setUseQresync(boolean useQresync) {
-            this.useQresync = useQresync;
         }
     }
 }
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/K9MailLib.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/K9MailLib.java
@@ -3,6 +3,7 @@ package com.fsck.k9.mail;
 
 public class K9MailLib {
     private static DebugStatus debugStatus = new DefaultDebugStatus();
+    private static ImapExtensionStatus imapExtensionStatus = new DefaultImapExtensionStatus();
 
     private K9MailLib() {
     }
@@ -42,6 +43,14 @@ public class K9MailLib {
         return debugStatus.debugSensitive();
     }
 
+    public static boolean shouldUseCondstore() {
+        return imapExtensionStatus.useCondstore();
+    }
+
+    public static boolean shouldUseQresync() {
+        return imapExtensionStatus.useQresync();
+    }
+
     public static void setDebugSensitive(boolean b) {
         if (debugStatus instanceof WritableDebugStatus) {
             ((WritableDebugStatus) debugStatus).setSensitive(b);
@@ -54,10 +63,28 @@ public class K9MailLib {
         }
     }
 
+    public static void setUseCondstore(boolean useCondstore) {
+        if (imapExtensionStatus instanceof WritableImapExtensionStatus) {
+            ((WritableImapExtensionStatus) imapExtensionStatus).setUseCondstore(useCondstore);
+        }
+    }
+
+    public static void setUseQresync(boolean useQresync) {
+        if (imapExtensionStatus instanceof WritableImapExtensionStatus) {
+            ((WritableImapExtensionStatus) imapExtensionStatus).setUseQresync(useQresync);
+        }
+    }
+
     public interface DebugStatus {
         boolean enabled();
 
         boolean debugSensitive();
+    }
+
+    public interface ImapExtensionStatus {
+        boolean useCondstore();
+
+        boolean useQresync();
     }
 
     public static void setDebugStatus(DebugStatus status) {
@@ -67,10 +94,23 @@ public class K9MailLib {
         debugStatus = status;
     }
 
+    public static void setImapExtensionStatus(ImapExtensionStatus status) {
+        if (status == null) {
+            throw new IllegalArgumentException("status cannot be null");
+        }
+        imapExtensionStatus = status;
+    }
+
     private interface WritableDebugStatus extends DebugStatus {
         void setEnabled(boolean enabled);
 
         void setSensitive(boolean sensitive);
+    }
+
+    private interface WritableImapExtensionStatus extends ImapExtensionStatus {
+        void setUseCondstore(boolean useCondstore);
+
+        void setUseQresync(boolean useQresync);
     }
 
     private static class DefaultDebugStatus implements WritableDebugStatus {
@@ -95,6 +135,31 @@ public class K9MailLib {
         @Override
         public void setSensitive(boolean sensitive) {
             this.sensitive = sensitive;
+        }
+    }
+
+    private static class DefaultImapExtensionStatus implements WritableImapExtensionStatus {
+        private boolean useCondstore = true;
+        private boolean useQresync = true;
+
+        @Override
+        public boolean useCondstore() {
+            return useCondstore;
+        }
+
+        @Override
+        public boolean useQresync() {
+            return useQresync;
+        }
+
+        @Override
+        public void setUseCondstore(boolean useCondstore) {
+            this.useCondstore = useCondstore;
+        }
+
+        @Override
+        public void setUseQresync(boolean useQresync) {
+            this.useQresync = useQresync;
         }
     }
 }
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/Capabilities.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/Capabilities.java
@@ -3,7 +3,6 @@ package com.fsck.k9.mail.store.imap;
 
 class Capabilities {
     public static final String IDLE = "IDLE";
-    public static final String ENABLE = "ENABLE";
     public static final String CONDSTORE = "CONDSTORE";
     public static final String QRESYNC = "QRESYNC";
     public static final String SASL_IR = "SASL-IR";
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
@@ -604,7 +604,7 @@ public class ImapConnection {
     }
 
     void enableQresync() throws IOException, MessagingException {
-        if (!qresyncEnabled && hasCapability(Capabilities.QRESYNC) && K9MailLib.shouldUseQresync()) {
+        if (!qresyncEnabled && hasCapability(Capabilities.QRESYNC) && ImapConfig.shouldUseQresync()) {
             executeSimpleCommand(String.format("%s %s %s", Commands.ENABLE, Capabilities.CONDSTORE,
                     Capabilities.QRESYNC));
             qresyncEnabled = true;
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
@@ -604,9 +604,8 @@ public class ImapConnection {
     }
 
     void enableQresync() throws IOException, MessagingException {
-        if (!qresyncEnabled && hasCapability(Capabilities.ENABLE) && hasCapability(Capabilities.QRESYNC)
-                && K9MailLib.shouldUseQresync()) {
-            executeSimpleCommand(String.format("%s %s %s", Capabilities.ENABLE, Capabilities.CONDSTORE,
+        if (!qresyncEnabled && hasCapability(Capabilities.QRESYNC) && K9MailLib.shouldUseQresync()) {
+            executeSimpleCommand(String.format("%s %s %s", Commands.ENABLE, Capabilities.CONDSTORE,
                     Capabilities.QRESYNC));
             qresyncEnabled = true;
         }
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
@@ -603,7 +603,7 @@ public class ImapConnection {
     }
 
     private void enableQresyncIfAvailable() throws IOException, MessagingException {
-        if (hasCapability(Capabilities.ENABLE) && hasCapability(Capabilities.QRESYNC)) {
+        if (hasCapability(Capabilities.ENABLE) && hasCapability(Capabilities.QRESYNC) && K9MailLib.shouldUseQresync()) {
             executeSimpleCommand(String.format("%s %s %s", Capabilities.ENABLE, Capabilities.CONDSTORE,
                     Capabilities.QRESYNC));
         }
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
@@ -687,7 +687,8 @@ public class ImapConnection {
         return isListResponse && hierarchyDelimiterValid;
     }
 
-    boolean hasCapability(String capability) {
+    boolean hasCapability(String capability) throws IOException, MessagingException {
+        requestCapabilitiesIfNecessary();
         return capabilities.contains(capability.toUpperCase(Locale.US));
     }
 
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
@@ -77,6 +77,7 @@ public class ImapConnection {
     private Exception stacktraceForClose;
     private boolean open = false;
     private boolean retryXoauth2WithNewToken = true;
+    private boolean qresyncEnabled = false;
 
 
     public ImapConnection(ImapSettings settings, TrustedSocketFactory socketFactory,
@@ -130,7 +131,7 @@ public class ImapConnection {
             extractOrRequestCapabilities(responses);
 
             enableCompressionIfRequested();
-            enableQresyncIfAvailable();
+            enableQresync();
 
             retrievePathPrefixIfNecessary();
             retrievePathDelimiterIfNecessary();
@@ -602,10 +603,12 @@ public class ImapConnection {
         }
     }
 
-    private void enableQresyncIfAvailable() throws IOException, MessagingException {
-        if (hasCapability(Capabilities.ENABLE) && hasCapability(Capabilities.QRESYNC) && K9MailLib.shouldUseQresync()) {
+    void enableQresync() throws IOException, MessagingException {
+        if (!qresyncEnabled && hasCapability(Capabilities.ENABLE) && hasCapability(Capabilities.QRESYNC)
+                && K9MailLib.shouldUseQresync()) {
             executeSimpleCommand(String.format("%s %s %s", Capabilities.ENABLE, Capabilities.CONDSTORE,
                     Capabilities.QRESYNC));
+            qresyncEnabled = true;
         }
     }
 
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
@@ -130,7 +130,7 @@ public class ImapConnection {
             extractOrRequestCapabilities(responses);
 
             enableCompressionIfRequested();
-            enableExtensionsIfAvailable();
+            enableQresyncIfAvailable();
 
             retrievePathPrefixIfNecessary();
             retrievePathDelimiterIfNecessary();
@@ -602,14 +602,10 @@ public class ImapConnection {
         }
     }
 
-    private void enableExtensionsIfAvailable() throws IOException, MessagingException {
-        if (hasCapability(Capabilities.ENABLE)) {
-            if (hasCapability(Capabilities.CONDSTORE)) {
-                executeSimpleCommand(String.format("%s %s", Capabilities.ENABLE, Capabilities.CONDSTORE));
-            }
-            if (hasCapability(Capabilities.QRESYNC)) {
-                executeSimpleCommand(String.format("%s %s", Capabilities.ENABLE, Capabilities.QRESYNC));
-            }
+    private void enableQresyncIfAvailable() throws IOException, MessagingException {
+        if (hasCapability(Capabilities.ENABLE) && hasCapability(Capabilities.QRESYNC)) {
+            executeSimpleCommand(String.format("%s %s %s", Capabilities.ENABLE, Capabilities.CONDSTORE,
+                    Capabilities.QRESYNC));
         }
     }
 
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -126,22 +126,28 @@ public class ImapFolder extends Folder<ImapMessage> {
         }
     }
 
-    public void open(int mode, long cachedUidValidity, long cachedHighestModSeq, List<Long> cachedUids)
-            throws MessagingException {
-        internalOpen(mode, cachedUidValidity, cachedHighestModSeq, cachedUids, false);
+    public QresyncResponse openUsingQresync(int mode, long cachedUidValidity, long cachedHighestModSeq)
+        throws MessagingException {
+        return openUsingQresync(mode, cachedUidValidity, cachedHighestModSeq, 0);
+    }
+
+    public QresyncResponse openUsingQresync(int mode, long cachedUidValidity, long cachedHighestModSeq,
+            long smallestUid) throws MessagingException {
+        SelectOrExamineResponse response = internalOpen(mode, cachedUidValidity, cachedHighestModSeq, smallestUid, false);
 
         if (messageCount == -1) {
             throw new MessagingException("Did not find message count during open");
         }
+        return response.getQresyncResponse();
     }
 
     SelectOrExamineResponse internalOpen(int mode, long cachedUidValidity, long cachedHighestModSeq,
-            List<Long> cachedUids, boolean reuseConnection) throws MessagingException {
+            Long smallestUid, boolean reuseConnection) throws MessagingException {
         if (reuseConnection && isOpen() && this.mode == mode) {
             // Make sure the connection is valid. If it's not we'll close it down and continue
             // on to get a new one.
             try {
-                return SelectOrExamineResponse.parse(executeSimpleCommand(Commands.NOOP), store.getPermanentFlagsIndex());
+                return SelectOrExamineResponse.parse(executeSimpleCommand(Commands.NOOP), this);
             } catch (IOException ioe) {
                 /* don't throw */ ioExceptionHandler(connection, ioe);
             }
@@ -161,7 +167,7 @@ public class ImapFolder extends Folder<ImapMessage> {
             SelectOrExamineCommand command;
             if (connection.isQresyncCapable()) {
                 command = SelectOrExamineCommand.createWithQresyncParameter(mode, escapedFolderName, cachedUidValidity,
-                        cachedHighestModSeq, cachedUids);
+                        cachedHighestModSeq, smallestUid);
             } else if (connection.isCondstoreCapable()) {
                 command = SelectOrExamineCommand.createWithCondstoreParameter(mode, escapedFolderName);
             } else {
@@ -169,8 +175,7 @@ public class ImapFolder extends Folder<ImapMessage> {
             }
 
             SelectOrExamineResponse response = SelectOrExamineResponse.parse(
-                    executeSimpleCommand(command.createCommandString()),
-                    store.getPermanentFlagsIndex());
+                    executeSimpleCommand(command.createCommandString()), this);
 
             /*
              * If the command succeeds we expect the folder has been opened read-write unless we
@@ -856,25 +861,7 @@ public class ImapFolder extends Folder<ImapMessage> {
             message.setModSeq(modSeq);
         }
         if (fetchList.containsKey("FLAGS")) {
-            ImapList flags = fetchList.getKeyedList("FLAGS");
-            if (flags != null) {
-                for (int i = 0, count = flags.size(); i < count; i++) {
-                    String flag = flags.getString(i);
-                    if (flag.equalsIgnoreCase("\\Deleted")) {
-                        message.setFlagInternal(Flag.DELETED, true);
-                    } else if (flag.equalsIgnoreCase("\\Answered")) {
-                        message.setFlagInternal(Flag.ANSWERED, true);
-                    } else if (flag.equalsIgnoreCase("\\Seen")) {
-                        message.setFlagInternal(Flag.SEEN, true);
-                    } else if (flag.equalsIgnoreCase("\\Flagged")) {
-                        message.setFlagInternal(Flag.FLAGGED, true);
-                    } else if (flag.equalsIgnoreCase("$Forwarded")) {
-                        message.setFlagInternal(Flag.FORWARDED, true);
-                        /* a message contains FORWARDED FLAG -> so we can also create them */
-                        store.getPermanentFlagsIndex().add(Flag.FORWARDED);
-                    }
-                }
-            }
+            ImapUtility.setMessageFlags(fetchList, message, store);
         }
 
         if (fetchList.containsKey("INTERNALDATE")) {
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -182,6 +182,10 @@ public class ImapFolder extends Folder<ImapMessage> {
             if (response.hasOpenMode()) {
                 this.mode = response.getOpenMode();
             }
+            if (response == null) {
+                // This shouldn't happen
+                return null;
+            }
             exists = true;
             uidValidity = response.getUidValidity();
             highestModSeq = response.getHighestModSeq();
@@ -434,7 +438,7 @@ public class ImapFolder extends Folder<ImapMessage> {
         return !(highestModSeq == INVALID_HIGHEST_MOD_SEQ);
     }
 
-    public boolean supportsQresync() throws IOException, MessagingException {
+    boolean supportsQresync() throws IOException, MessagingException {
         return supportsModSeq() && connection.isQresyncCapable();
     }
 
@@ -1355,7 +1359,7 @@ public class ImapFolder extends Folder<ImapMessage> {
         return getName().hashCode();
     }
 
-    private ImapStore getStore() {
+    ImapStore getStore() {
         return store;
     }
 
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -52,8 +52,8 @@ public class ImapFolder extends Folder<ImapMessage> {
     };
     private static final int MORE_MESSAGES_WINDOW_SIZE = 500;
     private static final int FETCH_WINDOW_SIZE = 100;
-    static final int INVALID_UID_VALIDITY = -1;
-    static final int INVALID_HIGHEST_MOD_SEQ = -1;
+    public static final int INVALID_UID_VALIDITY = -1;
+    public static final int INVALID_HIGHEST_MOD_SEQ = -1;
 
 
     protected volatile int messageCount = -1;
@@ -177,13 +177,13 @@ public class ImapFolder extends Folder<ImapMessage> {
              * If the command succeeds we expect the folder has been opened read-write unless we
              * are notified otherwise in the responses.
              */
-            this.mode = mode;
-            if (response.hasOpenMode()) {
-                this.mode = response.getOpenMode();
-            }
             if (response == null) {
                 // This shouldn't happen
                 return null;
+            }
+            this.mode = mode;
+            if (response.hasOpenMode()) {
+                this.mode = response.getOpenMode();
             }
             exists = true;
             uidValidity = response.getUidValidity();
@@ -733,13 +733,9 @@ public class ImapFolder extends Folder<ImapMessage> {
                 String uid = fetchList.getKeyedString("UID");
                 long msgSeq = response.getLong(0);
                 if (uid != null) {
-                    try {
-                        msgSeqUidMap.put(msgSeq, uid);
-                        if (K9MailLib.isDebug()) {
-                            Timber.v("Stored uid '%s' for msgSeq %d into map", uid, msgSeq);
-                        }
-                    } catch (Exception e) {
-                        Timber.e("Unable to store uid '%s' for msgSeq %d", uid, msgSeq);
+                    msgSeqUidMap.put(msgSeq, uid);
+                    if (K9MailLib.isDebug()) {
+                        Timber.v("Stored uid '%s' for msgSeq %d into map", uid, msgSeq);
                     }
                 }
 
@@ -1263,14 +1259,14 @@ public class ImapFolder extends Folder<ImapMessage> {
 
         try {
             List<ImapResponse> expungeResponses = executeSimpleCommand("EXPUNGE");
-            handleExpungeResponses(expungeResponses);
+            updateHighestModSeqFromResponses(expungeResponses);
             return ImapUtility.extractVanishedUids(expungeResponses);
         } catch (IOException ioe) {
             throw ioExceptionHandler(connection, ioe);
         }
     }
 
-    private void handleExpungeResponses(List<ImapResponse> imapResponses) {
+    private void updateHighestModSeqFromResponses(List<ImapResponse> imapResponses) {
         for (ImapResponse imapResponse : imapResponses) {
             Long highestModSeq = ImapUtility.extractHighestModSeq(imapResponse);
             if (highestModSeq != null) {
@@ -1368,8 +1364,12 @@ public class ImapFolder extends Folder<ImapMessage> {
         return getName().hashCode();
     }
 
-    ImapStore getStore() {
+    private ImapStore getStore() {
         return store;
+    }
+
+    Set<Flag> getPermanentFlags() {
+        return getStore().getPermanentFlagsIndex();
     }
 
     protected String getLogId() {
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -126,12 +126,12 @@ public class ImapFolder extends Folder<ImapMessage> {
         }
     }
 
-    public QresyncResponse openUsingQresync(int mode, long cachedUidValidity, long cachedHighestModSeq)
+    public QresyncResponse open(int mode, long cachedUidValidity, long cachedHighestModSeq)
         throws MessagingException {
-        return openUsingQresync(mode, cachedUidValidity, cachedHighestModSeq, 0);
+        return open(mode, cachedUidValidity, cachedHighestModSeq, 0);
     }
 
-    public QresyncResponse openUsingQresync(int mode, long cachedUidValidity, long cachedHighestModSeq,
+    public QresyncResponse open(int mode, long cachedUidValidity, long cachedHighestModSeq,
             long smallestUid) throws MessagingException {
         SelectOrExamineResponse response = internalOpen(mode, cachedUidValidity, cachedHighestModSeq, smallestUid, false);
 
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -202,27 +202,6 @@ public class ImapFolder extends Folder<ImapMessage> {
         }
     }
 
-    void updateHighestModSeq(List<ImapMessage> changedMessages) throws IOException, MessagingException {
-        if (!supportsModSeq()) {
-            return;
-        }
-
-        FetchProfile fp = new FetchProfile();
-        fp.add(FetchProfile.Item.MODSEQ);
-        fetch(changedMessages, fp, null);
-
-        long newHighestModSeq = 0;
-        for (ImapMessage message : changedMessages) {
-            long modSeq = message.getModSeq();
-            if (modSeq > newHighestModSeq) {
-                newHighestModSeq = modSeq;
-            }
-        }
-        if (newHighestModSeq != 0) {
-            highestModSeq = newHighestModSeq;
-        }
-    }
-
     @Override
     public boolean isOpen() {
         return connection != null;
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -121,7 +121,7 @@ public class ImapFolder extends Folder<ImapMessage> {
 
     @Override
     public void open(int mode) throws MessagingException {
-        internalOpen(mode, INVALID_UID_VALIDITY, INVALID_HIGHEST_MOD_SEQ, true);
+        internalOpen(mode, INVALID_UID_VALIDITY, INVALID_HIGHEST_MOD_SEQ);
 
         if (messageCount == -1) {
             throw new MessagingException("Did not find message count during open");
@@ -130,7 +130,7 @@ public class ImapFolder extends Folder<ImapMessage> {
 
     public QresyncParamResponse open(int mode, long cachedUidValidity, long cachedHighestModSeq)
             throws MessagingException {
-        SelectOrExamineResponse response = internalOpen(mode, cachedUidValidity, cachedHighestModSeq, false);
+        SelectOrExamineResponse response = internalOpen(mode, cachedUidValidity, cachedHighestModSeq);
 
         if (messageCount == -1) {
             throw new MessagingException("Did not find message count during open");
@@ -138,9 +138,9 @@ public class ImapFolder extends Folder<ImapMessage> {
         return response.getQresyncParamResponse();
     }
 
-    SelectOrExamineResponse internalOpen(int mode, long cachedUidValidity, long cachedHighestModSeq,
-            boolean reuseConnection) throws MessagingException {
-        if (reuseConnection && isOpen() && this.mode == mode) {
+    SelectOrExamineResponse internalOpen(int mode, long cachedUidValidity, long cachedHighestModSeq)
+            throws MessagingException {
+        if (isOpen() && this.mode == mode) {
             // Make sure the connection is valid. If it's not we'll close it down and continue
             // on to get a new one.
             try {
@@ -163,6 +163,7 @@ public class ImapFolder extends Folder<ImapMessage> {
             String escapedFolderName = ImapUtility.encodeString(encodedFolderName);
             SelectOrExamineCommand command;
             if (connection.isQresyncCapable() && K9MailLib.shouldUseQresync()) {
+                connection.enableQresync();
                 command = SelectOrExamineCommand.createWithQresyncParameter(mode, escapedFolderName, cachedUidValidity,
                         cachedHighestModSeq);
             } else if (connection.isCondstoreCapable() && K9MailLib.shouldUseCondstore()) {
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -128,7 +128,7 @@ public class ImapFolder extends Folder<ImapMessage> {
         }
     }
 
-    public QresyncParamResponse open(int mode, long cachedUidValidity, long cachedHighestModSeq)
+    public QresyncParamResponse openUsingQresyncParam(int mode, long cachedUidValidity, long cachedHighestModSeq)
             throws MessagingException {
         SelectOrExamineResponse response = internalOpen(mode, cachedUidValidity, cachedHighestModSeq);
 
@@ -439,8 +439,8 @@ public class ImapFolder extends Folder<ImapMessage> {
         return !(highestModSeq == INVALID_HIGHEST_MOD_SEQ);
     }
 
-    boolean supportsQresync() throws IOException, MessagingException {
-        return supportsModSeq() && connection.isQresyncCapable();
+    boolean doesConnectionSupportQresync() throws IOException, MessagingException {
+        return connection.isQresyncCapable();
     }
 
     public long getUidValidity() {
@@ -527,6 +527,9 @@ public class ImapFolder extends Folder<ImapMessage> {
 
     public void fetchChangedMessageFlagsUsingCondstore(List<ImapMessage> messages, long modseq,
             MessageRetrievalListener<ImapMessage> listener) throws MessagingException {
+        if (messages == null || messages.isEmpty()) {
+            return;
+        }
         checkOpen();
 
         List<Long> uids = new ArrayList<>(messages.size());
@@ -753,7 +756,7 @@ public class ImapFolder extends Folder<ImapMessage> {
                 }
 
                 if (listener != null) {
-                    listener.messageStarted(uid, messageNumber++, messageMap.size());
+                    listener.messageStarted(uid, messageNumber, messageMap.size());
                 }
 
                 ImapMessage imapMessage = (ImapMessage) message;
@@ -775,6 +778,7 @@ public class ImapFolder extends Folder<ImapMessage> {
                 if (listener != null) {
                     listener.messageFinished(imapMessage, messageNumber, messageMap.size());
                 }
+                messageNumber++;
             } else {
                 handleUntaggedResponse(response);
             }
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -955,6 +955,14 @@ public class ImapFolder extends Folder<ImapMessage> {
                     Timber.d("Got untagged EXPUNGE with messageCount %d for %s", messageCount, getLogId());
                 }
             }
+
+            if (equalsIgnoreCase(response.get(0), "VANISHED") && messageCount > 0) {
+                List<String> vanishedUids = ImapUtility.extractVanishedUids(Collections.singletonList(response));
+                messageCount -= vanishedUids.size();
+                if (K9MailLib.isDebug()) {
+                    Timber.d("Got untagged VANISHED with messageCount %d for %s", messageCount, getLogId());
+                }
+            }
         }
     }
 
@@ -1253,25 +1261,20 @@ public class ImapFolder extends Folder<ImapMessage> {
 
         try {
             List<ImapResponse> expungeResponses = executeSimpleCommand("EXPUNGE");
-            return handleExpungeResponses(expungeResponses);
+            handleExpungeResponses(expungeResponses);
+            return ImapUtility.extractVanishedUids(expungeResponses);
         } catch (IOException ioe) {
             throw ioExceptionHandler(connection, ioe);
         }
     }
 
-    private List<String> handleExpungeResponses(List<ImapResponse> imapResponses) {
-        List<String> expungedUids = new ArrayList<>();
+    private void handleExpungeResponses(List<ImapResponse> imapResponses) {
         for (ImapResponse imapResponse : imapResponses) {
-            if (imapResponse.getTag() == null && ImapResponseParser.equalsIgnoreCase(imapResponse.get(0), "VANISHED")) {
-                expungedUids = ImapUtility.getImapSequenceValues(imapResponse.getString(1));
-            } else {
-                Long highestModSeq = ImapUtility.extractHighestModSeq(imapResponse);
-                if (highestModSeq != null) {
-                    this.highestModSeq = highestModSeq;
-                }
+            Long highestModSeq = ImapUtility.extractHighestModSeq(imapResponse);
+            if (highestModSeq != null) {
+                this.highestModSeq = highestModSeq;
             }
         }
-        return expungedUids;
     }
 
     @Override
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -162,10 +162,10 @@ public class ImapFolder extends Folder<ImapMessage> {
             String encodedFolderName = folderNameCodec.encode(getPrefixedName());
             String escapedFolderName = ImapUtility.encodeString(encodedFolderName);
             SelectOrExamineCommand command;
-            if (connection.isQresyncCapable()) {
+            if (connection.isQresyncCapable() && K9MailLib.shouldUseQresync()) {
                 command = SelectOrExamineCommand.createWithQresyncParameter(mode, escapedFolderName, cachedUidValidity,
                         cachedHighestModSeq);
-            } else if (connection.isCondstoreCapable()) {
+            } else if (connection.isCondstoreCapable() && K9MailLib.shouldUseCondstore()) {
                 command = SelectOrExamineCommand.createWithCondstoreParameter(mode, escapedFolderName);
             } else {
                 command = SelectOrExamineCommand.create(mode, escapedFolderName);
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
@@ -548,22 +548,37 @@ public class ImapFolder extends Folder<ImapMessage> {
         return getMessages(start, end, earliestDate, false, listener);
     }
 
-    public List<ImapMessage> getChangedMessagesUsingCondstore(long modseq) throws MessagingException {
+    public void fetchChangedMessageFlagsUsingCondstore(List<Long> uids, long modseq,
+            MessageRetrievalListener<ImapMessage> listener) throws MessagingException {
         checkOpen();
-        if (!supportsModSeq()) {
-            String errorMessage = "Folder %s does not support modification sequences";
-            throw new MessagingException(String.format(Locale.US, errorMessage, name));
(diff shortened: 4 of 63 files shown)


k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
```diff
@@ -505,6 +547,24 @@ public ImapMessage getMessage(String uid) throws MessagingException {
         return getMessages(start, end, earliestDate, false, listener);
     }
 
+    public List<ImapMessage> getChangedMessagesUsingCondstore(long modseq) throws MessagingException {
+        checkOpen();
+        if (!supportsModSeq()) {
+            throw new MessagingException(String.format(Locale.US, "Folder %s does not support modification " +
```
philipwhiuk: Put the string on a new line so you don't have to build a new one every time.

rhari991:
Currently, the sync process is as follows :

1) Get the UIDs of the most recent n messages on the server, where n is the number of messages displayed to the user (via UID SEARCH). From these, messages which arrived on a date earlier than the earliest poll date are removed.

2) Delete local messages which are not present in the above group. Then check and update whether more messages are present in the mailbox (via one or more UID SEARCH). 

3) Divide the messages into two groups - one for messages that need to be downloaded fully (if not present locally) and one for syncing flags (if it is present locally).

4) The unsynced messages are fetched.
    a) Initially, the flags and envelope for all unsynced messages are fetched.
    b) The messages are then divided into large and small message groups depending on the maximum auto-download message size.
    c) Download all the small messages fully (using grouped UID FETCH requests).
    d) Download partial/sane bodies for large messages (one UID FETCH per message, and multiple if structure is present).

5) The flags for all the messages present in the local store are fetched and updated.

When CONDSTORE is used :

- In step 1, the UIDs for all messages that are present/going to be present in the mailbox are fetched. This is necessary in order to sync remote deletions and some use cases such as the user pressing the "Load more messages" button (historical data needs to be fetched). Because of these reasons, this step cannot be optimized using the previously stored mod-sequence.

- Related to the point above, one optimization could be to remove the "Load more messages" button and instead automatically download more messages if the user scrolls to the bottom of the message list. This should feel more natural, and will free us from having to make a network request to check if more messages are available.

- For step 5, the optimization that can be done is to only fetch flags for those messages whose contents have changed. This is done by sending an extra UID SEARCH command with the previously known mod sequence and then syncing the flags for the returned message uids only.

- An extra UID FETCH command also needs to be issued every time we get a response through IDLE in order to update the highest modseq value.

Valodim:
description matches with my understanding of the algorithm.

Regarding the other points:

- Fetching all UIDs for finding deleted messages - shouldn't that be solved by QRESYNC?

- I don't understand how a fetch-on-scroll mechanism would free us from that network request. Either way, it would also be too much of a semantic shift to bring in here.

- Regarding IDLE, I think we are not doing us any favors trying to be clever there. Just open a connection, then when anything happens just trigger the usual sync process.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/PermanentFlagsResponse.java
```diff
@@ -22,10 +22,6 @@ private PermanentFlagsResponse(Set<Flag> flags, boolean canCreateKeywords) {
     }
 
     public static PermanentFlagsResponse parse(ImapResponse response) {
-        if (response.isTagged() || !equalsIgnoreCase(response.get(0), Responses.OK) || !response.isList(1)) {
```
Valodim: can this never happen? if it can't, maybe a `throw new IllegalArgumentException` would be appropriate?

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/QresyncResponse.java
```diff
@@ -0,0 +1,54 @@
+package com.fsck.k9.mail.store.imap;
+
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.fsck.k9.mail.MessagingException;
+
+
+public class QresyncResponse {
```
Valodim: there is no `QRESYNC` command. is this a response to a `SELECT` with `QRESYNC` parameter? if so, the name should reflect that

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/QresyncResponse.java
```diff
@@ -0,0 +1,54 @@
...
+    public static QresyncResponse parse(List<ImapResponse> imapResponses, ImapFolder folder) throws MessagingException {
+        QresyncResponse qresyncResponse = new QresyncResponse();
+        for (ImapResponse imapResponse : imapResponses) {
+            handleExpungedUids(imapResponse, qresyncResponse);
+            handleModifiedMessages(imapResponse, qresyncResponse, folder);
+        }
+        return qresyncResponse;
+    }
+
+    private static void handleExpungedUids(ImapResponse imapResponse, QresyncResponse qresyncResponse) {
```
Valodim: why make this static when you give an instance of the class as a parameter?

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineCommand.java
```diff
@@ -0,0 +1,63 @@
...
+
+import static com.fsck.k9.mail.Folder.OPEN_MODE_RW;
+
+
+class SelectOrExamineCommand {
+
+    private static final long INVALID_UIDVALIDITY = -1L;
+    private static final long INVALID_HIGHESTMODSEQ = -1L;
+    private static final long INVALID_SMALLEST_UID = -1L;
+
```
Valodim: nitpick: two linebreaks between static and non-static members

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineCommand.java
```diff
@@ -0,0 +1,63 @@
+package com.fsck.k9.mail.store.imap;
+
+
+import static com.fsck.k9.mail.Folder.OPEN_MODE_RW;
+
+
+class SelectOrExamineCommand {
+
```
Valodim: nitpick: no empty line necessary here

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineCommand.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final long INVALID_HIGHESTMODSEQ = -1L;
+    private static final long INVALID_SMALLEST_UID = -1L;
+
+    private int mode;
+    private String escapedFolderName;
+    private boolean useCondstore;
+    private long cachedUidValidity;
+    private long cachedHighestModSeq;
+    private long smallestUid;
+
```
Valodim: nitpick: two linebreaks between fields and methods

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineCommand.java
```diff
@@ -0,0 +1,63 @@
...
+                INVALID_SMALLEST_UID);
+    }
+
+    static SelectOrExamineCommand createWithCondstoreParameter(int mode, String escapedFolderName) {
+        return new SelectOrExamineCommand(mode, escapedFolderName, true, INVALID_UIDVALIDITY, INVALID_HIGHESTMODSEQ,
+                INVALID_SMALLEST_UID);
+    }
+
+    static SelectOrExamineCommand createWithQresyncParameter(int mode, String escapedFolderName,
+            long cachedUidValidity, long cachedHighestModSeq, long smallestUid) {
```
Valodim: why is `useCondstore` false if we use QRESYNC? seems counter-intuitive

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineResponse.java
```diff
@@ -1,52 +1,121 @@
...
+        PermanentFlagsResponse permanentFlagsResponse = PermanentFlagsResponse.parse(imapResponse);
+        if (permanentFlagsResponse == null) {
+            return;
+        }
+
+        permanentFlags.addAll(permanentFlagsResponse.getFlags());
+        selectOrExamineResponse.canCreateKeywords = permanentFlagsResponse.canCreateKeywords();
+    }
+
+    private static Boolean isModeReadWrite(ImapResponse imapResponse) {
```
Valodim: it's not obvious to me under what circumstances this might return null, as opposed to false. if it happens when the response doesn't contain a valid indicator, should that ever happen? if you think this is the right signature for this method, the name should reflect that failure case, e.g. by adding `IfAvailable`

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,268 @@
...
+
+            Timber.v("SYNC: About to get local folder %s and open it", folderName);
+            localFolder = SyncUtils.getOpenedLocalFolder(account, folderName);
+            localFolder.updateLastUid();
+
+            Store remoteStore = account.getRemoteStore();
+            Timber.v("SYNC: About to get remote folder %s", folderName);
+            Folder remoteFolder = remoteStore.getFolder(folderName);
+            if (!(remoteFolder instanceof ImapFolder)) {
+                throw new MessagingException("A non-IMAP account was provided to ImapSyncInteractor");
```
Valodim: this seems like it's a bug if it ever happens, i.e. should crash with IllegalArgumentException

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,268 @@
...
+            }
+            imapFolder = (ImapFolder) remoteFolder;
+
+            if (!SyncUtils.verifyOrCreateRemoteSpecialFolder(account, folderName, imapFolder, listener, controller)) {
+                return;
+            }
+
+            QresyncResponse qresyncResponse;
+            Timber.v("SYNC: About to open remote IMAP folder %s", folderName);
+            List<String> uids = localFolder.getAllMessageUids();
```
Valodim: if I'm seeing this right, we do nothing with these uids besides getting the smallest one. a specialized query should be much more performant here.

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,268 @@
...
+                qresyncResponse = imapFolder.open(Folder.OPEN_MODE_RW, localFolder.getUidValidity(),
+                        localFolder.getHighestModSeq(), smallestUid);
+            }
+
+            boolean qresyncEnabled = qresyncResponse != null;
+            List<String> expungedUids = new ArrayList<>();
+            if (Expunge.EXPUNGE_ON_POLL == account.getExpungePolicy()) {
+                Timber.d("SYNC: Expunging folder %s:%s", account.getDescription(), folderName);
+                if (qresyncEnabled) {
+                    expungedUids = imapFolder.expungeUsingQresync();
```
Valodim: so this code path is triggered when QresyncResponse is non-null, but doesn't actually depend on that response? I don't fully know the context here but that seems weird

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,268 @@
...
+            int remoteMessageCount = imapFolder.getMessageCount();
+            if (remoteMessageCount < 0) {
+                throw new IllegalStateException("Message count " + remoteMessageCount + " for folder " + folderName);
+            } else {
+                Timber.v("SYNC: Remote message count for folder %s is %d", folderName, remoteMessageCount);
+            }
+
+            handleUidValidity();
+            int newMessages;
+            if (!qresyncEnabled) {
```
Valodim: [[ marker to resume review later at this point - train ride is over :) ]]

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,268 @@
...
+
+            localFolder.setUidValidity(imapFolder.getUidValidity());
+            updateHighestModSeqIfNecessary(localFolder, imapFolder);
+
+            int unreadMessageCount = localFolder.getUnreadMessageCount();
+            for (MessagingListener l : controller.getListeners()) {
+                l.folderStatusChanged(account, folderName, unreadMessageCount);
+            }
+
+            /* Notify listeners that we're finally done. */
```
Valodim: this comment seems out of place :)

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineResponse.java
```diff
@@ -1,52 +1,121 @@
...
+
+    public static SelectOrExamineResponse parse(List<ImapResponse> imapResponses, ImapFolder folder) throws IOException,
+            MessagingException {
+        SelectOrExamineResponse selectOrExamineResponse = new SelectOrExamineResponse();
+        for (ImapResponse imapResponse : imapResponses) {
+            if (imapResponse.isTagged() || !equalsIgnoreCase(imapResponse.get(0), Responses.OK)
+                    || !imapResponse.isList(1)) {
+                continue;
+            }
+            handleUidValidity(imapResponse, selectOrExamineResponse);
```
Valodim: these methods all just `parse`, they don't `handle`

k9mail/src/main/java/com/fsck/k9/controller/NonQresyncSyncInteractor.java
```diff
@@ -0,0 +1,161 @@
...
+import com.fsck.k9.mail.Message;
+import com.fsck.k9.mail.MessageRetrievalListener;
+import com.fsck.k9.mail.MessagingException;
+import com.fsck.k9.mail.store.imap.ImapFolder;
+import com.fsck.k9.mail.store.imap.ImapMessage;
+import com.fsck.k9.mailstore.LocalFolder;
+import timber.log.Timber;
+
+
+class NonQresyncSyncInteractor {
```
rhari991: This covers the CONDSTORE and regular IMAP use cases, since they differ only in the way they handle local flag changes. The QRESYNC has a very different sequence of commands, so it has a separate class.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolderPusher.java
```diff
@@ -504,6 +503,12 @@ private int processUntaggedResponse(long oldMessageCount, ImapResponse response,
                             Timber.d("Got untagged EXPUNGE for msgseq %d for %s", msgSeq, getLogId());
                         }
 
+                        if (K9MailLib.isDebug()) {
+                            Timber.d("Scheduling removal of message with seq %d due to expunge", msgSeq);
+                        }
+
+                        removeMsgUids.add("thisisahack");
```
rhari991: Whoops, I was going to add a comment about this, but it slopped my mind. This code does not trigger the removal of messages from the local store when an EXPUNGE message is received through IDLE. This is because EXPUNGE responses contain only the sequence number of the corresponding messages, and we depend on a mapping of sequence numbers to UIDs to get the corresponding UID. This mapping is populated through previously received FETCH responses. When CONDSTORE/QRESYNC is used, we send FETCH requests for only a selected subset of messages (the ones which have changed), so the UIDs of the remaining ones are not available and cannot be expunged from the local store.

I suppose the simplest way to fix it would be to use a UID SEARCH command to get the corresponding UID. However, since my next task is to rewrite the IDLE-handling code, which will invoIve removing most of this, I decided to do that instead.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/PermanentFlagsResponse.java
rhari991: I moved it to `SelectOrExamineResponse` since the same check was being made in 3 places.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineCommand.java
rhari991: It's not really too important since the parameter is not used if QRESYNC is used, but I see your point. I'll change this.

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
rhari991: When QRESYNC is enabled, an EXPUNGE command returns all the expunged message UIDs. These need to be taken into account when syncing remote deletions in addition to those returned by the SELECT-QRESYNC command, since the EXPUNGE occurred after the mailbox was opened.

k9mail/src/main/java/com/fsck/k9/mailstore/LocalFolder.java
```diff
@@ -82,6 +83,8 @@
     private long mFolderId = -1;
     private int mVisibleLimit = -1;
     private String prefId = null;
+    private long uidValidity = -1;
+    private long highestModSeq = -1;
```
rhari991: Done

k9mail/src/main/java/com/fsck/k9/controller/LegacySyncInteractor.java
```diff
@@ -0,0 +1,272 @@
...
+import com.fsck.k9.mail.Message;
+import com.fsck.k9.mail.MessagingException;
+import com.fsck.k9.mail.Store;
+import com.fsck.k9.mailstore.LocalFolder;
+import com.fsck.k9.mailstore.LocalFolder.MoreMessages;
+import com.fsck.k9.mailstore.LocalMessage;
+import timber.log.Timber;
+
+
+class LegacySyncInteractor {
```
rhari991: Added

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapMessage.java
```diff
@@ -10,6 +10,9 @@
 
 
 public class ImapMessage extends MimeMessage {
+
+    private long modSeq;
```
rhari991: This is actually not needed anymore, I've removed it now.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapUtility.java
```diff
@@ -193,4 +196,45 @@ public static String join(String delimiter, Collection<? extends Object> tokens)
...
+            }
+        }
+    }
+
+    static Long extractHighestModSeq(ImapResponse imapResponse) {
+        for (Object token : imapResponse) {
+            if (token instanceof ImapList) {
+                ImapList list = (ImapList) token;
+                if (list.size() < 2 || !(equalsIgnoreCase(list.get(0), Responses.HIGHESTMODSEQ)
+                        || equalsIgnoreCase(list.get(1), Responses.NOMODSEQ)) ||
```
rhari991: Whoops, that was an oversight. Fixed now.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/QresyncResponse.java
rhari991: Changed to `QresyncParamResponse`.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/QresyncResponse.java
rhari991: Modified

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineResponse.java
rhari991: Modified.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineResponse.java
rhari991: I'm not sure when null is returned either. I've added `IfAvailable`.

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
rhari991: Changed

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
rhari991: Removed.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineCommand.java
rhari991: Removed.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineCommand.java
rhari991: I see only one.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineCommand.java
rhari991: I see only one.

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
rhari991: I've removed this, since we don't use this parameter anymore.

rhari991:
@Valodim made the changes.

k9mail/src/main/java/com/fsck/k9/controller/MessagingController.java
```diff
@@ -726,899 +709,65 @@ public void run() {
...
-                    System.currentTimeMillis());
-
-        } finally {
-            if (providedRemoteFolder == null) {
-                closeFolder(remoteFolder);
-            }
-
-            closeFolder(tLocalFolder);
+        String storeUri = account.getStoreUri();
+        if (storeUri.startsWith("imap")) {
```
Valodim: could better be decided in an `ImapStore.isStoreUriImap` method

k9mail/src/main/java/com/fsck/k9/controller/QresyncSyncInteractor.java
```diff
@@ -0,0 +1,128 @@
...
+            } else {
+                messageDownloader.processDownloadedFlags(account, localFolder, imapMessage, headerProgress,
+                        modifiedMessages.size());
+            }
+        }
+
+        Timber.v("SYNC: Received %d new message UIDs in QRESYNC response", remoteMessages.size());
+    }
+
+    private void findOldRemoteMessagesToDownload(List<ImapMessage> remoteMessages) throws MessagingException {
```
rhari991: For QRESYNC, old messages are only downloaded if needed, and are specifically chosen. For the CONDSTORE/plain IMAP case, we take every message in the N most recent messages of the remote mailbox, and filter it down from there.

k9mail/src/main/java/com/fsck/k9/controller/MessageDownloader.java
```diff
@@ -0,0 +1,713 @@
...
+import com.fsck.k9.mailstore.LocalFolder;
+import com.fsck.k9.mailstore.LocalMessage;
+import com.fsck.k9.mailstore.MessageRemovalListener;
+import com.fsck.k9.notification.NotificationController;
+import timber.log.Timber;
+
+import static com.fsck.k9.controller.SyncUtils.modeMismatch;
+
+
+class MessageDownloader {
```
rhari991: I added a flag to control fetching of flags for new messages (used in the QRESYNC case, when we get the flags in the SELECT-QRESYNC response). Other than that, everything is the same.

k9mail/src/main/java/com/fsck/k9/controller/QresyncSyncInteractor.java
```diff
@@ -0,0 +1,128 @@
...
+        findNewRemoteMessagesToDownload(remoteMessages, qresyncParamResponse, messageDownloader);
+        if (imapFolder.getMessageCount() >= localFolder.getVisibleLimit() && imapFolder.getMessageCount() >=
+                (remoteMessages.size() + localFolder.getMessageCount())) {
+            findOldRemoteMessagesToDownload(remoteMessages);
+        }
+
+        return messageDownloader.downloadMessages(account, imapFolder, localFolder, remoteMessages, false, true, false);
+    }
+
+    private void findNewRemoteMessagesToDownload(List<ImapMessage> remoteMessages, QresyncParamResponse qresyncParamResponse,
```
Valodim: `remoteMessages` also isn't a very descriptive name, took me a while to understand what it's used for. `remoteMessageDownloadQueue` maybe?

k9mail/src/main/java/com/fsck/k9/mailstore/StoreSchemaDefinition.java
```diff
@@ -85,6 +81,8 @@ private static void dbCreateDatabaseFromScratch(SQLiteDatabase db) {
                 "unread_count INTEGER, " +
                 "visible_limit INTEGER, " +
                 "status TEXT, " +
+                "uid_validity INTEGER default 0," +
+                "highest_mod_seq INTEGER default 0," +
```
rhari991: changed

k9mail/src/main/java/com/fsck/k9/mailstore/LocalFolder.java
```diff
@@ -971,7 +996,42 @@ public LocalMessage doDbWork(final SQLiteDatabase db) throws WrappedException, U
...
+                        }
+                    } catch (MessagingException e) {
+                        throw new WrappedException(e);
+                    } finally {
+                        Utility.closeQuietly(cursor);
+                    }
+
+                    return result;
+                }
+            }).get(0);
```
rhari991: changed

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/QresyncParamResponse.java
```diff
@@ -0,0 +1,57 @@
...
+    private List<String> expungedUids;
+    private List<ImapMessage> modifiedMessages;
+
+    private QresyncParamResponse(List<ImapResponse> imapResponses, ImapFolder folder) throws MessagingException {
+        expungedUids = new ArrayList<>();
+        modifiedMessages = new ArrayList<>();
+        parse(imapResponses, folder);
+    }
+
+    public static QresyncParamResponse newInstance(List<ImapResponse> imapResponses, ImapFolder folder)
```
rhari991: Changed to `fromSelectOrExamineResponse`

k9mail/src/main/java/com/fsck/k9/controller/QresyncSyncInteractor.java
rhari991: Renamed to `processFetchResponses` and `remoteMessagesToDownload`

k9mail/src/main/java/com/fsck/k9/controller/QresyncSyncInteractor.java
```diff
@@ -0,0 +1,128 @@
...
+        long smallestLocalUid = 1;
+        if (uid != null) {
+            smallestLocalUid = Long.parseLong(uid);
+        }
+
+        for (ImapMessage imapMessage : modifiedMessages) {
+            long remoteMessageUid = Long.parseLong(imapMessage.getUid());
+            Message localMessage = localFolder.getMessage(imapMessage.getUid());
+
+            if (remoteMessageUid < smallestLocalUid) {
```
rhari991: Done

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,260 @@
...
+                    expungedUids = imapFolder.expungeUsingQresync();
+                } else {
+                    imapFolder.expunge();
+                }
+            }
+
+            int remoteMessageCount = imapFolder.getMessageCount();
+            if (remoteMessageCount < 0) {
+                throw new IllegalStateException("Message count " + remoteMessageCount + " for folder " + folderName);
+            } else {
```
rhari991: Changed

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,260 @@
...
+                Timber.v("SYNC: Remote message count for folder %s is %d", folderName, remoteMessageCount);
+            }
+
+            handleUidValidity();
+            int newMessages;
+            if (!qresyncEnabled) {
+                NonQresyncSyncInteractor syncInteractor = new NonQresyncSyncInteractor(account, localFolder, imapFolder,
```
rhari991: Changed to `QresyncExtensionHandler`, `NonQresyncExtensionHandler` and `continueSync`.

k9mail/src/main/java/com/fsck/k9/controller/MessageDownloader.java
```diff
@@ -0,0 +1,713 @@
...
+import com.fsck.k9.mail.MessagingException;
+import com.fsck.k9.mail.Part;
+import com.fsck.k9.mail.internet.MessageExtractor;
+import com.fsck.k9.mailstore.LocalFolder;
+import com.fsck.k9.mailstore.LocalMessage;
+import com.fsck.k9.mailstore.MessageRemovalListener;
+import com.fsck.k9.notification.NotificationController;
+import timber.log.Timber;
+
+import static com.fsck.k9.controller.SyncUtils.modeMismatch;
```
rhari991: fixed

k9mail/src/main/java/com/fsck/k9/controller/MessagingController.java
rhari991: done

k9mail/src/main/java/com/fsck/k9/controller/QresyncSyncInteractor.java
```diff
@@ -0,0 +1,128 @@
...
+        int oldMessagesEnd = oldMessagesStart - 1 + localFolder.getVisibleLimit() - remoteMessages.size() -
+                localFolder.getMessageCount();
+
+        if (oldMessagesEnd > oldMessagesStart) {
+            List<ImapMessage> oldMessages = imapFolder.getMessages(oldMessagesStart, oldMessagesEnd, earliestDate, null);
+            Timber.v("SYNC: Fetched %d old message UIDs for folder %s", oldMessages.size(), folderName);
+            for (ImapMessage oldMessage : oldMessages) {
+                headerProgress.incrementAndGet();
+                for (MessagingListener l : controller.getListeners(listener)) {
+                    l.synchronizeMailboxHeadersProgress(account, folderName, headerProgress.get(), oldMessages.size());
```
rhari991: It does, through the `MessageDownloader.processDownloadedFlags` method. I've move the notification lines back to their methods to make it easier to understand.

k9mail/src/main/java/com/fsck/k9/controller/QresyncSyncInteractor.java
```diff
@@ -0,0 +1,128 @@
...
+                }
+                remoteMessages.add(oldMessage);
+            }
+        }
+
+        for (MessagingListener l : controller.getListeners(listener)) {
+            l.synchronizeMailboxHeadersFinished(account, folderName, headerProgress.get(), remoteMessages.size());
```
rhari991: Whoops, did not notice that. Fixed.

k9mail/src/main/java/com/fsck/k9/mailstore/LocalFolder.java
```diff
@@ -971,7 +1000,40 @@ public LocalMessage doDbWork(final SQLiteDatabase db) throws WrappedException, U
...
+                        open(OPEN_MODE_RO);
+
+                        cursor = db.rawQuery(
+                                "SELECT min(uid) " +
+                                        "FROM messages " +
+                                        "WHERE empty = 0 AND deleted = 0 AND " +
+                                        "folder_id = ? ORDER BY date DESC",
+                                new String[] { Long.toString(mFolderId) });
+
+                        if (cursor.getCount() != 1) {
```
Valodim: `if (!cursor.moveToFirst())`

k9mail/src/main/java/com/fsck/k9/controller/NonQresyncExtensionHandler.java
```diff
@@ -0,0 +1,163 @@
...
+
+        if (account.syncRemoteDeletions()) {
+            imapSyncInteractor.syncRemoteDeletions(findDeletedMessageUids(localUidMap, remoteUidMap));
+        }
+
+        // noinspection UnusedAssignment, free memory early?
+        localUidMap = null;
+
+        if (localFolder.getHighestModSeq() != 0 && imapFolder.supportsModSeq()) {
+            downloadChangedMessageFlags(remoteMessages, messageDownloader);
```
Valodim: this method name should indicate its use of condstore.

k9mail/src/main/java/com/fsck/k9/controller/MessageDownloader.java
```diff
@@ -0,0 +1,711 @@
...
+        downloadLargeMessages(account, remoteFolder, localFolder, largeMessages, progress, unreadBeforeStart,
+                newMessages, todo, fp);
+        largeMessages.clear();
+
+        /*
+         * Refresh the flags for any messages in the local store that we didn't just
+         * download.
+         */
+
+        refreshLocalMessageFlags(account, remoteFolder, localFolder, syncFlagMessages, progress, todo);
```
Valodim: what exactly does this method call depend on? disregarding use of this method in legacy sync, would it be possible to move this over to where flag sync happens with condstore otherwise?

k9mail/src/main/java/com/fsck/k9/controller/NonQresyncExtensionHandler.java
rhari991: Changed

k9mail/src/main/java/com/fsck/k9/controller/MessageDownloader.java
Valodim: that sounds fine, as long as it doesn't cause more roundtrips. code duplication with legacy isn't an issue, just copy that stuff over there untouched.

k9mail/src/main/java/com/fsck/k9/controller/QresyncExtensionHandler.java
```diff
@@ -74,25 +82,28 @@ private void findNewRemoteMessagesToDownload(List<ImapMessage> remoteMessages, Q
 
         for (ImapMessage imapMessage : modifiedMessages) {
             long remoteMessageUid = Long.parseLong(imapMessage.getUid());
-            Message localMessage = localFolder.getMessage(imapMessage.getUid());
-
             if (remoteMessageUid < smallestLocalUid) {
                 continue;
             }
+            Message localMessage = localFolder.getMessage(imapMessage.getUid());
```
Valodim: this call loads the message including mime header parsing and stuff, when all we want is "does it not exist, or is its body not downloaded". no need to do this now, but please add a TODO comment to optimize this call

rhari991: Done

k9mail/src/main/java/com/fsck/k9/controller/MessageDownloader.java
rhari991: Done. I moved the flag synchronization code to a separate class and used it before calling `MessageDownloader`, which now handles only the downloading of new messages.

k9mail/src/main/res/values/strings.xml
```diff
@@ -256,6 +256,10 @@ Please submit bug reports, contribute new features and ask questions at
     <string name="debug_enable_debug_logging_summary">Log extra diagnostic information</string>
     <string name="debug_enable_sensitive_logging_title">Log sensitive information</string>
     <string name="debug_enable_sensitive_logging_summary">May show passwords in logs.</string>
+    <string name="debug_use_condstore_extension_title">Use CONDSTORE</string>
```
Valodim: I think I would word this in a way that implies more heavily that something is being disabled that would usually be enabled. "Don't use CONDSTORE" / "Don't use CONDSTORE even if available"?

k9mail/src/main/java/com/fsck/k9/K9.java
```diff
@@ -51,27 +52,51 @@
 
 
 public class K9 extends Application {
+    public static final String LOG_TAG = "k9";
```
Valodim: why did these things move all over the place?

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/SelectOrExamineResponse.java
```diff
@@ -33,6 +29,39 @@ public static SelectOrExamineResponse newInstance(List<ImapResponse> imapRespons
...
+        if ("READ-ONLY".equalsIgnoreCase(responseCode)) {
+            return Boolean.FALSE;
+        } else if ("READ-WRITE".equalsIgnoreCase(responseCode)) {
+            return Boolean.TRUE;
+        }
+
+        return null;
+    }
+
+    private static boolean isResponseValid(List<ImapResponse> imapResponses) {
```
Valodim: did you do some "order static methods" in android studio? it's fine if you do it but please keep these types of changes in separate commits so they don't make review unnecessarily hard :)

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolderPusher.java
```diff
@@ -120,6 +120,44 @@ private void superHandleUntaggedResponse(ImapResponse response) {
         super.handleUntaggedResponse(response);
     }
 
+    /**
+     * Ensure the DONE continuation is only sent when the IDLE command was sent and hasn't completed yet.
+     */
+    private static class IdleStopper {
```
Valodim: unrelated changes moving things about. again, please keep those in separate commits.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolderPusher.java
rhari991: I've removed the code changes in `ImapFolderPusher` and will add that in a later commit.

Valodim: sounds good. a TODO and maybe an error to the log should be fine as an intermediate state, deletion on push isn't super crucial :+1:

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/Capabilities.java
```diff
@@ -3,7 +3,6 @@
 
 class Capabilities {
     public static final String IDLE = "IDLE";
-    public static final String ENABLE = "ENABLE";
```
Valodim: moving static around. why is this still a thing? :(

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
```diff
@@ -604,9 +604,8 @@ private void enableCompression() throws IOException, MessagingException {
     }
 
     void enableQresync() throws IOException, MessagingException {
-        if (!qresyncEnabled && hasCapability(Capabilities.ENABLE) && hasCapability(Capabilities.QRESYNC)
-                && K9MailLib.shouldUseQresync()) {
-            executeSimpleCommand(String.format("%s %s %s", Capabilities.ENABLE, Capabilities.CONDSTORE,
+        if (!qresyncEnabled && hasCapability(Capabilities.QRESYNC) && K9MailLib.shouldUseQresync()) {
+            executeSimpleCommand(String.format("%s %s %s", Commands.ENABLE, Capabilities.CONDSTORE,
```
Valodim: :+1:

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapUtility.java
```diff
@@ -241,9 +239,13 @@ static Long extractHighestModSeq(ImapResponse imapResponse) {
     static List<String> extractVanishedUids(List<ImapResponse> imapResponses) {
         List<String> uids = new ArrayList<>();
         for (ImapResponse imapResponse : imapResponses) {
-            if (imapResponse.getTag() == null && ImapResponseParser.equalsIgnoreCase(imapResponse.get(0), "VANISHED")
-                    && imapResponse.isString(1)) {
-                uids = ImapUtility.getImapSequenceValues(imapResponse.getString(1));
+            if (imapResponse.getTag() == null && ImapResponseParser.equalsIgnoreCase(imapResponse.get(0), "VANISHED")) {
```
Valodim: can you add examples of the imap response lines that are expected and parsed here as a comment? I think that gives the reader a better impression of what's going on than deducting the format from the lines :)

k9mail-library/src/test/java/com/fsck/k9/mail/store/imap/selectedstate/command/UidFetchCommandTest.java
```diff
@@ -106,8 +107,8 @@ private UidFetchCommand createUidFetchCommand(int maximumAutoDownloadMessageSize
         return new UidFetchCommand.Builder()
                 .maximumAutoDownloadMessageSize(maximumAutoDownloadMessageSize)
                 .idSet(Collections.singletonList(uid))
-                .messageParams(fetchProfile, Collections.singletonMap(String.valueOf(uid),
-                        (Message) createImapMessage(String.valueOf(uid))))
+                .messageParams(fetchProfile, new HashMap<>(Collections.singletonMap(String.valueOf(uid),
```
Valodim: a builder shouldn't take mutable data structures. if this parameter actually needs a mutable hashmap, rather than an immutable Map type, something is wrong

k9mail-library/src/test/java/com/fsck/k9/mail/store/imap/ImapResponseHelper.java
```diff
@@ -43,4 +50,49 @@ public static ImapResponse createImapResponse(String response) throws IOExceptio
         }
         return ids;
     }
+
+    static List<ImapResponse> createFolderOpenedResponse(int openMode, Long uidValidity, Long highestModSeq,
```
Valodim: I see the intention here to make the code shorter, but this is too much logic in a test, which obscures the actual test cases and makes them less isolated.

If we have some imap response in a test that works, the cases where this stops working should be very very rare - why would a reply that is valid in a test no longer be valid at any point? This method is far from trivial, and changing what it does also changes the meaning of all unit tests it is called in, in ways that are very hard to predict.

k9mail-library/src/test/java/com/fsck/k9/mail/store/imap/SelectOrExamineResponseTest.java
```diff
@@ -141,20 +144,15 @@ public void newInstance_withoutOkResponse_shouldReturnNull() throws Exception {
...
-                createImapResponse("* OK [HIGHESTMODSEQ " + TEST_HIGHESTMODSEQ + "] Highest mailbox mod-sequence"),
-                createImapResponse("* FLAGS (\\Answered \\Flagged \\Deleted \\Seen \\Draft NonJunk $MDNSent)"),
-                createImapResponse("* OK [PERMANENTFLAGS (\\Answered \\Flagged \\Deleted \\Seen \\Draft NonJunk " +
-                        "$MDNSent \\*)] Flags permitted."),
-                createImapResponse("* VANISHED (EARLIER) 41,200,230:233"),
-                createImapResponse("* 49 FETCH (UID 117 FLAGS (\\Seen \\Answered) MODSEQ (90060115194045001))"),
-                createImapResponse("* 50 FETCH (UID 119 FLAGS (\\Draft $MDNSent) MODSEQ (90060115194045308))"),
-                createImapResponse(lastLine)
-        );
+    private List<ImapResponse> createMockImapResponses(String lastLine) throws IOException {
```
Valodim: related to createFolderOpenedResponse comment above: if you compare this new code to the old one, it's very hard to tell what is going on exactly. in the old code, it's just a plain response that is simple to parse.

the `createMockImapResponses` method serves a similar purpose to the `createFolderOpenedResponse` one, but is much less complex.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/Capabilities.java
rhari991: Actually I was removing this. Since support for QRESYNC implies support for ENABLE, a check for this capability should not be required.

Valodim: oh, you're right! I was confused because it's back in `Commands` right below :)

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapUtility.java
rhari991: Add a comment on GitHub or a comment in the actual file ?

rhari991: Added

k9mail-library/src/test/java/com/fsck/k9/mail/store/imap/selectedstate/command/UidFetchCommandTest.java
rhari991: Changed

k9mail-library/src/test/java/com/fsck/k9/mail/store/imap/ImapResponseHelper.java
rhari991: Changed to use a data fixture

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -177,54 +174,40 @@ private void handleUidValidity() throws MessagingException {
         long cachedUidValidity = localFolder.getUidValidity();
         long currentUidValidity = imapFolder.getUidValidity();
 
-        if (cachedUidValidity != 0L && cachedUidValidity != currentUidValidity) {
+        if (localFolder.isCachedUidValidityValid() && cachedUidValidity != currentUidValidity) {
```
Valodim: maybe just `hasCachedUidValidity`?

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -243,11 +226,30 @@ void syncRemoteDeletions(List<String> deletedMessageUids) throws IOException, Me
...
         if (moreMessages != null && moreMessages == MoreMessages.UNKNOWN) {
             final Date earliestDate = account.getEarliestPollDate();
-            int remoteStart = getRemoteStart(localFolder, imapFolder);
+            int remoteStart = syncHelper.getRemoteStart(localFolder, imapFolder);
             updateMoreMessages(earliestDate, remoteStart);
         }
     }
 
+    private void getAndOpenLocalFolder(SyncHelper syncHelper) throws MessagingException {
+        if (localFolder == null || !localFolder.isOpen()) {
```
Valodim: under what circumstances does this happen? if an instance of this class requires initialization before it can be used, and there are multiple entry points, maybe just use an `initalize()` method that takes care of the whole thing?

either way a method that starts with `get` should return something, not cause side effects like this.

(also see comment below about the `syncRemoteDeletions` control flow before proceeding on this)

k9mail/src/main/java/com/fsck/k9/controller/NonQresyncExtensionHandler.java
```diff
@@ -47,15 +47,15 @@ int continueSync(MessageDownloader messageDownloader, FlagSyncHelper flagSyncHel
...
+        int remoteStart = syncHelper.getRemoteStart(localFolder, imapFolder);
 
         Timber.v("SYNC: About to fetch UIDs for messages %d through %d in folder %s",
                 remoteStart, remoteMessageCount, folderName);
 
         findRemoteMessagesToDownload(localUidMap, remoteMessages, remoteUidMap, remoteStart);
 
         if (account.syncRemoteDeletions()) {
-            imapSyncInteractor.syncRemoteDeletions(findDeletedMessageUids(localUidMap, remoteUidMap));
+            imapSyncInteractor.syncRemoteDeletions(findDeletedMessageUids(localUidMap, remoteUidMap), syncHelper);
```
Valodim: this construct seems weird to me - the imap sync interactor calls to either `NonQresyncExtensionHandler` or to `QresyncExtensionHandler`, both of which then call back to `imapSyncInteractor` for syncing remote deletions. and inside that method, the imap folder and local folders are also possibly opened anew?

this is really weird control flow. I'm also not sure if the name `syncRemoteDeletions` fits for this method - doesn't it just take a list of message uids and delete them locally? how does it differ from a method that would be called `deleteLocalMessages`?

I understand that the messages that are to be deleted come from different sources in the NonQresync and Qresync control flows, but is it necessary that deletion, but does this deletion have to happen in the `continueSync` method?

k9mail/src/main/java/com/fsck/k9/controller/MessagingController.java
```diff
@@ -726,899 +714,65 @@ public void run() {
...
+        For any message we have locally but not remotely, delete the local message to keep cache clean.
+        Download larger parts of any new messages.
+        (Optional) Download small attachments in the background.
+         */
+
+        Timber.i("Synchronizing folder %s:%s", account.getDescription(), folderName);
 
         for (MessagingListener l : getListeners(listener)) {
-            l.synchronizeMailboxStarted(account, folder);
+            l.synchronizeMailboxStarted(account, folderName);
```
Valodim: I think we should try to keep these pairs of listener callbacks in the same method where possible. in this case, just move them to the beginning of the performSync methods, and also into the "don't ever sync outbox or errors folder" statement

k9mail/src/main/java/com/fsck/k9/controller/MessagingController.java
```diff
@@ -726,899 +714,65 @@ public void run() {
...
-                closeFolder(remoteFolder);
-            }
-
-            closeFolder(tLocalFolder);
+        String storeUri = account.getStoreUri();
+        if (ImapStore.isStoreUriImap(storeUri)) {
+            ImapSyncInteractor syncInteractor = new ImapSyncInteractor(account, folderName, listener, this);
+            syncInteractor.performSync(flagSyncHelper, messageDownloader, syncHelper);
+        } else {
+            LegacySyncInteractor.performSync(account, folderName, listener, this, messageDownloader, syncHelper);
```
Valodim: what about providedRemoteFolder?

rhari991: It's only use case is for IMAP IDLE. I think I removed that prematurely for when IDLE will get it's own class. I'll move it back.

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
rhari991: @Valodim The reason I made a method for that is to make the `syncRemoteDeletions` method viable for testing. If it runs before `performSync` , an NPE occurs since the folder instance variables have not been initialized yet, and after the `performSync` method runs, the folders will have been closed.
Is there a better way to handle this ? How about opening the folders in the constructor ?

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
rhari991: Done

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
rhari991: It's no longer used in two places. Made an `initialize` method as well.

k9mail/src/main/java/com/fsck/k9/controller/NonQresyncExtensionHandler.java
rhari991: Moved both the delete local messages method and update more messages method into `SyncHelper`

k9mail/src/main/java/com/fsck/k9/controller/MessagingController.java
rhari991: Done

k9mail/src/main/java/com/fsck/k9/controller/MessagingController.java
rhari991: Removed the `providedRemoteFolder` parameter

k9mail/src/main/java/com/fsck/k9/controller/NonQresyncExtensionHandler.java
```diff
@@ -55,7 +53,8 @@ int continueSync(MessageDownloader messageDownloader, FlagSyncHelper flagSyncHel
         findRemoteMessagesToDownload(localUidMap, remoteMessages, remoteUidMap, remoteStart);
 
         if (account.syncRemoteDeletions()) {
-            imapSyncInteractor.syncRemoteDeletions(findDeletedMessageUids(localUidMap, remoteUidMap), syncHelper);
+            List<String> deletedUids = findDeletedMessageUids(localUidMap, remoteUidMap);
+            syncHelper.deleteLocalMessages(deletedUids, account, localFolder, imapFolder, controller, listener);
```
Valodim: :+1:

k9mail/src/main/java/com/fsck/k9/controller/MessagingController.java
```diff
@@ -756,9 +752,10 @@ Upload any local messages that are marked as PENDING_UPLOAD (Drafts, Sent, Trash
         String storeUri = account.getStoreUri();
         if (ImapStore.isStoreUriImap(storeUri)) {
             ImapSyncInteractor syncInteractor = new ImapSyncInteractor(account, folderName, listener, this);
-            syncInteractor.performSync(flagSyncHelper, messageDownloader, syncHelper);
+            syncInteractor.performSync(flagSyncHelper, messageDownloader, notificationController, syncHelper);
```
Valodim: how did you pick which of these parameters are passed to the constructor, and which to the method?

stuff like helper, downloader and controller seem more like constructor injected parameters, while folderName and listeners sound more suited as method parameters to me.

rhari991: I believe I passed the classes that I had created by refactoring as method parameters. I'll change this.

k9mail/src/main/java/com/fsck/k9/controller/QresyncExtensionHandler.java
```diff
@@ -0,0 +1,138 @@
...
+        String folderName = imapFolder.getName();
+        Date earliestDate = account.getEarliestPollDate();
+        final AtomicInteger headerProgress = new AtomicInteger(0);
+
+        int newLocalMessageCount = remoteMessagesToDownload.size() + localFolder.getMessageCount();
+        int oldMessagesStart = syncHelper.getRemoteStart(localFolder, imapFolder);
+        int oldMessagesEnd = oldMessagesStart - 1 + localFolder.getVisibleLimit() - newLocalMessageCount;
+
+        if (oldMessagesEnd > oldMessagesStart) {
+            List<ImapMessage> oldMessages = imapFolder.getMessages(oldMessagesStart, oldMessagesEnd, earliestDate, null);
```
rhari991: Looks like I made a mistake here.

The SELECT response for QRESYNC gives the UID + flags for any new messages that arrived since the last sync. However, in this line, we fetch only the UIDs (via UID SEARCH) of old messages that need to be need to be downloaded to fill the mailbox. 
This means that the flags for old messages are never downloaded.

I think the best way to fix this would be to use a FETCH command here instead of a UID SEARCH to fetch the UIDs and flags of old messages using their sequence numbers. This will mean adding support for FETCH (we only support UID FETCH currently).

Is there a better way ? @Valodim

rhari991: The problem with the above approach is that FETCH does not allow us to specify that we only need non-deleted messages. So we might end up with an unfilled mailbox at the end of synchronization.

Valodim: yup, `FETCH` seems like the way to go here

k9mail/src/main/java/com/fsck/k9/mailstore/LocalStore.java
```diff
@@ -1189,11 +1191,11 @@ public void doDbWork(SQLiteDatabase db, String selectionSet, String[] selectionA
                     throws UnavailableStorageException {
 
                 db.execSQL("UPDATE messages SET " + flagColumn + " = " + ((newState) ? "1" : "0") +
-                        " WHERE id IN (" +
-                        "SELECT m.id FROM threads t " +
-                        "LEFT JOIN messages m ON (t.message_id = m.id) " +
-                        "WHERE m.empty = 0 AND m.deleted = 0 " +
-                        "AND t.root" + selectionSet + ")",
+                                " WHERE id IN (" +
```
rhari991: @Valodim should I fix these spaces ? They seem to be auto-generated

k9mail/src/main/java/com/fsck/k9/mailstore/LocalStore.java
```diff
@@ -594,9 +596,9 @@ public boolean isCopyCapable() {
...
     List<LocalMessage> getMessages(
-        final MessageRetrievalListener<LocalMessage> listener,
-        final LocalFolder folder,
-        final String queryString, final String[] placeHolders
+            final MessageRetrievalListener<LocalMessage> listener,
+            final LocalFolder folder,
+            final String queryString, final String[] placeHolders
     ) throws MessagingException {
         final List<LocalMessage> messages = new ArrayList<>();
         final int j = database.execute(false, new DbCallback<Integer>() {
```
rhari991: @Valodim @cketti Any idea why we first query 10 rows and then the remaining ones in this method ?

rhari991:
There are 7 tests in `MessagingControllerTest` (which test the `sendPendingMessagesSynchronous` method) and 8 tests in `MigrationTest` that are failing in the Travis build. These are failing for me when I run them on the master branch too. Does this happen with you guys too or am I missing something ? @Valodim @philipwhiuk @cketti 

When running `MigrationTest` , I get this exception for all the failed tests 

`android.content.res.Resources$NotFoundException: Unable to find resource ID #0x7f07018e in packages [android, org.robolectric.default]

	at org.robolectric.shadows.ShadowAssetManager.getResName(ShadowAssetManager.java:868)
	at org.robolectric.shadows.ShadowAssetManager.resolveResourceValue(ShadowAssetManager.java:676)
	at org.robolectric.shadows.ShadowAssetManager.resolve(ShadowAssetManager.java:636)
	at org.robolectric.shadows.ShadowAssetManager.getAndResolve(ShadowAssetManager.java:630)
	at org.robolectric.shadows.ShadowAssetManager.getResourceText(ShadowAssetManager.java:226)
	at android.content.res.AssetManager.getResourceText(AssetManager.java)
	at android.content.res.Resources.getText(Resources.java:295)
	at android.content.res.Resources.getString(Resources.java:385)
	at android.content.Context.getString(Context.java:377)
	at com.fsck.k9.Account.<init>(Account.java:331)
	at com.fsck.k9.Preferences.newAccount(Preferences.java:113)
	at com.fsck.k9.mailstore.MigrationTest.getNewAccount(MigrationTest.java:725)
	at com.fsck.k9.mailstore.MigrationTest.setUp(MigrationTest.java:49)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:24)
	at org.robolectric.RobolectricTestRunner$2.evaluate(RobolectricTestRunner.java:339)
	at org.robolectric.RobolectricTestRunner.runChild(RobolectricTestRunner.java:259)
	at org.robolectric.RobolectricTestRunner.runChild(RobolectricTestRunner.java:41)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.robolectric.RobolectricTestRunner$1.evaluate(RobolectricTestRunner.java:199)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:117)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:42)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:262)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:84)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)`

The related line, `Account.java:331` seems to take a String resource which has not been mocked yet ?

For the failing tests in `MessagingControllerTest`, the assertions are failing.

philipwhiuk:
I believe this occurs because IntelliJ (and thus Android Studio)'s default Run Configuration is based on the project working directory not the module working directory ( `/` instead of `/k9mail` ).

When I run the tests for `master` via `./gradlew test` they pass.

When I run your branches's tests in the same way they fail.

rhari991:
@philipwhiuk thanks for the tip, setting the JUnit working directory did the trick

k9mail-library/src/main/java/com/fsck/k9/mail/K9MailLib.java
```diff
@@ -3,6 +3,7 @@
```
philipwhiuk: I'm not a big fan of making this a super object in terms of configuration. `ImapConfig` would be a good name for a class that could house the new stuff being added here.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
```diff
@@ -679,14 +689,19 @@ private boolean isListResponse(ImapResponse response) {
...
+        requestCapabilitiesIfNecessary();
         return capabilities.contains(capability.toUpperCase(Locale.US));
     }
 
-    public boolean isCondstoreCapable() throws IOException, MessagingException  {
+    public boolean isCondstoreCapable() throws IOException, MessagingException {
         return hasCapability(Capabilities.CONDSTORE);
     }
 
+    boolean isQresyncCapable() throws IOException, MessagingException {
```
philipwhiuk: Why is `isCondstoreCapable` public and this not?

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
```diff
@@ -144,27 +157,39 @@ public void open(int mode) throws MessagingException {
...
              * are notified otherwise in the responses.
              */
             this.mode = mode;
-
-            for (ImapResponse response : responses) {
-                handlePermanentFlags(response);
+            if (response.hasOpenMode()) {
+                this.mode = response.getOpenMode();
+            }
+            if (response == null) {
```
philipwhiuk: This null check is pointless. Line 182 would throw an NPE before it got here.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
```diff
@@ -652,72 +709,79 @@ public void fetch(List<ImapMessage> messages, FetchProfile fetchProfile,
...
+                ImapList fetchList = (ImapList) response.getKeyedValue("FETCH");
+                String uid = fetchList.getKeyedString("UID");
+                long msgSeq = response.getLong(0);
+                if (uid != null) {
+                    try {
+                        msgSeqUidMap.put(msgSeq, uid);
+                        if (K9MailLib.isDebug()) {
+                            Timber.v("Stored uid '%s' for msgSeq %d into map", uid, msgSeq);
                         }
+                    } catch (Exception e) {
```
philipwhiuk: Exception is a very broad catch case. Try to narrow this down a bit.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
```diff
@@ -1202,6 +1257,28 @@ public void expunge() throws MessagingException {
...
+        try {
+            List<ImapResponse> expungeResponses = executeSimpleCommand("EXPUNGE");
+            handleExpungeResponses(expungeResponses);
+            return ImapUtility.extractVanishedUids(expungeResponses);
+        } catch (IOException ioe) {
+            throw ioExceptionHandler(connection, ioe);
+        }
+    }
+
+    private void handleExpungeResponses(List<ImapResponse> imapResponses) {
```
philipwhiuk: This doesn't really 'handle' the response - it just searches them for a higher mod seq.

`updateHighestModSequenceFromResponses`.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
```diff
@@ -1291,7 +1368,7 @@ public int hashCode() {
         return getName().hashCode();
     }
 
-    private ImapStore getStore() {
+    ImapStore getStore() {
```
philipwhiuk: Why is this necessary?

k9mail/src/main/java/com/fsck/k9/controller/NonQresyncExtensionHandler.java
```diff
@@ -0,0 +1,170 @@
...
+            for (Message message : messages) {
+                if (message instanceof ImapMessage) {
+                    imapMessages.add((ImapMessage) message);
+                }
+            }
+            imapFolder.fetchChangedMessageFlagsUsingCondstore(imapMessages, cachedHighestModSeq,
+                    new MessageRetrievalListener<ImapMessage>() {
+                        @Override
+                        public void messageStarted(String uid, int number, int ofTotal) {
+
```
philipwhiuk: Nothing?

k9mail/src/main/java/com/fsck/k9/controller/SyncHelper.java
```diff
@@ -0,0 +1,268 @@
...
+     * can't create we'll abort. This will happen on every single Pop3 folder as
+     * designed and on Imap folders during error conditions. This allows us
+     * to treat Pop3 and Imap the same in this code.
+     */
+    boolean verifyOrCreateRemoteSpecialFolder(Account account, String folderName, Folder remoteFolder,
+            MessagingListener listener, MessagingController controller) throws MessagingException {
+        if (folderName.equals(account.getTrashFolderName()) ||
+                folderName.equals(account.getSentFolderName()) ||
+                folderName.equals(account.getDraftsFolderName())) {
+            if (!remoteFolder.exists()) {
```
philipwhiuk: Can you not combine these `if's` to reduce the nesting a bit?

k9mail/src/main/java/com/fsck/k9/mailstore/LocalFolder.java
```diff
@@ -74,6 +75,8 @@
     private static final long serialVersionUID = -1973296520918624767L;
     private static final int MAX_BODY_SIZE_FOR_DATABASE = 16 * 1024;
     static final long INVALID_MESSAGE_PART_ID = -1;
+    private static final int INVALID_UID_VALIDITY = -1;
```
philipwhiuk: Redefining the same constants as in the library. Maybe they should be visible if you need them here?

k9mail/src/main/java/com/fsck/k9/mailstore/LocalFolder.java
```diff
@@ -183,10 +188,23 @@ public Void doDbWork(final SQLiteDatabase db) throws WrappedException {
...
     }
 
-    void open(Cursor cursor) throws MessagingException {
+    public void open(Cursor cursor) throws MessagingException {
         mFolderId = cursor.getInt(LocalStore.FOLDER_ID_INDEX);
         mName = cursor.getString(LocalStore.FOLDER_NAME_INDEX);
         mVisibleLimit = cursor.getInt(LocalStore.FOLDER_VISIBLE_LIMIT_INDEX);
+
+        //The index check here is because of an issue that occurs while upgrading the db to V55 - at that point, the
+        //uid_validity and highest_mod_Seq columns do not exist yet
```
philipwhiuk: This is why we have DatabaseMigration...

k9mail/src/main/java/com/fsck/k9/mailstore/LocalMessage.java
```diff
@@ -135,6 +106,38 @@ void populateFromGetMessageCursor(Cursor cursor) throws MessagingException {
...
+            String[] flags = flagList.split(",");
+
+            for (String flag : flags) {
+                try {
+                    setFlagInternal(Flag.valueOf(flag), true);
+                }
+
+                catch (Exception e) {
+                    if (!"X_BAD_FLAG".equals(flag)) {
+                        Timber.w("Unable to parse flag %s", flag);
```
philipwhiuk: Why not check for X_BAD_FLAG first?

k9mail-library/src/main/java/com/fsck/k9/mail/K9MailLib.java
rhari991: Done

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapConnection.java
rhari991: `isCondstoreCapable` is used in the `FolderSelectedStateCommand.getCommandLengthLimit()` method, which is in the `selectedstate` subpackage.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
rhari991: fixed

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
rhari991: Funny, the `ConcurrentHashMap.put()` method does not throw any checked exceptions, so this try/catch block is useless. I'm not sure why this was there in the first place. I moved around a lot of original code without modifying it.

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
rhari991: Done

k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
rhari991: It's not strictly necessary here, my bad

k9mail/src/main/java/com/fsck/k9/controller/NonQresyncExtensionHandler.java
rhari991: I don't think anything is required here ?

k9mail/src/main/java/com/fsck/k9/controller/SyncHelper.java
rhari991: Done

k9mail/src/main/java/com/fsck/k9/mailstore/LocalFolder.java
rhari991: Done

k9mail/src/main/java/com/fsck/k9/mailstore/LocalFolder.java
rhari991: The problem is that the `open(Cursor)` method is called when upgrading to V55, but the `uid_validity` and `highest_mod_seq` columns do not exist yet. They are added only in V61. There does not seem to be a reliable way to recreate the open(Cursor) method in `MigrationTo61` .

k9mail/src/main/java/com/fsck/k9/mailstore/LocalMessage.java
rhari991: Done

rhari991:
@philipwhiuk made the changes

rhari991:
@Valodim @philipwhiuk I'm not sure why/how the `CertificationException` is showing up, could you have a look ?

Valodim:
possibly related to #2675?

PR #2611, opened 2017-06-30
Rewrite some UI of account setup (WIP)

daquexian:
Rewrite some UI of account setup.

And convert four activities, `AccountSetupBasics`, `AccountSetupCheckSettings`, `AccountSetupIncoming`, `AccountSetupOutgoing` to MVP pattern. Other activities about account setup are not modified for the time being.

Model       -- `Account`
View         -- `Activity`
Presenter -- `Presenter` class

I put most logic into `Presenter`. However there are some codes that I have no idea how to deal with. Please give me some advice :)

I also changed the work flow a bit. `AccountSetupBasics` now is only an activity that let users enter email and password. Advanced options, for example client certificate, are only available in `AccountSetupIncoming` and `AccountSetupOutgoing`. `AccountSetupCheckSettings` contains automatic account configuration and incoming/outgoing settings checking.

NOTE: this branch is only about UI changes. So functional code changes(e.g. auto configuration) are not included here.


k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/AccountSetupOutgoing.java
```diff
@@ -437,47 +375,35 @@ private void updateAuthPlainTextFromSecurityType(ConnectionSecurity securityType
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
         if (resultCode == RESULT_OK) {
```
daquexian: One of my questions is, does the logic here need to be put in presenter, too? If so, several methods scattered in presenter and view will be produced, and each methods only contains up to 1~5 lines. It will make the code hard to understand.

k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/AccountSetupOutgoing.java
```diff
@@ -437,47 +375,35 @@ private void updateAuthPlainTextFromSecurityType(ConnectionSecurity securityType
...
         String newHost = mServerView.getText().toString();
         int newPort = Integer.parseInt(mPortView.getText().toString());
-        ServerSettings server = new ServerSettings(Type.SMTP, newHost, newPort, securityType, authType, username, password, clientCertificateAlias);
-        uri = Transport.createTransportUri(server);
-        mAccount.deleteCertificate(newHost, newPort, CheckDirection.OUTGOING);
-        mAccount.setTransportUri(uri);
-        AccountSetupCheckSettings.actionCheckSettings(this, mAccount, CheckDirection.OUTGOING);
+
+        boolean requireLogin = mRequireLoginView.isChecked();
+        presenter.next(username, password, clientCertificateAlias, newHost, newPort, securityType,
```
daquexian: Is it ok to declare a method with so many arguments? Is there any way better?

k9mail/src/main/java/com/fsck/k9/helper/Utility.java
```diff
@@ -111,21 +111,22 @@ public static boolean requiredFieldValid(TextView view) {
 
 
     public static boolean requiredFieldValid(Editable s) {
+        return s != null && requiredFieldValid(s.toString());
```
daquexian: To test presenter classes, is it needed to make all involved static methods in `Utility` non-static?

k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/AccountSetupOutgoing.java
```diff
@@ -82,18 +89,10 @@ public void onCreate(Bundle savedInstanceState) {
...
-        try {
-            if (new URI(mAccount.getStoreUri()).getScheme().startsWith("webdav")) {
-                mAccount.setTransportUri(mAccount.getStoreUri());
-                AccountSetupCheckSettings.actionCheckSettings(this, mAccount, CheckDirection.OUTGOING);
-            }
-        } catch (URISyntaxException e) {
-            // TODO Auto-generated catch block
-            e.printStackTrace();
-        }
+        Account account = Preferences.getPreferences(this).getAccount(accountUuid);
```
Valodim: just pass the account uuid here. it's incoming data from the intent, but the view doesn't have to know what to do with it

k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/AccountSetupOutgoing.java
```diff
@@ -115,27 +114,21 @@ public void onCreate(Bundle savedInstanceState) {
...
         if (savedInstanceState != null && savedInstanceState.containsKey(EXTRA_ACCOUNT)) {
             accountUuid = savedInstanceState.getString(EXTRA_ACCOUNT);
-            mAccount = Preferences.getPreferences(this).getAccount(accountUuid);
+            account = Preferences.getPreferences(this).getAccount(accountUuid);
+            presenter.setAccount(account);
         }
 
         try {
-            ServerSettings settings = Transport.decodeTransportUri(mAccount.getTransportUri());
+            ServerSettings settings = Transport.decodeTransportUri(account.getTransportUri());
```
Valodim: all of the view updating should happen in the presenter

k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/AccountSetupOutgoing.java
```diff
@@ -354,68 +347,13 @@ private void updateViewFromAuthType() {
...
-        boolean hasValidPasswordSettings = hasValidUserName
-                && !isAuthTypeExternal
-                && Utility.requiredFieldValid(mPasswordView);
 
-        boolean hasValidExternalAuthSettings = hasValidUserName
-                && isAuthTypeExternal
-                && hasConnectionSecurity
-                && hasValidCertificateAlias;
+        presenter.revokeInvalidSettings(getSelectedAuthType(), getSelectedSecurity());
+        presenter.validateFields(mClientCertificateSpinner.getAlias(), mServerView.getText().toString(),
```
Valodim: the view should only tell the presenter "what is happening", but not "what to do".

the event in this case is that input changed, so this should fire something like `onInputChanged`. the presenter then decides what to do with that, i.e. verify the input, and updates the view accordingly.

k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/AccountSetupOutgoing.java
Valodim: That depends. Usually, the flow will be something like `MvpView.showSomeDialog()` which will lead to a callback to `Presenter.onSomeDialogDismissed()` or similar. The view definitely shouldn't decide what to do (like the `finish()` here)

Sometimes it's easier to just hand through the `onActivityResult`, but that leads to a stronger coupling between the android view code and presenter, and should not be the typical case.

k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/AccountSetupOutgoing.java
Valodim: Could make a model class for that, but I'm not sure that will help too much. I'd leave it as it is for now.

k9mail/src/main/java/com/fsck/k9/helper/Utility.java
Valodim: If there is no need to substitute those methods for testing purposes, then no. Otherwise, classes should ideally be transformed to non-static in one coherent change in a single commit

k9mail/src/main/java/com/fsck/k9/helper/EmailHelper.java
```diff
@@ -2,8 +2,6 @@
 
 
 public final class EmailHelper {
-    private EmailHelper() {}
```
Valodim: the purpose of this is to make the class non-instantiable

k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/OutgoingPresenter.java
```diff
@@ -0,0 +1,118 @@
...
+                && isAuthTypeExternal
+                && hasConnectionSecurity
+                && hasValidCertificateAlias;
+
+        boolean enabled = Utility.domainFieldValid(server)
+                        && Utility.requiredFieldValid(port)
+                        && (!requireLogin
+                                || hasValidPasswordSettings || hasValidExternalAuthSettings);
+
+        view.enableNext(enabled);
```
Valodim: `setNextButtonEnabled`

k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/OutgoingPresenter.java
```diff
@@ -0,0 +1,118 @@
...
+            clientCertificateAlias = null;
+        }
+
+        ServerSettings server = new ServerSettings(Type.SMTP, host, port, connectionSecurity,
+                authType, username, password, clientCertificateAlias);
+        String uri = Transport.createTransportUri(server);
+        account.deleteCertificate(host, port, CheckDirection.OUTGOING);
+        account.setTransportUri(uri);
+
+        view.checkAccount(account);
```
Valodim: "checking an account" is not something a view should do

k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/AccountSetupOutgoing.java
Valodim: this is the whole point of MVP, to remove as much logic and responsibility from the view as possible. to put it differently, there should be as few `if`s in an mvp view as possible, and those should only affect simple things like hiding or showing a view, not make actual decisions.

daquexian:
@Valodim I rewrote outgoing classes. I think they are more "MVP" now :) Could you please review it so that I can do better when rewriting other classes?

k9mail/src/main/java/com/fsck/k9/activity/setup/outgoing/OutgoingPresenter.java
```diff
@@ -57,37 +55,40 @@ public void setAccount(String accountUuid) {
                 return;
             }
         } catch (URISyntaxException e) {
-            Timber.e(e, "Account retrieved from intent doesn't have valid store URI");
-            return;
+            view.onAccountLoadFailure(e);
```
daquexian: The way to handle this error is from original codes. But is it ok to just show a toast and do nothing else like finishing the whole activity?

k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSetup.java
```diff
@@ -120,6 +174,52 @@ private void setSelection(int position) {
...
+                namesView.start();
+                break;
+        }
     }
 
+    @Override
+    public void doPositiveClick(int dialogId) {
+        switch (dialogId) {
+            case R.id.dialog_account_setup_error: {
+                // presenter.onEditDetailClickedWhenError();
```
daquexian: Here is one of the places where I don't know what is the best way. The listener must be an activity. In order to call presenter's methods, should I declare a method in view and then call methods in presenter indirectly?

k9mail/src/main/java/com/fsck/k9/notification/AuthenticationErrorNotifications.java
```diff
@@ -59,8 +59,8 @@ public void clearAuthenticationErrorNotification(Account account, boolean incomi
 
     PendingIntent createContentIntent(Context context, Account account, boolean incoming) {
         Intent editServerSettingsIntent = incoming ?
-                AccountSetupIncoming.intentActionEditIncomingSettings(context, account) :
-                AccountSetupOutgoing.intentActionEditOutgoingSettings(context, account);
+                IncomingView.intentActionEditIncomingSettings(context, account) :
```
daquexian: When we setup a new account, the activity will finish when the whole process ends. But sometimes the activity is started for only one view in it. My opinion is using some flags to control it (sounds silly and not elegant). Could you please tell me whether there is some better way?

k9mail/src/main/java/com/fsck/k9/activity/setup/checksettings/CheckSettingsPresenter.java
```diff
@@ -138,12 +145,25 @@ private void checkOutgoing() {
...
-                K9.setServicesEnabled(K9.app);
+                if (!editSettings) {
+                    //We've successfully checked outgoing as well.
+                    account.setDescription(account.getEmail());
+                    K9.setServicesEnabled(K9.app);
 
-                view.goToNames();
+                    view.goToNames();
+                } else {
+                    boolean isPushCapable = false;
```
daquexian: `CheckSettingsPresenter` now has some logic unrelated to checking settings. But it seems that there is no better place than it.

k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSetup.java
daquexian: I solved it in b5d44d2

daquexian:
The UI is updated and I removed "options" dialog because the options can be set in account settings and I personally find this dialog useless, too.

I'm afraid the UI now is not "polished", because I am not professional designer and the existing mockup is not so proper and detailed. It will be helpful if a designer can discuss with us.

k9mail/src/main/res/layout/account_setup_incoming.xml
```diff
@@ -243,11 +284,12 @@
                     android:contentDescription="@string/account_setup_incoming_compression_label"
                     />
             </LinearLayout>
-            <View
+            <!--<View
```
daquexian: Almost all these xmls has this view. What is it for? It seemed that nothing happened after I removed them.

PR #2692, opened 2017-08-22
move responsibility for destroying messages from LocalMessage to LocalFolder

Valodim:
Quick refactoring PR: This moves the responsibility of `destroy`ing messages from being shared between `LocalMessage` and `LocalFolder` entirely into `LocalFolder`. The parts in `LocalMessage` included stuff like threading, which really isn't something a message itself should worry about.

This also allows a `LocalFolder` to destroy messages given their id, messageId and messagePartId, which should come in handy if we want to delete messages efficiently without loading their content first.


Valodim:
renamed `mId` to `messageId` and the previous `messageId` to `messageIdHeader`. should be clear now.

philipwhiuk:
I have to say, I'm not big on LocalFolder being responsible for every action you can take on a message. I think we really want 'LocalActions' as a some kind of Controller/Visitor/Strategy, leaving LocalMessage and LocalFolder as models.

But this better than what have now.

PR #2695, opened 2017-08-22
Cleanup smtp transport

Valodim:
cleanup of SmtpTransport

I also moved the uri serialization stuff of ServerSettings into a TransportUris class. The transports themselves really shouldn't care about how their settings objects are serialized in a different part of the codebase. The same goes for `*Store` classes, but that's more and not directly related work.

Related note, check out this gem of our codebase: 
https://github.com/k9mail/k-9/blob/master/k9mail-library/src/main/java/com/fsck/k9/mail/transport/smtp/SmtpTransport.java#L460


philipwhiuk:
In my defence it's slightly better than it was:

https://github.com/k9mail/k-9/commit/8f0b7b5059cad905c1c580c17b4e61549b523ce5

Valodim:
all good ;)

also, 'the hell these tests are failing? they work for me locally

Valodim:
see I just included that so you could find it and feel better ;P

note to self: k9mail:testDebug does not run tests in k9mail-lib

PR #2701, opened 2017-08-24
Rewrite expunge

rhari991:
This PR rewrites the `expunge()` method in `ImapFolder` to ensure that only messages that were marked with the `\Deleted` flag by K-9 Mail are expunged. This process is described in [section 4.2.4 of RFC 4549](https://tools.ietf.org/html/rfc4549#section-4.2.4)


k9mail-library/src/main/java/com/fsck/k9/mail/store/imap/ImapFolder.java
```diff
@@ -1191,12 +1193,32 @@ public String getUidFromMessageId(Message message) throws MessagingException {
...
         try {
-            executeSimpleCommand("EXPUNGE");
+            if (connection.isUidPlusCapable()) {
+                UidExpungeCommand uidExpungeCommand = new UidExpungeCommand.Builder()
+                        .idSet(knownDeletedUids)
+                        .build();
+
+                uidExpungeCommand.execute(connection, this);
+            } else {
+                UidSearchCommand searchCommand = new UidSearchCommand.Builder()
```
Valodim: I think this warrants its own method. also, a reference to the rfc would be nice, since this type of fallback method from a spec is not very common

rhari991: Done

Valodim: ...I don't think so, actually?

rhari991: I have moved the UidSearchCommand and it's execution into a new method `getUnknownDeletedUids()` and have added a comment regarding the RFC at the top of the try block

Valodim: oh, I see. I meant the whole thing in the else block, maybe something like `expungeWithoutUidPlus`?

rhari991: Ah ok, done

rhari991:
Rebased in #2927

PR #2722, opened 2017-08-31
Handling RFC2047 non-compliant splitting of characters

philipwhiuk:
Implementing handling of RFC2047 non-compliant behaviour - from #2719 feedback

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/internet/DecoderUtil.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/internet/DecoderUtil.java
@@ -158,18 +158,24 @@ class DecoderUtil {
                 if (previousWord != null) {
                     sb.append(decodeEncodedWord(previousWord));
                     sb.append(sep);
+                    sb.append(body.substring(begin, end));
                     previousWord = null;
-                }
-            } else if (previousWord != null) {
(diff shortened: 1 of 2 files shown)


philipwhiuk:
Fixed the broken test and added an extra unit test to catch the behaviour that broke in the other test. :)

The way this works is we keep the previous encoded word and don't decode it until either we reach some decoded text or we reach some differently encoded text (so we won't combine Q & B encodings or UTF-8 and GB2312 encoded data). So consecutive identically encoded data is merged together and then decoded as one piece. This fixes the non-compliant data that splits over characters while handling all RFC compliant behaviour correctly.

Valodim:
Handling of whitespace between encoded-words is quite involved, I would suggest adding some tests based on the examples in section 7 of rfc2047:

> =?iso-8859-1?q?this is some text?=

=?iso-8859-1?q?this is some text?=

> =?iso-8859-1?q?this=20is=20some=20text?=

this is some text

> =?ISO-8859-1?Q?a?= b

a b

> =?ISO-8859-1?Q?a?=                         
       =?ISO-8859-1?Q?b?=

ab

> =?ISO-8859-1?Q?a_b?=

a b

Valodim:
on a related note, would you mind pushing your branches to the k9mail org's repository? It's more convenient to add commits to other people's PRs that way, handling remotes is kind of a pita.

PR #2724, opened 2017-09-01
Pipeline support

Valodim:
This picks up @YESALAM's smtp pipelining. I made a new PR because I couldn't force-push to his branch directly. See the original PR #2388 for details.


philipwhiuk:
Thanks Valodim, my git-foo was failing me. Your extra commit looks good too.

PR #2725, opened 2017-09-02
Handling RFC2047 non-compliant splitting of characters 

philipwhiuk:
From #2722


philipwhiuk:
@cketti Are you happy with this improvement to #2719 - it resolves the points you made there.

PR #2741, opened 2017-09-07
iCal (ICS) support

philipwhiuk:
So I've improved this PR, rebased it etc etc. I added support for insecure iCal parts (when I originally wrote the PR I didn't understand `extraParts`. I made some slight refactoring to make it clearer what extra parts were). I don't want to show every possible thing in the iCal spec - that's the point of a calendar app. I've just made it usable for the stuff you get in a typical event.

This PR still only supports **reading** the common iCal events. It doesn't support responding to them inside K-9. Outlook has a nice 'Accept / Tentative / Decline' thing which I'm tempted to implement. But that can be a separate PR - this stands up fine on it's own.

You can still download and open the iCal file corresponding to an event in Google if you want to.


cketti:
I haven't yet looked at this very closely. But I saw a couple of FIXME comments and that's why I added the "don't merge" label.

@philipwhiuk: Can you please add screenshots to show off the feature.

Small note: There's no need to add a license file for a library whose source code doesn't live inside our repository.

philipwhiuk:
The FIXME comments are mostly marking code that was already commented out. But I'll take a look and see why they were commented out before and fix it.

philipwhiuk:
screenshot_20180119-000353

Valodim:
rebased on current master. I dropped the locked ical events, this is fallback anyways so showing those as locked attachments is fine.

I dislike how much code this adds, but it *is* a cool feature, especially if we have 1) automatic input into calendar, 2) quick accept/reject replies.

KlavsKlavsen:
this would be really really nice to have

cketti:
Love the idea. But I feel a feature like this needs more UI/UX design work upfront.

PR #2777, opened 2017-09-17
Some updates to MessageViewInfoExtractor

### LINKED ISSUE #2776: S/MIME signed emails in sent folder misleadingly shown as PGP encrypted mails
### Expected behavior

I am using S/MIME to sign e-mails on my computer. When I open sent mails in k9 mail app on my smartphone the mail body is shown correctly, I can read the sent message, because it is just signed. I also see the attachment (signature file" smime.p7s") there.

### Actual behavior

Since last update the S/MIME signed mails are shown correctly in AutoPreview (I see text there), but after clicking on them to open the mail and read the full message. Instead a message "this mail is encrypted / This e-mail is with OpenPGP encrypted. To read it you need compatible OpenPGP-App installed and configured / Choose OpenPGPG app" appeaers and I have no chance to read the message.

### Steps to reproduce
1. use a s/mime to sign e-mails from your computer
2. open the sent mail on your smartphone with k9 mail app
3. then this message appears

### Environment
K-9 Mail version: 5.300 (installed 11th Sept. 2017)

Android version: 7.0

Valodim:
I changed MessageViewInfoExtractor a little, making it more explicit when crypto is going on and when it isn't. I also got rid of `MessageCryptoSplitter` in the process, which was super tightly coupled to the extractor behavior in the first place.

The conditions under which a message is considered pgp-encrypted but no provider is installed are now also the way they really should be, which fixes #2776 

Oh and I nailed that behavior down with a bunch of tests. I ran the tests before and after, and only the tests that involve the "openpgp no provider" error state fail, all others attest to unchanged behavior.


k9mail/src/main/java/com/fsck/k9/mailstore/MessageViewInfoExtractor.java
```diff
@@ -71,43 +73,58 @@ public static MessageViewInfoExtractor getInstance() {
...
-        if (cryptoMessageParts != null) {
-            rootPart = cryptoMessageParts.contentPart;
-            cryptoResultAnnotation = cryptoMessageParts.contentCryptoAnnotation;
-            extraParts = cryptoMessageParts.extraParts;
-        } else {
+        ArrayList<Part> extraParts = new ArrayList<>();
+        Part cryptoContentPart = MessageCryptoStructureDetector.findPrimaryEncryptedOrSignedPart(message, extraParts);
+
+        if (cryptoContentPart == null) {
             if (annotations != null && !annotations.isEmpty()) {
```
philipwhiuk: Maybe rename this parameter 'cryptoAnnotations'.

k9mail/src/main/java/com/fsck/k9/mailstore/MessageViewInfoExtractor.java
```diff
@@ -71,43 +73,58 @@ public static MessageViewInfoExtractor getInstance() {
...
+            throws MessagingException {
+        if (cryptoContentPartAnnotation != null && cryptoContentPartAnnotation.hasReplacementData()) {
+            cryptoContentPart = cryptoContentPartAnnotation.getReplacementData();
         }
 
-        AttachmentResolver attachmentResolver = AttachmentResolver.createFromPart(rootPart);
+        List<AttachmentViewInfo> extraAttachmentInfos = new ArrayList<>();
+        ViewableExtractedText extraViewable = extractViewableAndAttachments(extraParts, extraAttachmentInfos);
+
+        MessageViewInfo messageViewInfo = extractSimpleMessageForView(message, cryptoContentPart);
```
philipwhiuk: cryptoContentPart is 'decryptoedContentPart' ?

Valodim: well, sometimes. the content part that is actually displayed may be replaced with something else from the crypto annotation (four lines above)

philipwhiuk: Hence decyptoed rather than decrypted. Too subtle I guess. Maybe `viewableVersionOfContentPart`

As in:

```
Part viewableVersionOfContentPart;
if (...)
   viewableVersionOfContentPart = cryptoContentPartAnnotation.getReplacementData();
else
  viewableVersionOfContentPart = cryptoContentPart;
```

Valodim: Eh. It's a 9 line method, I don't think we have to be too picky here

Valodim:
Went through all my test mails again with this branch. I'm good with this now.

PR #2816, opened 2017-10-04
query recipients also from crypto provider, if available

Valodim:
What it says on the tin. When recipients are loaded from a query, and a crypto provider is configured, the contacts are also filled from the crypto provider.

Luckily I had added the necessary columns to the autocrypt query interface already, making this a very simple operation. I had just kinda forgotten about it.


Valodim:
threw in a couple unit tests for good measure

PR #2899, opened 2017-11-02
Show the name of each account imported

jyn514:
For each account, show `Imported <old account> as <new account>`.

Does not currently support localization.

Replaces `// TODO: show each account imported (and possibly old name)` with `// TODO: localization`.

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/activity/Accounts.java
+++ b/k9mail/src/main/java/com/fsck/k9/activity/Accounts.java
@@ -1565,12 +1565,11 @@ public class Accounts extends K9ListActivity implements OnItemClickListener {
 
         @Override
         protected String generateMessage(Accounts activity) {
-            // TODO: localization
 
             StringBuilder result = new StringBuilder();
             for (AccountDescriptionPair account : mImportResults.importedAccounts) {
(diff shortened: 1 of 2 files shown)


philipwhiuk:
Have you tested the UX of this with large numbers of accounts?

k9mail/src/main/java/com/fsck/k9/activity/Accounts.java
```diff
@@ -1565,12 +1565,18 @@ private void showAccountsImportedDialog(ImportResults importResults, String file
 
         @Override
         protected String generateMessage(Accounts activity) {
-            //TODO: display names of imported accounts (name from file *and* possibly new name)
+            // TODO: localization
 
+            StringBuilder result = new StringBuilder();
+            for (AccountDescriptionPair account : mImportResults.importedAccounts) {
+                result.append("Imported ").append(account.original.name)
```
philipwhiuk: It shouldn't be that hard to localise this.

jyn514:
I have not tested this with large numbers of accounts; I don't have many of my own to test with. I may get around to adding more sample accounts to androidTest, which would help to test other areas as well.

PR #2911, opened 2017-11-04
Simplified logic

jyn514:
Changed some if-clauses to `return <condition>`
Removed some temporary variables in favor of using directly
Replaced some ternaries with boolean expressions
Replaced loop meant to get first element only with more readable code

### DIFF
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/internet/MessageExtractor.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/internet/MessageExtractor.java
@@ -447,24 +447,14 @@ public class MessageExtractor {
             dispositionFilename = MimeUtility.getHeaderParameter(disposition, "filename");
         }
 
-        boolean isAttachmentDisposition = "attachment".equalsIgnoreCase(dispositionType) || dispositionFilename != null;
-        if (isAttachmentDisposition) {
+        if ("attachment".equalsIgnoreCase(dispositionType) || dispositionFilename != null) {
             return false;
         }
 
-        if (part.isMimeType("text/html")) {
-            return true;
-        }
+        return part.isMimeType("text/html")
+                || part.isMimeType("text/plain")
+                || part.isMimeType("application/pgp");
 
-        if (part.isMimeType("text/plain")) {
-            return true;
-        }
-
-        if (part.isMimeType("application/pgp")) {
-            return true;
-        }
-
-        return false;
     }
 
     private static String getContentDisposition(Part part) {
--- a/k9mail-library/src/main/java/com/fsck/k9/mail/internet/MessageExtractor.java
+++ b/k9mail-library/src/main/java/com/fsck/k9/mail/internet/MessageExtractor.java
@@ -447,13 +447,12 @@ public class MessageExtractor {
             dispositionFilename = MimeUtility.getHeaderParameter(disposition, "filename");
         }
 
+        // if part is an attachment disposition
         if ("attachment".equalsIgnoreCase(dispositionType) || dispositionFilename != null) {
             return false;
         }
 
-        return part.isMimeType("text/html")
-                || part.isMimeType("text/plain")
-                || part.isMimeType("application/pgp");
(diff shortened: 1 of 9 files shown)


k9mail-library/src/main/java/com/fsck/k9/mail/store/webdav/DataSet.java
```diff
@@ -52,10 +52,11 @@ public void finish() {
      */
     public Map<String, String> getSpecialFolderToUrl() {
         // We return the first (and only) map
-        for (Map<String, String> folderMap : mData.values()) {
-            return folderMap;
+        if (mData.size() == 1) {
```
cketti: This changes the behavior if `mData` contains more than one element. It shouldn't according to the comment, but from glancing at the code it's not immediately obvious that comment is always true.

I'd change the code to the following:

```java
return mData.isEmpty() ? Collections.<String, String>emptyMap() : mData.values().iterator().next();
```

k9mail/src/main/java/com/fsck/k9/Account.java
```diff
@@ -1201,10 +1201,7 @@ public synchronized boolean setFolderSyncMode(FolderMode syncMode) {
         if (syncMode == FolderMode.NONE && oldSyncMode != FolderMode.NONE) {
             return true;
         }
-        if (syncMode != FolderMode.NONE && oldSyncMode == FolderMode.NONE) {
-            return true;
-        }
-        return false;
+        return syncMode != FolderMode.NONE && oldSyncMode == FolderMode.NONE;
```
cketti: If you combine the last two `return` statements you might as well only have one. However, I think the old way makes the code easier to read/understand.

k9mail/src/main/java/com/fsck/k9/autocrypt/AutocryptHeader.java
```diff
@@ -100,20 +100,10 @@ public boolean equals(Object o) {
...
-        if (!addr.equals(that.addr)) {
-            return false;
-        }
-        if (!parameters.equals(that.parameters)) {
-            return false;
-        }
-
-        return true;
+        return isPreferEncryptMutual == that.isPreferEncryptMutual
+                && Arrays.equals(keyData, that.keyData)
```
cketti: Put operators at the end of the previous line.

k9mail/src/main/java/com/fsck/k9/controller/MessagingController.java
```diff
@@ -3773,11 +3773,7 @@ private boolean shouldNotifyForMessage(Account account, LocalFolder localFolder,
             return false;
         }
 
-        if (account.isNotifyContactsMailOnly() && !contacts.isAnyInContacts(message.getFrom())) {
-            return false;
-        }
-
-        return true;
+        return !account.isNotifyContactsMailOnly() || contacts.isAnyInContacts(message.getFrom());
```
cketti: I find the old way makes this easier to read.

k9mail-library/src/main/java/com/fsck/k9/mail/internet/MessageExtractor.java
```diff
@@ -447,24 +447,14 @@ private static Boolean isPartTextualBody(Part part) throws MessagingException {
             dispositionFilename = MimeUtility.getHeaderParameter(disposition, "filename");
         }
 
-        boolean isAttachmentDisposition = "attachment".equalsIgnoreCase(dispositionType) || dispositionFilename != null;
```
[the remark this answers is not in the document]
jyn514: For this one I added a comment instead of creating a variable; is that acceptable or would you prefer the variable?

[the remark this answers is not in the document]
cketti: We generally try avoid comments. It's easy to change code without updating comments. Giving a thing a descriptive name is often the easiest way to avoid a comment.

cketti:
Thanks

PR #2936, opened 2017-11-13
Split synchronization code

rhari991:
This is the first PR towards supporting CONDSTORE and QRESYNC.

This PR splits `synchronizeMailboxSynchronous` in `MessagingController` into two pathways(`ImapSyncInteractor` and `LegacySyncInteractor`) depending on whether the protocol is IMAP or non-IMAP. 

`ImapSyncInteractor` contains code that is copied over from `synchronizeMailboxSynchronous` and then split into multiple classes (`SyncHelper`, `FlagSyncHelper` and `MessageDownloader`).

`LegacySyncInteractor` contains code directly copied over from `synchronizeMailboxSynchronous`.

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/controller/FlagSyncHelper.java
+++ b/k9mail/src/main/java/com/fsck/k9/controller/FlagSyncHelper.java
@@ -109,7 +109,7 @@ class FlagSyncHelper<T extends Message> {
 
     private boolean syncFlags(LocalMessage localMessage, Message remoteMessage) throws MessagingException {
         boolean messageChanged = false;
-        if (localMessage == null || localMessage.isSet(Flag.DELETED)) {
+        if (localMessage == null) {
             return false;
         }
         if (remoteMessage.isSet(Flag.DELETED)) {
--- a/k9mail/src/main/java/com/fsck/k9/controller/FlagSyncHelper.java
+++ b/k9mail/src/main/java/com/fsck/k9/controller/FlagSyncHelper.java
@@ -1,0 +1,130 @@
+package com.fsck.k9.controller;
+
+
+import java.util.EnumSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import android.content.Context;
+import android.support.annotation.VisibleForTesting;
+
+import com.fsck.k9.Account;
+import com.fsck.k9.activity.MessageReference;
+import com.fsck.k9.helper.Contacts;
+import com.fsck.k9.mail.FetchProfile;
+import com.fsck.k9.mail.Flag;
+import com.fsck.k9.mail.Folder;
+import com.fsck.k9.mail.Message;
+import com.fsck.k9.mail.MessagingException;
+import com.fsck.k9.mailstore.LocalFolder;
+import com.fsck.k9.mailstore.LocalMessage;
+import com.fsck.k9.notification.NotificationController;
+import timber.log.Timber;
+
+
+class FlagSyncHelper<T extends Message> {
+
+    private static final Set<Flag> SYNC_FLAGS = EnumSet.of(Flag.SEEN, Flag.FLAGGED, Flag.ANSWERED, Flag.FORWARDED);
+
+    private final Context context;
+    private final MessagingController controller;
+    private final Contacts contacts;
+    private final NotificationController notificationController;
+    private final SyncHelper syncHelper;
+
+    public static FlagSyncHelper newInstance(Context context, MessagingController controller, SyncHelper syncHelper) {
+        Context appContext = context.getApplicationContext();
+        Contacts contacts = Contacts.getInstance(context);
+        NotificationController notificationController = NotificationController.newInstance(appContext);
+        return new FlagSyncHelper(appContext, controller, contacts, notificationController, syncHelper);
+    }
+
+    private FlagSyncHelper(Context context, MessagingController controller, Contacts contacts,
+            NotificationController notificationController, SyncHelper syncHelper) {
+        this.context = context;
+        this.controller = controller;
+        this.contacts = contacts;
+        this.notificationController = notificationController;
+        this.syncHelper = syncHelper;
+    }
+
+    void refreshLocalMessageFlags(final Account account, final Folder<T> remoteFolder, final LocalFolder localFolder,
+            List<T> syncFlagMessages) throws MessagingException {
+        final String folderName = remoteFolder.getName();
+        if (remoteFolder.supportsFetchingFlags()) {
+            Timber.d("SYNC: About to sync flags for %d remote messages for folder %s", syncFlagMessages.size(),
+                    folderName);
+
+            FetchProfile fp = new FetchProfile();
+            fp.add(FetchProfile.Item.FLAGS);
+
+            List<T> undeletedMessages = new LinkedList<>();
+            for (T message : syncFlagMessages) {
+                if (!message.isSet(Flag.DELETED)) {
+                    undeletedMessages.add(message);
+                }
+            }
+
+            remoteFolder.fetch(undeletedMessages, fp, null);
+
+            final AtomicInteger progress = new AtomicInteger(0);
+            int todo = syncFlagMessages.size();
+            for (Message remoteMessage : syncFlagMessages) {
+                processDownloadedFlags(account, localFolder, remoteMessage);
+                progress.incrementAndGet();
+                for (MessagingListener l : controller.getListeners()) {
+                    l.synchronizeMailboxProgress(account, folderName, progress.get(), todo);
+                }
+            }
+        }
+    }
+
+    @VisibleForTesting
+    void processDownloadedFlags(Account account, LocalFolder localFolder, Message remoteMessage)
+            throws MessagingException {
+        String folderName = localFolder.getName();
+        LocalMessage localMessage = localFolder.getMessage(remoteMessage.getUid());
+        boolean messageChanged = syncFlags(localMessage, remoteMessage);
+        if (messageChanged) {
+            boolean shouldBeNotifiedOf = false;
+            if (localMessage.isSet(Flag.DELETED) || syncHelper.isMessageSuppressed(localMessage, context)) {
+                for (MessagingListener l : controller.getListeners()) {
+                    l.synchronizeMailboxRemovedMessage(account, folderName, localMessage);
+                }
+            } else {
+                if (syncHelper.shouldNotifyForMessage(account, localFolder, localMessage, contacts)) {
+                    shouldBeNotifiedOf = true;
+                }
+            }
+
+            // we're only interested in messages that need removing
+            if (!shouldBeNotifiedOf) {
+                MessageReference messageReference = localMessage.makeMessageReference();
+                notificationController.removeNewMailNotification(account, messageReference);
+            }
+        }
+    }
+
+    private boolean syncFlags(LocalMessage localMessage, Message remoteMessage) throws MessagingException {
+        boolean messageChanged = false;
+        if (localMessage == null || localMessage.isSet(Flag.DELETED)) {
+            return false;
+        }
+        if (remoteMessage.isSet(Flag.DELETED)) {
+            if (localMessage.getFolder().syncRemoteDeletions()) {
+                localMessage.setFlag(Flag.DELETED, true);
+                messageChanged = true;
+            }
+        } else {
+            for (Flag flag : SYNC_FLAGS) {
+                if (remoteMessage.isSet(flag) != localMessage.isSet(flag)) {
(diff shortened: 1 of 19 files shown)


k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,180 @@
...
+            Timber.i("Done synchronizing folder %s:%s", account.getDescription(), folderName);
+
+        } catch (AuthenticationFailedException e) {
+            controller.handleAuthenticationFailure(account, true);
+            for (MessagingListener l : controller.getListeners(listener)) {
+                l.synchronizeMailboxFailed(account, folderName, "Authentication failure");
+            }
+        } catch (Exception e) {
+            Timber.e(e, "synchronizeMailbox");
+            // If we don't set the last checked, it can try too often during
```
philipwhiuk: This comment really wants moving to above line 135.

Or we could just test for the behaviour and describe the reason for the behaviour we want in the test name.

k9mail/src/main/java/com/fsck/k9/controller/LegacySyncInteractor.java
```diff
@@ -0,0 +1,336 @@
...
+import com.fsck.k9.mail.MessagingException;
+import com.fsck.k9.mail.Store;
+import com.fsck.k9.mailstore.LocalFolder;
+import com.fsck.k9.mailstore.LocalFolder.MoreMessages;
+import com.fsck.k9.mailstore.LocalMessage;
+import com.fsck.k9.mailstore.LocalStore;
+import com.fsck.k9.notification.NotificationController;
+import timber.log.Timber;
+
+/* This class contains code that used to be present directly in the MessagingController. It used to represent a common
```
philipwhiuk: Write this as proper Javadoc.

```
/**
 * This class...
 * It is used...
 **/
```

k9mail/src/main/java/com/fsck/k9/controller/MessageDownloader.java
```diff
@@ -0,0 +1,492 @@
...
+        final String folder = remoteFolder.getName();
+
+        final Date earliestDate = account.getEarliestPollDate();
+        remoteFolder.fetch(unsyncedMessages, fp,
+                new MessageRetrievalListener<T>() {
+                    @Override
+                    public void messageFinished(T message, int number, int ofTotal) {
+                        try {
+                            if (message.isSet(Flag.DELETED) || message.olderThan(earliestDate)) {
+                                if (K9.isDebug()) {
```
philipwhiuk: Maybe extract `processOldOrDeletedMessage()` to reduce the nesting here.

k9mail/src/main/java/com/fsck/k9/controller/SyncHelper.java
```diff
@@ -0,0 +1,262 @@
...
+import com.fsck.k9.mail.MessagingException;
+import com.fsck.k9.mailstore.LocalFolder;
+import com.fsck.k9.mailstore.LocalFolder.MoreMessages;
+import com.fsck.k9.mailstore.LocalMessage;
+import timber.log.Timber;
+
+
+class SyncHelper {
+
+    private static SyncHelper INSTANCE;
```
philipwhiuk: Why do you need an instance at all - these all seem to be static helper methods.

Also, this class could benefit from tests.

One idea is to make it an object, but inject it into classes. Then you could mock it. Only do that if it's useful though. If it's just a utils class then make it just a bunch of static methods and test them individually (as well as just using them in the rest of the code).

k9mail/src/main/java/com/fsck/k9/mailstore/LocalFolder.java
```diff
@@ -894,6 +894,42 @@ public LocalMessage doDbWork(final SQLiteDatabase db) throws WrappedException, U
         }
     }
 
+    public LocalMessage getMessageUidAndFlags(final String uid) throws MessagingException {
```
philipwhiuk: No tests.

k9mail/src/main/java/com/fsck/k9/controller/SyncHelper.java
```diff
@@ -0,0 +1,262 @@
...
+                || (aMode == Account.FolderMode.FIRST_CLASS &&
+                fMode != Folder.FolderClass.FIRST_CLASS)
+                || (aMode == Account.FolderMode.FIRST_AND_SECOND_CLASS &&
+                fMode != Folder.FolderClass.FIRST_CLASS &&
+                fMode != Folder.FolderClass.SECOND_CLASS)
+                || (aMode == Account.FolderMode.NOT_SECOND_CLASS &&
+                fMode == Folder.FolderClass.SECOND_CLASS);
+    }
+
+    void evaluateMessageForDownload(final Message message, final String folderName, final LocalFolder localFolder,
```
philipwhiuk: Prefer `remoteMessage` to `message`.

k9mail/src/main/java/com/fsck/k9/mailstore/LocalMessage.java
```diff
@@ -132,6 +104,38 @@ void populateFromGetMessageCursor(Cursor cursor) throws MessagingException {
...
+            }
+        }
+
+        boolean deleted = (cursor.getInt(cursor.getColumnIndex("deleted")) == 1);
+        boolean read = (cursor.getInt(cursor.getColumnIndex("read")) == 1);
+        boolean flagged = (cursor.getInt(cursor.getColumnIndex("flagged")) == 1);
+        boolean answered = (cursor.getInt(cursor.getColumnIndex("answered")) == 1);
+        boolean forwarded = (cursor.getInt(cursor.getColumnIndex("forwarded")) == 1);
+
+        setFlagInternal(Flag.DELETED, deleted);
```
philipwhiuk: Anyone know if it's possible to (for instance) have deleted in the database as 0 but DELETED in the `flags` column`?

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,180 @@
...
+                    Timber.d("SYNC: Expunging folder %s:%s", account.getDescription(), folderName);
+                    imapFolder.expunge();
+                }
+                imapFolder.open(Folder.OPEN_MODE_RO);
+            }
+
+            notificationController.clearAuthenticationErrorNotification(account, true);
+
+            int remoteMessageCount = imapFolder.getMessageCount();
+            if (remoteMessageCount < 0) {
```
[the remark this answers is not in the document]
philipwhiuk: (Don't try and fix this here though, it's more me griping about the current code).

k9mail/src/main/java/com/fsck/k9/controller/FlagSyncHelper.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @VisibleForTesting
+    void processDownloadedFlags(Account account, LocalFolder localFolder, Message remoteMessage)
+            throws MessagingException {
+        String folderName = localFolder.getName();
+        LocalMessage localMessage = localFolder.getMessage(remoteMessage.getUid());
+        boolean messageChanged = syncFlags(localMessage, remoteMessage);
+        if (messageChanged) {
+            boolean shouldBeNotifiedOf = false;
+            if (localMessage.isSet(Flag.DELETED) || syncHelper.isMessageSuppressed(localMessage, context)) {
```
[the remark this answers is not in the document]
rhari991: Hmm yeah, I'm not sure why it was written like that. I've rewritten the code to notify the listeners in this case.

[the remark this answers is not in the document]
rhari991: Yeah I'm not sure why it was written that way.

k9mail/src/main/java/com/fsck/k9/controller/FlagSyncHelper.java
```diff
@@ -0,0 +1,130 @@
...
+            if (!shouldBeNotifiedOf) {
+                MessageReference messageReference = localMessage.makeMessageReference();
+                notificationController.removeNewMailNotification(account, messageReference);
+            }
+        }
+    }
+
+    private boolean syncFlags(LocalMessage localMessage, Message remoteMessage) throws MessagingException {
+        boolean messageChanged = false;
+        if (localMessage == null || localMessage.isSet(Flag.DELETED)) {
```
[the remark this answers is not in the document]
rhari991: Removed

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,180 @@
...
+        LocalFolder localFolder = null;
+        ImapFolder imapFolder = null;
+
+        try {
+            Timber.d("SYNC: About to process pending commands for account %s", account.getDescription());
+
+            try {
+                controller.processPendingCommandsSynchronous(account);
+            } catch (Exception e) {
+                Timber.e(e, "Failure processing command, but allow message sync attempt");
```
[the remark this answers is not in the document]
rhari991: Yeah this has the potential to overwrite the user's actions that have not been sent to the server.

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
```diff
@@ -0,0 +1,180 @@
...
+                if (providedRemoteFolder instanceof ImapFolder) {
+                    Timber.v("SYNC: using providedRemoteFolder %s", folderName);
+                    imapFolder = (ImapFolder) providedRemoteFolder;
+                } else {
+                    throw new IllegalArgumentException("A non-IMAP account was provided to ImapSyncInteractor");
+                }
+            } else {
+                imapFolder = getImapFolder(account, folderName);
+
+                if (!syncHelper.verifyOrCreateRemoteSpecialFolder(account, folderName, imapFolder, listener, controller)) {
```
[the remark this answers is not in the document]
rhari991: Fixed

k9mail/src/main/java/com/fsck/k9/controller/ImapSyncInteractor.java
rhari991: Moved the comment

k9mail/src/main/java/com/fsck/k9/controller/LegacySyncInteractor.java
rhari991: Done

k9mail/src/main/java/com/fsck/k9/controller/MessageDownloader.java
rhari991: Done

k9mail/src/main/java/com/fsck/k9/controller/SyncHelper.java
rhari991: Changed

k9mail/src/main/java/com/fsck/k9/mailstore/LocalFolder.java
rhari991: I think it's a bit out of scope here. cketti mentioned that the tests for `LocalFolder` should test against a real SQLite db in https://github.com/k9mail/k-9/pull/2694#issuecomment-324682921

k9mail/src/main/java/com/fsck/k9/controller/SyncHelper.java
rhari991: I seem to remember that these methods used to be static, but then there were some complications in testing. I'll look again.

cketti:
When I suggested splitting out the IMAP sync code my intention was to completely separate the code from the "legacy" sync code so it can be evolved independently. This first attempt still shares code between the two sync implementations and doesn't bring us much closer to that goal.
Maybe it helps to know that in a second step I want us to move the "remote store" abstraction further up so that it includes most of what lives in `MessagingController` right now. That allows implementations to leverage protocol specific features to realize the high level operations (sync, push, copy, move, delete, change read state, etc.) more efficiently.

Don't worry about duplicated code right now. We can look into adding abstractions that allow code reuse later, once we have a better idea what the different implementations look like in isolation.

If that all sounds too confusing, let me know and I'll create a pull request with what I had in mind.

cketti:
Another thing: I don't like class names ending in "Helper". They're usually an indicator that the design can be improved. For now let's try to keep things in a single class. That should make it easier to change the code. We can always split it into multiple classes later.

Also, I don't think the "Interactor" suffix adds any value. `ImapSync` and `LegacySync` sound fine to me.

cketti:
I created PR #2996 with the first step of what I had in mind.

cketti:
Abandoned in favor of #2996 and (future) follow-up PRs.

PR #3075, opened 2018-01-10
Prefer encrypt setting

Valodim:
Adds a setting that enables Autocrypt's `prefer-encrypt=mutual`. The logic is trivial since it was already there, the hard part is the wording I guess. I'm ok with this for a first attempt, but there is probably room for improvement. In a way, the goal is to get the user *not* to just blindly enable this setting, like any variant of "Prefer encryption" or "Encrypt by default" would do.

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSettings.java
+++ b/k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSettings.java
@@ -9,6 +9,7 @@ import java.util.Map;
 
 import android.app.Dialog;
 import android.content.Context;
+import android.content.DialogInterface.OnCancelListener;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.AsyncTask;
@@ -22,6 +23,8 @@ import android.preference.Preference.OnPreferenceChangeListener;
 import android.preference.Preference.OnPreferenceClickListener;
 import android.preference.PreferenceScreen;
 import android.preference.RingtonePreference;
+import android.widget.CompoundButton;
+import android.widget.RadioGroup.OnCheckedChangeListener;
 import android.widget.Toast;
 
 import com.fsck.k9.Account;
@@ -46,6 +49,8 @@ import com.fsck.k9.mail.Folder;
 import com.fsck.k9.mail.Store;
 import com.fsck.k9.mailstore.StorageManager;
 import com.fsck.k9.service.MailService;
+import com.fsck.k9.ui.dialog.AutocryptPreferEncryptDialog;
+import com.fsck.k9.ui.dialog.AutocryptPreferEncryptDialog.OnPreferEncryptChangedListener;
 import org.openintents.openpgp.util.OpenPgpKeyPreference;
 import timber.log.Timber;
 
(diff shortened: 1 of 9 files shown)


k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSettings.java
```diff
@@ -711,6 +712,9 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
             });
 
             cryptoMenu.setOnPreferenceClickListener(null);
+
+            autocryptPreferEncryptMutual = (CheckBoxPreference) findPreference(PREFERENCE_NOTIFICATION_OPENS_UNREAD);
```
philipwhiuk: This is wrong.

k9mail/src/main/res/xml/account_settings_preferences.xml
```diff
@@ -479,6 +479,12 @@
             android:key="crypto_key"
             android:title="@string/account_settings_crypto_key" />
 
+        <CheckBoxPreference
+            android:key="autocrypt_prefer_encrypt"
```
philipwhiuk: I'd rather setting was either completely crypto agnostic or not. i.e. just "prefer_encrypt"

k9mail/src/main/res/values/strings.xml
```diff
@@ -1280,4 +1280,7 @@ Please submit bug reports, contribute new features and ask questions at
     <string name="openpgp_enabled_error_back">Back</string>
     <string name="openpgp_enabled_error_disable">Disable Encryption</string>
     <string name="openpgp_encryption">OpenPGP Encryption</string>
+    <string name="account_settings_crypto_prefer_encrypt">Encryption preference</string>
+    <string name="account_settings_crypto_prefer_encrypt_on">By default when replying to encrypted, or if all recipients also checked this option</string>
```
philipwhiuk: `also checked this option` -> `also would like encryption`

k9mail/src/main/res/values/strings.xml
```diff
@@ -1280,4 +1280,7 @@ Please submit bug reports, contribute new features and ask questions at
...
+    <string name="account_settings_crypto_prefer_encrypt_off">By default, when replying to encrypted messages</string>
```
philipwhiuk: 'when' -> 'only when'

k9mail/src/main/res/xml/account_settings_preferences.xml
Valodim: This is not a generic setting - unlike any other, the semantics are precisely defined by autocrypt. Hence the name

k9mail/src/main/java/com/fsck/k9/activity/setup/AccountSettings.java
Valodim: Derp.

k9mail/src/main/res/values/strings.xml
Valodim: This is a really tough one. Everyone generally "likes" or "prefers" encryption over not having that. But it's quite specifically checking this particular option (and only in an Autocrypt-capable client) that triggers it... hmm. I wonder if we should have a specific name (that can be exposed to users) for this setting it in the spec.

philipwhiuk:
Do you need a preferences upgrade process.

Also, do we need to do anything to make sure this preference is exported in the XML (can't remember if there's a code change needed there or not).

Valodim:
Not happy yet with the captions

PR #3089, opened 2018-01-13
Tidy up the usage of 'e-mail' in non-translated source text

philipwhiuk:
`grep -rni "e-mail" k9mail/src/ | grep -v "values\-" | grep -v "test"`

There's only one that's actually visible, the rest are comments.

Thanks to #3086


philipwhiuk:
There's a lot more on the website, but I'll wait for this to go in before I fix-up the website. Besides the website in general requires much work.

One option to prevent this is a `lint` check of course but I'd rather see contributions than add more hurdles.

PR #3105, opened 2018-01-17
Autocrypt gossip

Valodim:
This adds support for receiving and sending Autocrypt-Gossip headers. It's still a WIP PR, I'm currently working on the OpenKeychain before finishing this.

It depends on https://github.com/open-keychain/open-keychain/pull/2253, which changes the Autocrypt logic quite a bit to work according to Autocrypt 1.0 spec.

All of this is both forward-compatible (newer openkeychain is compatible to current k9), and backward-compatible (newer k9 works with old openkeychain, but silently fails sending or receiving gossip).


Valodim:
I thought so as well, that's when https://github.com/autocrypt/autocrypt/issues/345 came up ;) Will add those.

Valodim:
Aight, added a bunch of unit tests

PR #3166, opened 2018-02-03
update to most recent build tools

hannesa2:
and update almost all libs to most recent,
except glide (it causes some issues)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,15 @@ language: android
 android:
  components:
     - tools
-    - build-tools-27.0.2
+    - build-tools-27.0.3
     - android-26
     - extra-android-m2repository
+
+  licenses:
+    - 'android-sdk-preview-license-.+'
+    - 'android-sdk-license-.+'
+    - 'google-gdk-license-.+'
+    - '.+'
 
 jdk: oraclejdk8
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: android
 android:
  components:
     - tools
-    - build-tools-25.0.2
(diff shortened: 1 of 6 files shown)


hannesa2:
Travis seams top have some troubles, I can't solve this
image

fat-tire:
Have you seen #3159?  It does this + runtime permissions that would be needed when updating + more

cketti:
Thanks for the reminder. I will look into updating to the latest build tools.

hannesa2:
Just to understand, you @cketti closed this PR and similar #3159.
That means PR makes no sense here, because you do everything by yourself ?

cketti:
That question could have been answered by looking at past pull requests. Of course we do accept contributions.

But in this particular case I'd like to do it myself. Updating build tools and dependencies does have implications. And since I don't know you and your familiarity with the K-9 Mail code base I'd rather err on the safe side and assume your understanding of our code base is not good enough to assess the impact of the changes listed in the changelogs of the dependencies.

Some examples of problems with your pull request: 
* We can't update `commons-io` because we still support versions of Android that included version 2.4 of the library in an app's runtime classpath. To be fair, I should have documented that after running into it.
* `compileSdkVersion` is lower than the support library major version
* The PR didn't build on Travis
* The `WebDav*` tests fail. Probably due to some changes in the Android Gradle plugin

I'll try to allude to the difficulty of doing these kind of infrastructure upgrades when closing the next pull request like this. I didn't want to sound ungrateful. It's just that this is complicated and it's much easier for me to do it myself than explaining the many things that have to be taken into account.

PR #3274, opened 2018-03-22
Fix crash with pendingintents in messageview

Valodim:
Only the lower 16 bits can be used for the request code from fragments. This changes our own mask to use the 15th bit instead of the 16th.


Valodim:
Yes, this one should be backported for sure

Valodim:
Looked over issues and there are a bunch of crashes in the area, but none I could really attribute to this.

PR #3290, opened 2018-03-28
Introduce lifecycle-based OpenpgpApiManager for service connection

Valodim:
This offloads the openpgp service connection code into a specialized class (-200 lines recipientpresenter, yay). This class behaves fairly passively, returning simple callbacks and leaving all handling of error states to the caller.

I plan to use this in other places as well, including the e2e settings dialog. At the moment there are a couple of states that show pending intents for openpgp setup stuff, but I want to change those to go to the e2e settings in a follow-up PR.


Valodim:
The fixes related to #3292 aren't in here yet, will be in a separate PR for the openpgp key setting.

PR #3295, opened 2018-03-30
Added icons to preference list items

hendrikbl:
I added icons to the items in the setings list. I'm not 100% satisfied with the network icon but since it has only one sync setting I thought it would be fitting.

grafik


Valodim:
Nice, I like it! :+1:

Valodim:
rebased and merged as https://github.com/k9mail/k-9/commit/d0b3401c52ebce0f2763ee4687a78a69dc7c7929

PR #3319, opened 2018-04-09
Encrypted subject setting

Valodim:
What it says on the tin


philipwhiuk:
On/off per account is quite simplistic but it's a good start.

PR #3324, opened 2018-04-10
Limit MIME encode range in Content-Disposition to avoid over-encode

sksthrs:
Resolution about Issue #3323 .

This PR handle Content-Disposition as Content-Type (where only filenames are encoded in MIME).

This style ("filename" in Content-Disposition encoded in MIME) is not fully-RFC-compatible (Formally it should be encoded in RFC2231), but many mail application (like Microsoft Outlook) use this style.

### DIFF
--- a/k9mail/src/main/java/com/fsck/k9/message/MessageBuilder.java
+++ b/k9mail/src/main/java/com/fsck/k9/message/MessageBuilder.java
@@ -257,7 +257,8 @@ public abstract class MessageBuilder {
              */
             bp.addHeader(MimeHeader.HEADER_CONTENT_DISPOSITION, String.format(Locale.US,
                     "attachment;\r\n filename=\"%s\";\r\n size=%d",
-                    attachment.name, attachment.size));
+                    EncoderUtil.encodeIfNecessary(attachment.name, EncoderUtil.Usage.WORD_ENTITY, 7),
+                    attachment.size));
 
             mp.addBodyPart(bp);
         }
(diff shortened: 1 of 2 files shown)


philipwhiuk:
So we're breaking RFC compatibility to produce the same output? Or are we tolerating bad input.

I'm not keen on the first one.

cketti:
I'm with @philipwhiuk. I don't see a reason to deliberately violate RFCs just because others do the same. But I'm fine with adding support for reading those messages.

sksthrs:
My PR is not "breaking".
Maybe the first explanation was not enough.

Currently, K-9 is not RFC-compatible about handling "filename" in Content-Disposition.
"filename" (in Content-Disposition) should be encoded as RFC2231, but K-9 encodes in MIME (not RFC2231).

Plus, K-9 encodes both "key and value" (only value should be encoded).
So, attachment from K-9 itself disappears after sending.

I attached two images.

First one (k-9_attach_1.png) is a screenshot BEFORE sending.

k-9_attach_1

Second one (k-9_attach_2.png) is a screenshot AFTER receiving same mail, attachment disappeared.

k-9_attach_2

My PR fixes second point (To encode only value).
First point (encode in MIME, not RFC2231) is same as now. No breaking.

This is not ideal solution, but can fix real problem.
Do you only accept wholly RFC-compatible way?

philipwhiuk:
Okay. I think I know what I want to do with this:

I'd like to add tests, that cover the full gamut of RFC 2231 and it's interaction with RFC 2047 both for `Content-Type` (which is the RFC example) and `Content-Disposition`.

We can check we have tests that cover just the bit this PR fixes and we can @Ignore the other tests and create a new issue to cover that.

Valodim:
How about merging this and working from there? Even if there is more to do, this is better than what we have. We should make sure this doesn't get lost in the noise if we forget.

philipwhiuk:
I've refactored the MessageBuilder slightly and tidied up the test class a bit in #3352

PR #3335, opened 2018-04-13
Update providers.xml Yandex.{by,kz,ua}

bboa:


bboa:
@philipwhiuk, you suggest to leave Yandex in Russian providers place?
Not a big problem, I simply think that it's a bit wrong logic, to place by,kz into ru.
Or, you suggest to keep narod.ru provider?

I can make changes

PR #3349, opened 2018-04-19
Cache decrypted subjects in message table

Valodim:
With this PR, encrypted subjects are replaced with their plaintext in the `messages` table once the message has been decrypted.

I went for a dedicated `decrypted_subject` column first, but that was three times as much code and made things like message ordering unnecessarily complex. This approach simply replaces the text in the `subject` column, and adds a X_SUBJECT_DECRYPTED flag to avoid replacing it redundantly in the future.


philipwhiuk:
LGTM

PR #3396, opened 2018-05-17
Search function for general settings

ByteHamster:
Added a way to search general settings. The search function is currently set to fuzzy matching.

Disclaimer: I developed the library.


cketti:
I like the idea of being able to search settings. But I'm not sure if I want us to display search so prominently. Maybe as an action in the toolbar?

cketti:
For many `Preference` sub-classes `%s` is valid in the `summary` attribute and should be replaced by the current preference value. I guess your options are:
* replace `%s` with the value manually
* don't display summaries that contain `%s`
* have a separate attribute for a setting description that is only displayed in search

ByteHamster:
@cketti I added an attribute that is only displayed in search. Additionally, I moved the search function to the options menu.

cketti:
It looks like you based the user experience on the settings app in Android 8. But I feel the search widget doesn't work well when search is a toolbar action. In that case it feels more natural to use a `SearchView`. I'd like us to have something similar to the way settings search was done in Android 7:

 

The additional attribute for settings descriptions introduces a lot of redundancy/noise. I think it's better to not display a description at all. I do love the settings path (General Settings > Display > Global), though :+1: 

I'm sorry for being so picky :disappointed: I really appreciate all the work you're putting into this.

ByteHamster:
> The additional attribute for settings descriptions introduces a lot of redundancy/noise. I think it's better to not display a description at all

Not display in cases where it needs a placeholder (%s) or never ever? If it's only for the placeholder, I would prefer to keep the additional attribute (but leave the values empty, so we not need more strings). The screenshot above shows redundant descriptions (they had been placeholders), but in General, I think they are really useful.

> I'd like us to have something similar to the way settings search was done in Android 7

I will see if I can do that without making the library too specific to K9. Should be possible, I guess.

hendrikbl:
I agree with @cketti . In fact it says in the [material design guidelines](https://material.io/design/navigation/search.html) that persistent searchbars should only be used when search is the primary focus of the app, which is not the case I guess.

ByteHamster:
I already changed the behavior:

screenshot_k-9_mail_20180531-095055

The PR is still not completely ready:

- [ ] In some cases the search bar is not closed
- [ ] The search bar is not updated when selecting a history entry

cketti:
Looks and feels very nice now! Some small issues:
* Dismissing the search text input field by pressing the "X" button leaves an empty screen (presumably the search fragment with 0 results).
* The text input widget looks like it starts too far to the right.
* The search results list is displayed with padding, whereas the regular settings view doesn't use padding outside of the list.

ByteHamster:
@cketti Since I merged your commits that split the modules, I can no longer build debug builds without enabling multidex. Looks like adding my library adds the few more methods that make the app too big. Before splitting the modules, it worked without multidex. **Be careful. This PR enables multidex currently.**

> The text input widget looks like it starts too far to the right.

I did not manage to fix this. I am just deriving from `android.support.v7.widget.SearchView` without changing the layout. Seems to be either intended by Google or a bug in their library.

ByteHamster:
@cketti Would be nice if you could review (and maybe merge) my PR because the current restructuring of the app requires solving lots of merge conflicts currently.

ByteHamster:
I have updated the PR with your changes included. Does not look much different to me. By the way, do you plan to migrate the font settings screen to a fragment in the future? The search function currently just calls the activity but does not highlight the search result.

cketti:
> Does not look much different to me.

Indeed. The position of the `SearchView` still looks off. Any hints as to why that is are welcome. This is not a blocker, though.

> By the way, do you plan to migrate the font settings screen to a fragment in the future?

Sure, eventually. Any help with that would be highly appreciated.

Looks like I was wrong when I said this feature was almost ready to be merged. Unfortunately, `SearchView´ doesn't play well with orientation changes. It looks like we have to manually save and restore the state of it :(

ByteHamster:
@cketti: Ping ;)

Is there something left to do?

Issue #3510, opened 2018-07-19
More consistent default positioning of menu items


Meteor0id:
I noticed today that in the overview of accounts, the "new message" icon is in the middle of the menu bar, wit to it's right a plus symbol to add new accounts.

When going into an account, the new message symbol shifts one position to the right.

To me this has lead to some unconcious confusion many times, when I intent to write a new email but accidentally press the wrong button by muscle memory. Only now I realised this is because the icon is not in the same position in all menu's where it appears.

Suggestion:
**On the account overview menu, swap the "add account" button and the "new message button", so that buttons which occur on multiple views are consistently positioned accros those views.**

It's a minor detail, but it makes life easier.

Meteor0id:
I had a more indept look at the sorting of menu items. I now see that on ladscape mode the first 3 items are shown as icons, the rest as menu items.
composing emails is of upmost importance on a email client, therefor it deserves a place on the menu bar as an icon.

Instead of switching the compose buttons with the new identity button on the identities overview, I suggest switching "compose" with "sort"on the inbox view. This way the compose button is prominetly displayed, while the sort button becomes the fisrt item in the menu. Composing is probably done more spontaniously and more often than sorting.

Meteor0id:
Looking at the code, I haven't been able to figure out where the order of menu items is determined.

PR #3511, opened 2018-07-19
Attempt to order menu items similar across views

Meteor0id:
https://github.com/k9mail/k-9/issues/3510

I am not sure what I am doing, please check test and correct what I am attempting to do.

Not all views display the same items. Attempt to organize items in the same order if they occur in multiple views.

### DIFF
--- a/app/ui/src/main/res/menu/message_list_option.xml
+++ b/app/ui/src/main/res/menu/message_list_option.xml
@@ -16,7 +16,8 @@
         android:id="@+id/previous_message"
         android:title="@string/previous_action"
         android:icon="?attr/iconActionPreviousMessage"
-        app:showAsAction="always" />
+        app:showAsAction="always"
+        android:orderInCategory="1"/>
 
     <!-- MessageView -->
     <item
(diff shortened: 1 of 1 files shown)


Valodim:
Hi, sorry for the delay. It's very hard to tell what was moved where, and why. Can you describe what your considerations were?

Meteor0id:
I made an attempt to place icons in the menu consistently across views. I do not know for certain how the order of icons in defined in Android, so my attempt might not be the best way to do that.

What I attempt to do is to make sure the compose button is not hidden in landscape mode, and also make sure the compose button does not shift positions between the account overview and the inbox view. It is quite a nuisance when the button shifts around between views. Some users act on muscle memory, and after adding a second or more accounts that means we press the wrong button very often. The solution is quite simple:
On the account overview menu, swap the "add account" button and the "new message button", so that buttons which occur on multiple views are consistently positioned across those views.

Meteor0id:
Somehow I thought it was much more complicated than the order of items in the xml file. But alright, fixed it.

So the only thing this PR does is move Compose ahead of sort, so the compose button does not switch positions between the account overview and the inbox view. That's all.

PR #3528, opened 2018-07-26
Add Option for Archive Action in Notification

jkuester:
Fixes https://github.com/k9mail/k-9/issues/3403

Adds a new notification option allowing the user to chose to have an 'Archive' button on new message notifications instead of the 'Delete' button.

screenshot_k-9_mail_notifications

screenshot_k-9_mail_move_action

screenshot_k-9_mail_move_trigger

screenshot_trebuchet_actual_notification


jkuester:
No worries!  As I was implementing this it struck me that the existing approach to notification actions was lacking.  However, I did not want to try and just totally refactor it without some input on a go-forward vision.  I figured I would submit what I had and either it would be accepted, or it would spark a conversation about a better direction to go.  

So, thank you for the vision in #3530!

PR #3551, opened 2018-08-12
Added password Encryption method to avoid storing password as plain pa…

kramprabhu:
Added password encryption method to avoid storing password as plain password in user device.


kramprabhu:
Resubmitting with different account.

PR #3635, opened 2018-10-01
Change ReplyToParser to prefer From over List-Post header

vzickner:
Behavior before: ReplyToParser preferred List-Post header over the From header which made the list to the default reply-to address.

Behavior after: ReplyToParser prefers the From header over the List-Post header since mailing list tools like mailman also set the Reply-To header to the mailing list in case the admin would like to have the mailing list as default return. This follows also the behavior that most desktop mail clients are using (like Thunderbird, Apple Mail and Outlook).

This PR doesn't contain a reference to an issue ID since I didn't create a bug report or feature request.

### DIFF
--- a/app/core/src/main/java/com/fsck/k9/helper/ReplyToParser.java
+++ b/app/core/src/main/java/com/fsck/k9/helper/ReplyToParser.java
@@ -26,10 +26,10 @@ public class ReplyToParser {
 
         if (replyToAddresses.length > 0) {
             candidateAddress = replyToAddresses;
-        } else if (listPostAddresses.length > 0) {
-            candidateAddress = listPostAddresses;
-        } else {
+        } else if (fromAddresses.length > 0) {
             candidateAddress = fromAddresses;
+        } else {
+            candidateAddress = listPostAddresses;
         }
 
         boolean replyToAddressIsUserIdentity = account.isAnIdentity(candidateAddress);
--- a/app/core/src/test/java/com/fsck/k9/helper/ReplyToParserTest.java
+++ b/app/core/src/test/java/com/fsck/k9/helper/ReplyToParserTest.java
@@ -75,27 +75,27 @@ public class ReplyToParserTest extends RobolectricTest {
     }
 
     @Test
-    public void getRecipientsToReplyTo_should_prefer_listPost_over_from_field() throws Exception {
+    public void getRecipientsToReplyTo_should_prefer_from_over_listPost_field() throws Exception {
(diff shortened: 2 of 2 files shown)


vzickner:
This changes the behavior which was initially implemented in https://github.com/k9mail/k-9/commit/2e33e0dd9492fe2ededf5eaf17495a5dd5ff6f67

Valodim:
Hmm, I would think replying to the list is the expected behavior in most cases, and replying to the original author is the exception. On the other hand, for non-list mails "reply" replies to the sender only unlike "Reply to all". I guess there is no truly natural behavior for a "reply" button, so maybe we should offer "Reply to list" and "Reply to sender" buttons?

vzickner:
Hi @Valodim,

those two buttons would make it actually clear.

The reason for my pull request was initially that I replied by accident to a mailing list instead of the sender only. Most of the time it is worse when you are sending the email to a large group of persons than to one specific person by accident. When you would like to reply to the list I would maybe also reply to the sender as well (he might not be on the list) so I would say there the "reply all" button is also a fit. Are there other email applications which are going for a "reply to list" as a default reply button?

Thank you for your feedback!

Valentin

Valodim:
> Most of the time it is worse when you are sending the email to a large group of persons than to one specific person by accident

Agreed.

> When you would like to reply to the list I would maybe also reply to the sender as well (he might not be on the list) so I would say there the "reply all" button is also a fit

Agreed.

> Are there other email applications which are going for a "reply to list" as a default reply button?

I don't know :) I guess it also depends on the placement of the button - if there is a single big "reply" button and alternative ones a click away, it should clearly cover the common case, and I'm fairly sure that that's participating in the on-list discussion, rather than branching off to only the recipient.

I think we can be smart about those reply options, i.e. "reply" / "reply to all" for regular mails and "reply to sender" / "reply to list" for those with a list-post header. Would you want to take a shot at that? The info should be easily available in `RecipientPresenter` :)

vzickner:
I can give it a shot, but it might take some time since I pretty busy right now.

ilmaisin:
> Thanks for taking the time to contribute to K-9 Mail by creating this PR! Sorry, we didn't include it. But I think it started an important discussion.
> 
> We'll stick with the current behavior for now and change once we've figured out how we want reply to behave in the future. See #3738.

Or what about just doing it? Internet standards are not written hastily. Things like this have been already discussed throughly. No point in implementing surprising, non-standard behaviour.

Issue #3676, opened 2018-10-22
Ipv6 address not possible for server


paphko:
When setting up a new imap account, it is not possible to enter an ipv6 address like "2002:cc:d456:d300:654:33ff:fe33:65ea" as server address. The next button is simply not enabled. As soon as a colon is contained in the address, I cannot continue. Surrounding the ip with square brackets does also not work. It seems to me as a bug - or did I miss something?

ByteHamster:
You could try to enter the ip with square brackets:

    [abcd:efab:...]

paphko:
I tried that already, but the button is still not enabled...

GoneUp:
It looks like the function for validation domains only supports domains or IPv4 addresses. Code is here https://github.com/k9mail/k-9/blob/3af2e6c022e11e2dc534f14cb1820895858d4376/app/core/src/main/java/com/fsck/k9/helper/Utility.java#L110

realdadfish:
Normally entering a raw IPv4 or IPv6 address shouldn't be needed, but after switching phones I figured that my SMTP configuration no longer works. I suspect a broken IPv6 connection, because if I enter the IPv4 address of my server (and accept the certificate dialog), the SMTP connection test works. But due to the above error, I cannot really verify that it's IPv6-related actually.

PR #3705, opened 2018-11-11
Allow literal ipv6 hostname

artbristol:
Per issue #3676, the inputted hostname for server settings can now be a raw IPv6 address

Not sure if this very minor enhancement is worth the overhead of adding guava, but on the other hand it's quite a useful library in general


Valodim:
Sounds good. Hopefully the relevant methods don't have too many dependencies though

artbristol:
I have tried to copy the functionality but alas, it's a rabbit-hole. `InternetDomainName` has about 8 guava dependencies, including `ImmutableList`, which of course has other guava dependencies, etc etc.

Guava only adds 100KB to the apk size, or is it the conceptual weight you're worried about?

wiktor-k:
An alternative would be to a regexp (there are many examples) or... not to validate it! Just try to connect and if it works then it works.

PR #3741, opened 2018-11-22
Show accounts in the unified inbox drawer

morckx:
Currently, when in unified inbox mode, the drawer has an empty body, in
the best case. Simply list the mail accounts in the body of the unified
inbox drawer to allow for a quick navigation into them.

### DIFF
--- a/app/core/src/main/java/com/fsck/k9/Account.java
+++ b/app/core/src/main/java/com/fsck/k9/Account.java
@@ -16,6 +16,8 @@ import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 
 import android.content.Context;
+import android.graphics.drawable.Drawable;
+import android.graphics.drawable.GradientDrawable;
 import android.net.Uri;
 import android.support.annotation.Nullable;
 import android.text.TextUtils;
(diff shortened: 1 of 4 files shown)


app/core/src/main/java/com/fsck/k9/Account.java
```diff
@@ -821,6 +823,15 @@ public String getDisplayName() {
         return description != null ? description : getEmail();
     }
 
+    public Drawable getIcon() {
```
Valodim: I like this idea, but please move this into a separate class. Account is overloaded as it is.

app/ui/src/main/java/com/fsck/k9/activity/MessageList.java
```diff
@@ -619,6 +611,7 @@ public void openFolder(String folderName) {
 
     public void openUnifiedInbox() {
         drawer.selectUnifiedInbox();
+        drawer.setUserAccounts(preferences.getAccounts());
```
Valodim: This should probably use AccountsLiveData in some manner, which ties in with #3597

morckx:
> This looks like it is halfway there, if the accounts were shown in a persistent header, rather than as drawer items?
Absolutely! I finally had a look at gmail and the MaterialDrawer-Library and this seems indeed much better and even easier.

app/ui/src/main/java/com/fsck/k9/activity/MessageList.java
morckx: Yes, I'll keep an eye on that.

app/core/src/main/java/com/fsck/k9/Account.java
morckx: Well, it somehow belongs there, especially if we want to reuse the icons when AccountsLiveData is ready. However, if we switch to materialdrawer.AccountHeader it would make sense to use ContactPictures here, I guess. I'll have a look...

Valodim:
Awesome :+1:

morckx:
Replaced by #3745

PR #3743, opened 2018-11-23
Message reply button

Valodim:
Inspired by gmail app, I moved the reply button next to the sender, to where the "flag" star was before. But compared to the reply button it doesn't deserve a highly prominent position like that, maybe we can place it somewhere else?

image


cketti:
Looks nice. But won't it be annoying having to scroll up once a message has been read from top to bottom to be able to reply?

> Inspired by gmail app, I moved the reply button next to the sender, to where the "flag" star was before. But compared to the reply button it doesn't deserve a highly prominent position like that, maybe we can place it somewhere else?

Do it like the Gmail app? Always display the subject below the action bar. Put the star to the right of the subject?

Valodim:
>  But won't it be annoying having to scroll up once a message has been read from top to bottom to be able to reply?

The "send" menu is still in overflow. Gmail covers this with a bottom bar below the message.

> Do it like the Gmail app? Always display the subject below the action bar. Put the star to the right of the subject?

Can do. It's a bit more work so I wanted to hear feedback first (and split up the PRs)

Valodim:
image

how bout this?

wiktor-k:
Wow, looks *very nice*, I can't wait to build it and try it out live!

cketti:
Nice.

I'm not sure about the position of the encryption indicator. But that's easily something we can figure out later.

Valodim:
I moved the encryption indicator so the icons form a triangle rather than a jagged pattern. There's probably still room for improvement, but overall it's certainly better than before :)

Valodim:
And having them bigger with descriptions, too. I heard more than one person say "I can never remember which one of those buttons is for replying" on the old bottom bar. Personally I thought the arrow is the single obvious candidate, but people are reluctant to click on stuff without confirmation it's the right thing. So having big buttons with descriptions is definitely a good idea. That can be another PR, though :)

app/ui/src/main/java/com/fsck/k9/activity/MessageList.java
```diff
@@ -946,21 +940,6 @@ public boolean onOptionsItemSelected(MenuItem item) {
         } else if (id == R.id.delete) {
             messageViewFragment.onDelete();
             return true;
-        } else if (id == R.id.reply) {
```
cketti: Removing this disables the functionality in the "Send…" submenu of the action bar.

PR #3744, opened 2018-11-23
Move logic out of Account class (1)

Valodim:
These are the first two commits of #3737, before converting to Kotlin. This already gets rid of 500 lines in Account, and should be reasonable to review as a first step.

### DIFF
--- a/app/core/src/main/java/com/fsck/k9/Account.java
+++ b/app/core/src/main/java/com/fsck/k9/Account.java
@@ -467,32 +467,6 @@ public class Account implements BaseAccount, StoreConfig {
         }
     }
 
-    private static int findNewAccountNumber(List<Integer> accountNumbers) {
-        int newAccountNumber = -1;
-        Collections.sort(accountNumbers);
-        for (int accountNumber : accountNumbers) {
-            if (accountNumber > newAccountNumber + 1) {
-                break;
-            }
-            newAccountNumber = accountNumber;
-        }
-        newAccountNumber++;
-        return newAccountNumber;
-    }
-
-    private static List<Integer> getExistingAccountNumbers(Preferences preferences) {
-        List<Account> accounts = preferences.getAccounts();
-        List<Integer> accountNumbers = new ArrayList<>(accounts.size());
-        for (Account a : accounts) {
-            accountNumbers.add(a.getAccountNumber());
-        }
-        return accountNumbers;
-    }
-    public static int generateAccountNumber(Preferences preferences) {
-        List<Integer> accountNumbers = getExistingAccountNumbers(preferences);
-        return findNewAccountNumber(accountNumbers);
-    }
-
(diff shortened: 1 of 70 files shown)


app/core/src/main/java/com/fsck/k9/Account.java
```diff
@@ -641,132 +523,8 @@ public void move(Preferences preferences, boolean moveUp) {
...
-            if (useCompression != null) {
-                editor.putBoolean(accountUuid + ".useCompression." + type, useCompression);
-            }
-        }
-        saveIdentities(preferences.getStorage(), editor);
-
-        editor.commit();
-
+    public synchronized void save() {
+        DI.get(AccountManager.class).save(this);
```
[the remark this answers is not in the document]
Valodim: Absolutely. This was a shortcut to reduce the amount of changes necessary in other places.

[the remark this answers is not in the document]
Valodim: #3737 had that, but the PR grew too messy. That's why I kept the method around until I could inline it more cleanly. That's done now :)

Valodim:
Ok, this should be good to review again. Note that I tried to make only necessary changes in some places, to avoid doing too much in one PR. I got too tangled up in #3737 with making things more kotlin-y, and tried to keep it to more straightforward refactoring here.

Valodim:
Well it's "move logic out of account class", not "apply cosmetics"! :suspect:

I think I broke LocalKeyStore along the way, too. At least I got some errors earlier when I tested. I'll look at it tomorrow.

app/core/src/main/java/com/fsck/k9/Preferences.java
```diff
@@ -190,4 +194,43 @@ public Storage getStorage() {
     private BackendManager getBackendManager() {
         return DI.get(BackendManager.class);
     }
+
+    public void saveAccount(Account account) {
+        StorageEditor editor = storage.edit();
+
+        if (!accounts.containsKey(account.getUuid())) {
```
morckx: Also new accounts are already added to `accounts` with their uuids here. This results in new accounts always having -1 as  accountNumber. 

Using `if (account.getAccountNumber() < 0) {` instead seems to work for me.

Valodim: Good catch! This previously checked for the uids in committed storage.

Hm. @cketti is there a reason why accountNumber is sometimes used instead of uuid, e.g. in [NewMailNotifications](https://github.com/k9mail/k-9/blob/c043eb0df4f79f450c46abaac089fcb68adb7992/app/core/src/main/java/com/fsck/k9/notification/NewMailNotifications.java#L107)?

cketti: In that particular case the account's UUID probably would have done the trick. In other cases we need a number. `Account.getId()` might have been a better name, though.

PR #3745, opened 2018-11-23
Add accounts header to the drawer

morckx:
Show an accounts header like in gmail, but including the unified inbox

Replaces #3741

### DIFF
--- a/app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
+++ b/app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
@@ -134,8 +134,8 @@ public class K9Drawer {
                             return false;
                         } else {
                             Account account = (Account) ((ProfileDrawerItem) profile).getTag();
-                            updateUserAccountsAndFolders(account);
                             Accounts.openRealAccount(account, parent);
+                            updateUserAccountsAndFolders(account);
                             return false;
                         }
                     }
@@ -194,6 +194,7 @@ public class K9Drawer {
         if (account == null) {
             selectUnifiedInbox();
         } else {
+            unifiedInboxSelected = false;
             accountHeader.setActiveProfile((account.getAccountNumber()+1) << DRAWER_ACCOUNT_SHIFT);
             ViewModelProvider viewModelProvider = ViewModelProviders.of(parent, new MessageListViewModelFactory());
             MessageListViewModel viewModel = viewModelProvider.get(MessageListViewModel.class);
@@ -250,8 +251,6 @@ public class K9Drawer {
 
         if (openedFolderDrawerId != -1) {
             drawer.setSelection(openedFolderDrawerId, false);
-        } else if (unifiedInboxSelected) {
-            selectUnifiedInbox();
         }
     }
 
--- a/app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
+++ b/app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
@@ -5,6 +5,7 @@ import android.arch.lifecycle.Observer;
 import android.arch.lifecycle.ViewModelProvider;
 import android.arch.lifecycle.ViewModelProviders;
 import android.graphics.Color;
+import android.graphics.PorterDuff;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
@@ -94,6 +95,7 @@ public class K9Drawer {
     private AccountHeader buildAccountHeader() {
         AccountHeaderBuilder headerBuilder = new AccountHeaderBuilder()
                 .withActivity(parent)
+                .withHeaderBackground(R.drawable.drawer_header_background)
                 .withTranslucentStatusBar(false);
 
         if (!K9.isHideSpecialAccounts()) {
@@ -207,6 +209,7 @@ public class K9Drawer {
         } else {
             unifiedInboxSelected = false;
             accountHeader.setActiveProfile((account.getAccountNumber()+1) << DRAWER_ACCOUNT_SHIFT);
+            accountHeader.getHeaderBackgroundView().setColorFilter(account.getChipColor(), PorterDuff.Mode.OVERLAY);
             ViewModelProvider viewModelProvider = ViewModelProviders.of(parent, new MessageListViewModelFactory());
             MessageListViewModel viewModel = viewModelProvider.get(MessageListViewModel.class);
             viewModel.getFolders(account).observe(parent, new Observer<List<Folder>>() {
@@ -288,6 +291,7 @@ public class K9Drawer {
         unifiedInboxSelected = true;
         openedFolderServerId = null;
         accountHeader.setActiveProfile(DRAWER_ID_UNIFIED_INBOX);
+        accountHeader.getHeaderBackgroundView().setColorFilter(0xFFFFFFFF, PorterDuff.Mode.MULTIPLY);
         clearUserFolders();
     }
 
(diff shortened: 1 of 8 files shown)


morckx:
A different icon for the unified inbox is still missing and it's a bit unusual to have the unified inbox in the header, but I think it is the best place for it.
An icon that would catch both birds could be: [fu-users](https://fontawesome.com/icons/users?style=solid)

Valodim:
A screenshot showing off how it looks would be very welcome for this kind of PR :)

morckx:
screenshot_2018-11-23-23-40-27-207_com fsck k9

morckx:
Maybr like this? What do you think? 
screenshot_2018-11-24-12-07-39-810_com fsck k9 debug

morckx:
Or like in gmail using Contacts.getPhotoUri and the old vetsion as fallback (also for non-unique photos) 

screenshot_2018-11-24-14-38-32-988_com fsck k9 debug

wiktor-k:
Wow, looks very nice, I would say it's a Gmail app on the screenshot on the first sight :) :+1:

wiktor-k:
Wow, looks very nice, I would say it's a Gmail app on the screenshot on the first sight :) :+1:

app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
```diff
@@ -48,9 +64,11 @@
     private int iconFolderResId;
 
     private final List<Long> userFolderDrawerIds = new ArrayList<>();
+    private final List<Long> userAccountDrawerIds = new ArrayList<>();
     private boolean unifiedInboxSelected;
     private String openedFolderServerId;
 
+    private final Preferences preferences = DI.get(Preferences.class);
```
cketti: Ideally `K9Drawer` would be more like a dumb view. Pass in the accounts, folders, and information which of them are selected. And when items are clicked on, notify observers via callback. That'll make for a clear separation between behavior and presentation.

app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
```diff
@@ -64,24 +82,68 @@ public K9Drawer(MessageList parent, Bundle savedInstanceState) {
...
+                            .colorRes(R.color.material_drawer_background).backgroundColor(Color.GRAY)
+                            .sizeDp(56).paddingDp(8))
+                    .withSetSelected(unifiedInboxSelected)
+                    .withIdentifier(DRAWER_ID_UNIFIED_INBOX)
+            );
+        }
 
-            headerItemCount += 2;
+        List <Account> accounts = preferences.getAccounts();
+        for (int i = preferences.getAccounts().size() - 1; i >= 0; i--) {
```
cketti: Why are you adding accounts in reverse order?

app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
```diff
@@ -64,24 +82,68 @@ public K9Drawer(MessageList parent, Bundle savedInstanceState) {
...
+                .withOnAccountHeaderListener(new AccountHeader.OnAccountHeaderListener() {
+                    @Override
+                    public boolean onProfileChanged(View view, IProfile profile, boolean currentProfile) {
+                        if (profile.getIdentifier() == DRAWER_ID_UNIFIED_INBOX) {
+                            parent.openUnifiedInbox();
+                            return false;
+                        } else {
+                            Account account = (Account) ((ProfileDrawerItem) profile).getTag();
+                            updateUserAccountsAndFolders(account);
+                            Accounts.openRealAccount(account, parent);
```
cketti: `Accounts` will be removed once we have the account switcher in `MessageList` (and all the functionality reachable from `Accounts` is accessible from elsewhere). So the functionality to switch accounts should live in `MessageList`.

app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
```diff
@@ -48,9 +64,11 @@
     private int iconFolderResId;
 
     private final List<Long> userFolderDrawerIds = new ArrayList<>();
+    private final List<Long> userAccountDrawerIds = new ArrayList<>();
```
cketti: This field is never used anywhere.

app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
```diff
@@ -64,24 +82,68 @@ public K9Drawer(MessageList parent, Bundle savedInstanceState) {
...
         addFooterItems();
 
         initializeFolderIcons();
     }
 
-    private void addHeaderItems() {
+    private AccountHeader buildAccountHeader() {
+        AccountHeaderBuilder headerBuilder = new AccountHeaderBuilder()
+                .withActivity(parent)
+                //.withTextColorRes(R.color.primary_text_secondary_when_activated_material)
```
cketti: Please remove unused code.

app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
morckx: leftover from drawer item implementation...

app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
morckx: I'm groping a bit in the dark here, it would be great if you could take over this, or we can wait until `Accounts` is settled.

app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
morckx: leftover...

app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
morckx: yes, sure

morckx:
Thanks for the detailed review!

> But during my tests selecting another account sometimes wouldn't update the drawer (the message list was updated, however). I haven't yet looked into what is causing this.

Yes, indeed, that's partially due to the leftovers. I'll fix that.

> Right now the code in K9Drawer is very tightly coupled to a lot of other parts of the app. I'd love for this to be cleaned up before merging the code. But if you don't have time or don't know what I'm looking for exactly, let me know and I'll do it myself.

I can roughly imagine what you're aiming at, but i still have to guess, It would be great if you could do that.

morckx:
The folder list is still not updated sometimes.

ByteHamster:
It would be nice to have a generic background for the accounts header. That background could then be tinted in the account color.

Yatse and Transportr use a stripe pattern. DavDroid uses a plain color. Gmail uses your header from Google+. To begin with, the plain account color could look great (maybe darkened a bit).

app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
morckx: done

morckx:
> It would be nice to have a generic background for the accounts header. That background could then be tinted in the account color.

> Yatse and Transportr use a stripe pattern. DavDroid uses a plain color. Gmail uses your header from Google+. To begin with, the plain account color could look great (maybe darkened a bit).

It's quite tricky to choose a background with enough contrast to text and account icons because we cannot control the chosen account colors (chosen as fallback for account icons if no photo is specified) and can only assume that the user has not chosen his background color. We'd need quite a vivid image or maybe the chosen account color darkened or brightened quite a lot, depending on light/dark mode.

ByteHamster:
Maybe that can be done in a follow-up PR. In my opinion, it would also look great if the highlight color in the sidebar was switched automatically.

morckx:
screenshot_2018-11-27-21-13-38-916_com fsck k9
screenshot_2018-11-27-21-13-30-736_com fsck k9

ByteHamster:
What do you think about moving the sidebar above the toolbar, @cketti @morckx? Now that there is a header, there is less space for the folder list.

Valodim:
Yeah, I'd move it above the toolbar. Let's move that to follow-up PRs. Stuff I gather from this issue:

* [ ] change background image to tint with account color
* [ ] extract logic from K9Drawer
* [ ] move drawer above toolbar

Is the "folder bar sometimes not updated" still a thing?

morckx:
>  change background image to tint with account color

that's already done

> Is the "folder bar sometimes not updated" still a thing?

no, it's also fixed

Valodim:
perfect :)

PR #3762, opened 2018-11-28
(WIP) Change minSdkVersion to 19, use Storage Access Framework for attachments/images 

GoneUp:
Introduces saving of attachments with the Storage access framework. It also bumps the MinSdkVersion to 19 (Kitkat) to remove the legacy file saving code (Mainly FileBrowser) for attachments and setting export. With this change the only usage of the storage permission should be the use of external mail database. 

Current behaviour of this PR for saving attachments:
On Kitkat: Saving with simple/long press is handled via `ACTION_CREATE_DOCUMENT`
On >Lollipop: Saving with simple press via `ACTION_OPEN_DOCUMENT_TREE`, On long press with `ACTION_CREATE_DOCUMENT`

ACTION_OPEN_DOCUMENT_TREE will mimic the old behaviour, the user sets a initial directory (or over settings) and the file is saved there. If a file with the name already exists a unique name with -x is created. 

Current behaviour of this PR for saving images:
- On Kitkat: Legacy save
- On Lollipop: Saving via `ACTION_OPEN_DOCUMENT_TREE` on the default attachment location

What is left to do?
- Use`ACTION_CREATE_DOCUMENT`  for saving a image (that would provide a single way to save a file)
- Find a good replacment for the usage of the DownloadManager service. The api does only support paths and no content URIs. 
 I added a toast for now, but I don't feel like this a good way to this. Maybe a own Notification that displays the filename and support the opening of the file? 
- Code and commit log clean-up

Relevant issues: #3752 #2110

### DIFF
--- a/app/core/src/main/java/com/fsck/k9/K9.java
+++ b/app/core/src/main/java/com/fsck/k9/K9.java
@@ -916,6 +916,9 @@ public class K9 {
         hideHostnameWhenConnecting = state;
     }
 
+    /**
+     * @return Either a document uri on devices > Lollipop or a path on older devices
+     */
     public static String getAttachmentDefaultPath() {
         return attachmentDefaultPath;
     }
--- a/app/core/src/main/java/com/fsck/k9/helper/FileHelper.java
+++ b/app/core/src/main/java/com/fsck/k9/helper/FileHelper.java
@@ -11,6 +11,8 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Locale;
 
 import timber.log.Timber;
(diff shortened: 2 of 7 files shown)


GoneUp:
@cketti Thoughts on this? What would you like from a UI perspective?

GoneUp:
Okay, great. I gonna change the PR.

Valodim:
(slightly related and just making sure this isn't lost: For "external sd card" storage, no permission is necessary, so we don't have to worry about that)

wiktor-k:
This generally looks good!

I'm getting an exception when saving an attachment though (after the Storage Framework save dialog):

```
java.lang.RuntimeException: An error occurred while executing doInBackground()
	at android.os.AsyncTask$3.done(AsyncTask.java:353)
	at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
	at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
	at java.util.concurrent.FutureTask.run(FutureTask.java:271)
	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:245)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
	at java.lang.Thread.run(Thread.java:764)
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.net.Uri.toString()' on a null object reference
	at com.fsck.k9.ui.messageview.DownloadImageTask.doInBackground(DownloadImageTask.java:51)
	at com.fsck.k9.ui.messageview.DownloadImageTask.doInBackground(DownloadImageTask.java:32)
	at android.os.AsyncTask$2.call(AsyncTask.java:333)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	... 4 more
```

PR #3779, opened 2018-12-02
Pick a contrasting foreground for drawer account chips

williamvds:
Uses the greatest contrasting color from the Material Drawer's background colors
This makes the chip icon more visible against lighter backgrounds 

Some screenshots:  
dark background
light background

### DIFF
--- a/app/core/src/main/java/com/fsck/k9/helper/Utility.java
+++ b/app/core/src/main/java/com/fsck/k9/helper/Utility.java
@@ -9,10 +9,12 @@ import java.util.regex.Pattern;
 
 import android.content.Context;
 import android.database.Cursor;
+import android.graphics.Color;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.os.Handler;
 import android.os.Looper;
+import android.support.annotation.ColorInt;
 import android.text.Editable;
 import android.text.TextUtils;
 import android.widget.EditText;
@@ -322,5 +324,16 @@ public class Utility {
             sMainThreadHandler = new Handler(Looper.getMainLooper());
         }
         return sMainThreadHandler;
+    }
+
+    /**
+     * @return contrast ratio between the given colors
+     * @see <a href="https://www.w3.org/TR/WCAG20/#glossary">W3C's definition of contrast ratio</a>
+     */
+    public static float contrastRatio(@ColorInt int a, @ColorInt int b) {
+        float aLuminance = Color.luminance(a) + 0.05f,
+              bLuminance = Color.luminance(b) + 0.05f;
+
+        return Math.max(aLuminance, bLuminance) / Math.min(aLuminance, bLuminance);
     }
 }
--- a/app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
+++ b/app/ui/src/main/java/com/fsck/k9/ui/K9Drawer.java
@@ -9,6 +9,7 @@ import android.graphics.PorterDuff;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.support.v4.content.res.ResourcesCompat;
 import android.support.v4.widget.DrawerLayout;
 import android.util.TypedValue;
 import android.view.View;
@@ -18,6 +19,7 @@ import com.fsck.k9.K9;
 import com.fsck.k9.Preferences;
 import com.fsck.k9.activity.MessageList;
 import com.fsck.k9.helper.Contacts;
+import com.fsck.k9.helper.Utility;
 import com.fsck.k9.mailstore.Folder;
 import com.fsck.k9.ui.folders.FolderNameFormatter;
 import com.fsck.k9.ui.messagelist.MessageListViewModel;
@@ -131,8 +133,19 @@ public class K9Drawer {
                 photoUris.add(photoUri);
                 pdi.withIcon(photoUri);
             } else {
+                int fg = ResourcesCompat.getColor(parent.getResources(), R.color.material_drawer_background, null),
+                    bg = account.getChipColor();
+
+                float[] hsv = new float[3];
+                Color.colorToHSV(bg, hsv);
+                while (Utility.contrastRatio(fg, bg) < 3) {
+                    hsv[2] *= .9f;
+                    bg = Color.HSVToColor(hsv);
+                }
+
                 pdi.withIcon(new IconicsDrawable(parent, FontAwesome.Icon.faw_user_alt)
-                        .colorRes(R.color.material_drawer_background).backgroundColor(account.getChipColor())
+                        .color(fg)
+                        .backgroundColor(bg)
                         .sizeDp(56).paddingDp(14));
             }
             headerBuilder.addProfiles(pdi);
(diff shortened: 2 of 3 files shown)


ByteHamster:
I think this could get too colorful. In my opinion, an app should have only very few different colors on one screen. What do you think about using just black and white foregrounds, depending on how light the account color is?

cketti:
I don't know. It looks weird to have a mixed set of icons with light and dark foregrounds.

Currently the account header background is bothering me a lot more. I'd welcome experiments using different approaches. Maybe with different gradients or entirely without gradient. Using very light or dark versions of the account color as background so there's more contrast to the account icon. Or using the account color as header background and icon foreground color and (depending on the theme?) a light or dark background color for the account icons.

williamvds:
Darkening the color seems to be effective, though I'm not sure about it aesthetically

accountchipcontrastdarkening

(Background gradient disabled to act as a reference for the original color)

Algorithm:
```java
float[] hsv = new float[3];
Color.colorToHSV(bg, hsv);
while (Utility.contrastRatio(fg, bg) < 3) {
    hsv[2] *= .9f;
    bg = Color.HSVToColor(hsv);
}
```

Valodim:
might be my screen, but that text is pretty hard to read

wiktor-k:
> might be my screen, but that text is pretty hard to read

I think that's what they were referring to in "though I'm not sure about it aesthetically" :)

williamvds:
Yep, the background's gradient was disabled so the original color is obvious. Normally the gradient makes it more legible  
I'm not sure about it aesthetically because the method of picking the  darker color is rather dumb, and could use ones that don't work well  
There's probably a way to find nicer darker versions of a color, then the one that passes a minimum contrast could be used

morckx:
If you just want a darker tinted background you can decrease the grey values in [drawer_header_background.xml](https://github.com/k9mail/k-9/blob/e9d3c14ff8d5399f80daf45291a440c3183bebd0/app/ui/src/main/res/drawable/drawer_header_background.xml). There you can also get rid of the gradient. 

I'm on the one hand quite happy with the gradient, but, on the other hand, often think that it's a bit too much. Maybe a darker tint without gradient would be a good compromise.

ByteHamster:
I created some mock-ups. The account itself is the same on every image: white foreground and 50% black background. Personally, I think a plain background looks boring. Additionally, it feels too "strong", meaning too much area of the same color.

In every case, the contrast of the account image seems to be good.

## Plain color
  

## Multi-Gradient
 

## Material paper

Valodim:
The last one looks really nice, I like that :+1:

ByteHamster:
I just added another one with plain background and a big, slightly darker email icon.

cketti:
> The last one looks really nice, I like that

I like the "Material paper" version, too. Does it work well with K-9 Mail's default account colors? Or should we change what we use as default account colors? They're probably still Holo-based anyway.

ByteHamster:
A problem I see with my mockup is that the actual account color is no longer used. It's just using the dark versions. Is that a good idea?

morckx:
> A problem I see with my mockup is that the actual account color is no longer used. It's just using the dark versions. Is that a good idea?

My reasons for using a darkened version of the account colors were that other colors seem quite arbitrary and potentially misleading, and that when there is a photo for the account, the account color will be nowhere present in the drawer.

Can't you tint your material paper background with the account color just like it's done now, by only replacing the background drawable?

(Another idea to have at least the account color present was to have a colored circle around the photo, but that seemed neither ideal nor easy to implement.)

ByteHamster:
> Can't you tint your material paper background with the account color just like it's done now, by only replacing the background drawable?

Yes but in order to support yellow, we need to use a quite dark background image. To provide contrast for the account chip, we also need to darken the chip color. Therefore, the original (not darkened) color is never used.

Yellow is the only color that creates problems. What about just providing a discrete number of colors, without including yellow (or just some darker shade)? Google calendar also just supports a limited number of colors. Then we can use the actual color for the background and account chips without darkening.

williamvds:
Apologies for the lack of clarity - I wasn't suggesting that the background gradient be removed, it was simply removed when I took that screenshot so that you could compare the original account color with the one used in the chip

But while the conversation has derailed into the drawer background, might I suggest this:

drawerbackgroundproposal

Pros:
- uses the original account color
- text & account icon are legible
  - the gradient should probably end a bit darker

Cons:
- still looks a bit simple?
  - could be used in combination with one of @ByteHamster's suggestions

ByteHamster:
I like the horizontal gradient because it allows to use the actual account color. I think that it's a good step forward. Replacing the background with paper or mail icon can be done after fixing the account icons.

Did you add the gradient to the image or was it possible to enable a gradient using the drawer itself?

morckx:
> Did you add the gradient to the image or was it possible to enable a gradient using the drawer itself?

(If you mean me)  The gradient is defined  [here](https://github.com/k9mail/k-9/blob/e9d3c14ff8d5399f80daf45291a440c3183bebd0/app/ui/src/main/res/drawable/drawer_header_background.xml)

and this is the whole commit: https://github.com/k9mail/k-9/commit/e9d3c14ff8d5399f80daf45291a440c3183bebd0

ByteHamster:
I wondered if the gradient in @williamvds screenshot is added by simply removing the angle or by using something that's included in MaterialDrawer

ByteHamster:
I just checked out your branch. Looks nice. 

I think adding the material paper background AND the gradient might be too much. The gradient itself looks interesting enough and it works with all colors. :+1:

wiktor-k:
I'd also vote for "Material Paper", it looks very professional. Gradients are OK, not bad, but Material Paper is IMHO the best :+1:

morckx:
I also like the "Material Paper" background best. @ByteHamster: Do you have a source for it that we could use (with a suitable license)?

ByteHamster:
@morckx Unfortunately not. I just took the image for the mockup from Google images :P If everyone agrees that we want to use that background, I can create a similar one in Inkscape

ByteHamster:
Looks like Android <24 does not support gradients in paths. Therefore, adding the material paper can not be done with xml vector drawables. @cketti would you also accept a background in png (+svg source) format?

cketti:
> @cketti would you also accept a background in png (+svg source) format?

Sure, that's fine.

ByteHamster:
I added the material paper background in PR #3851

wiktor-k:
:thinking: should this be closed because of #3851?

ByteHamster:
The original issue this fixes (white avatar on yellow chip background) is not yet fixed. #3851 is only about the background of the drawer.

I talked to cketty a few days ago and he told me that he'd prefer restricting the possible account colors on the settings screen instead of allowing to select yellow and then display another color. 

Maybe we could use a color picker similar to the one Google Calendar uses. Other suggestion would be just a bunch of colored circles like this:
https://github.com/kristiyanP/colorpicker

wiktor-k:
Oh, right, now I see the thread was hijacked to discuss two things at once, thanks for the clarification @ByteHamster!

The colorpicker you linked to looks nice.

cketti:
With PR #3851 merged the original issue shouldn't be a problem anymore.

PR #3783, opened 2018-12-02
Switch to Toolbar

ByteHamster:
This allows to show the sidebar above the ActionBar, giving more space for folders.

### DIFF
--- a/app/ui/src/main/res/layout/about.xml
+++ b/app/ui/src/main/res/layout/about.xml
@@ -1,5 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
-<WebView xmlns:android="http://schemas.android.com/apk/res/android"
-    android:id="@+id/about_view"
-    android:layout_width="match_parent"
-    android:layout_height="match_parent" />
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+              android:orientation="vertical"
+              android:layout_height="fill_parent"
+              android:layout_width="fill_parent">
+
+    <include layout="@layout/toolbar" />
+
+    <WebView
+            android:id="@+id/about_view"
+            android:layout_width="match_parent"
+            android:layout_height="0dp"
+            android:layout_weight="1"/>
+</LinearLayout>
--- a/app/ui/src/main/java/com/fsck/k9/activity/AccountList.java
+++ b/app/ui/src/main/java/com/fsck/k9/activity/AccountList.java
@@ -39,7 +39,7 @@ public abstract class AccountList extends K9ListActivity implements OnItemClickL
 
         setResult(RESULT_CANCELED);
 
-        setContentView(R.layout.account_list);
+        setLayout(R.layout.account_list);
 
         ListView listView = getListView();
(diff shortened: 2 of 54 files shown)


cketti:
This doesn't play well with the changes from #3771. `K9PreferenceActivity` needs a theme with action bar.

ByteHamster:
Do you have a suggestion what should be done instead?

ByteHamster:
@cketti what do you think about adding new themes without an ActionBar and then adding a method that converts the R constant from the normal theme to the one without ActionBar? Then the theme without ActionBar can be applied only to MessageList.

This probably also reduces the diff but I am not sure if it is "clean" enough for you to convert themes.

cketti:
I like that you converted all of the activities to use toolbars. I think we should keep that, then add "legacy themes" with an action bar. Since they will only be used by `K9PreferenceActivity` subclasses we can omit most of the attributes of the regular K-9 themes.

To avoid having to call `setSupportActionBar()` in all activities I suggest we add a method `K9Activity.setLayout(int layoutResId)` that calls `setContentView()` and then `setSupportActionBar()`.

app/ui/src/main/res/layout/about.xml
```diff
@@ -1,5 +1,19 @@
...
-<WebView xmlns:android="http://schemas.android.com/apk/res/android"
-    android:id="@+id/about_view"
-    android:layout_width="match_parent"
-    android:layout_height="match_parent" />
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+              android:orientation="vertical"
+              android:layout_height="fill_parent"
+              android:layout_width="fill_parent">
+
+    <android.support.v7.widget.Toolbar
```
williamvds: The same toolbar element appears to be used in every layout that needs one  
How about pulling this out into its own layout, then using `<include layout="@layout/toolbar.xml"/>` here instead?

cketti:
There are lots of subclasses of `K9Activity` that haven't been updated yet and are now missing an actionbar.

Using a custom view with a toolbar doesn't seem to automatically disable the regular actionbar title. Adding `actionBar.setDisplayShowTitleEnabled(false);` in `MessageList` fixed the issue for me.

ByteHamster:
I hope that I have finally found all child activities. Sorry for the noise.

I verified that by clicking each activity in the list and searching for "setContentView". Do you think that is enough?

screenshot_20181204_221630

cketti:
I believe you found all activities. However, the following layouts have small display issues:
- `account_setup_account_type`: padding on root container which also affects toolbar
- `account_setup_options`: layout contains two toolbars
- `crypto_key_transfer`: padding on root container which also affects toolbar
- `upgrade_databases`: gravity attribute on root container which also affects toolbar

And toolbar doesn't seem to support elevation on API levels prior to 21. That's an issue for another pull request, though.

cketti:
Thanks!

ByteHamster:
Sorry for ignoring your review. I was busy with university

cketti:
No worries. You already did most of the work. I just added some minor cleanup. Thanks again!

PR #3814, opened 2018-12-11
Fix subject being overwritten with `null`

### LINKED ISSUE #3811: Subject is not shown when sending new mail
### Expected behavior
The subject I wrote when I created a new mail appears well in the sent mail.

### Actual behavior
The subject I wrote when composing new mail is not sent well

### Steps to reproduce
1. Move compose menu.
2. Write new subject and message. (ex. test subject / test message)
3. Send email
4. result ((no subject) / test message)
image

### Environment
K-9 Mail version: 5.600

Android version: 8.0.0

Account type (IMAP, POP3, ##WebDAV/Exchange):

wiktor-k:
Not using crypto-provider passes `null` as `plaintextSubject`. `saveDraft` already has protection against overwriting subject with `null`. This change adds the same check to `sendMessage`.

Fixes #3811.

There was an alternative solution of [returning subject for all messages](https://github.com/k9mail/k-9/issues/3811#issuecomment-445824005), but that would also add the `X_SUBJECT_DECRYPTED` flag (making it useless it if was set for all messages).

The fix is consistent with checks in other methods.

* Contains a reference to the issue that it fixes. :heavy_check_mark: 

@morckx, would you be so kind to test this fix? Thank you in advance!

### DIFF
--- a/app/core/src/main/java/com/fsck/k9/controller/MessagingController.java
+++ b/app/core/src/main/java/com/fsck/k9/controller/MessagingController.java
@@ -1457,7 +1457,9 @@ public class MessagingController {
             localFolder.appendMessages(Collections.singletonList(message));
             LocalMessage localMessage = localFolder.getMessage(message.getUid());
             localMessage.setFlag(Flag.X_DOWNLOADED_FULL, true);
-            localMessage.setCachedDecryptedSubject(plaintextSubject);
+            if (plaintextSubject != null) {
+                localMessage.setCachedDecryptedSubject(plaintextSubject);
+            }
             localFolder.close();
             sendPendingMessages(account, listener);
         } catch (Exception e) {


wiktor-k:
Is it just me or builds fail with some mysterious reason?

```
FAILURE: Build failed with an exception.
* What went wrong:
A problem occurred configuring project ':app:k9mail'.
> Failed to notify project evaluation listener.
   > java.lang.reflect.InvocationTargetException (no error message)
```

morckx:
> @morckx, would you be so kind to test this fix? Thank you in advance!

Your solution works for me with and without encryption. Thank you!

Valodim:
Thanks for the fix! The code was indeed confusing there, `plaintextSubject` didn't imply that this is indeed a "specially cached, non-encrypted" subject obviously enough.

PR #3817, opened 2018-12-12
Fix consistent crashes when using dark theme

wiktor-k:
Commit 212f36170 moved dark theme's parent to `Theme.AppCompat.NoActionBar` but the theme file for v21 was not updated. This caused `IllegalStateException` to be thrown when constructingK9Activities when dark theme has been selected.

The fix corrects the parent value.

* Contains a reference to the issue that it fixes. Didn't create issue for that.


wiktor-k:
A week of using the app without any changes to master will be a good test :) Thanks for merging!

PR #3829, opened 2018-12-16
Redesign About activity

williamvds:
Replaces the WebView with a normal layout  
I've done my best to adjust existing strings for:
  - debug_version_fmt (-> version)
  - app_license
  - about_title_fmt (-> about_title)

Roughly follows the design from [#1859](https://github.com/k9mail/k-9/issues/1859#issuecomment-270029045)

aboutredesign

I thought there was an issue specifically for this, but it seems it was just my imagination

### Added stuff
- icons (Google's `code` and `description`)
- link to source code
- link to contributors list
- libraries list their licenses

### TODO
- localize about_libraries
- make an `openURL` method for links?
- final XML cleanup

### DIFF
--- a/app/ui/src/main/java/com/fsck/k9/ui/settings/AboutActivity.kt
+++ b/app/ui/src/main/java/com/fsck/k9/ui/settings/AboutActivity.kt
@@ -36,6 +36,10 @@ class AboutActivity : K9Activity() {
             startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(getString(R.string.app_license_url))))
         }
 
+        findViewById<View>(R.id.source).setOnClickListener {
+            startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(getString(R.string.app_source_url))))
+        }
+
         webView = findViewById(R.id.about_view)
 
         val aboutHtml = buildHtml()
--- a/app/ui/src/main/java/com/fsck/k9/ui/settings/AboutActivity.kt
+++ b/app/ui/src/main/java/com/fsck/k9/ui/settings/AboutActivity.kt
@@ -1,17 +1,22 @@
 package com.fsck.k9.ui.settings
 
+import java.util.Calendar
+
 import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager
+import android.net.Uri
 import android.os.Bundle
 import android.view.Menu
 import android.view.MenuItem
+import android.view.View
 import android.webkit.WebView
+import android.widget.TextView
 import com.fsck.k9.activity.K9Activity
 import com.fsck.k9.ui.R
 import de.cketti.library.changelog.ChangeLog
+
 import timber.log.Timber
-import java.util.Calendar
 
 class AboutActivity : K9Activity() {
     private lateinit var webView: WebView
@@ -21,6 +26,16 @@ class AboutActivity : K9Activity() {
         setLayout(R.layout.about)
 
         supportActionBar!!.setDisplayHomeAsUpEnabled(true)
+
+        findViewById<TextView>(R.id.version).text = getVersionNumber()
+
+        val year = Integer.toString(Calendar.getInstance().get(Calendar.YEAR))
+        findViewById<TextView>(R.id.copyright).text = getString(R.string.app_copyright_fmt, year, year)
+
+        findViewById<View>(R.id.licenseLayout).setOnClickListener {
+            startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(getString(R.string.app_license_url))))
+        }
+
         webView = findViewById(R.id.about_view)
 
         val aboutHtml = buildHtml()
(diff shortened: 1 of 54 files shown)


wiktor-k:
I've tested this on a real device, looks and works very well :+1:

app/ui/src/main/java/com/fsck/k9/ui/settings/AboutActivity.kt
```diff
@@ -100,29 +102,42 @@ class AboutActivity : K9Activity() {
...
+                Library("TokenAutoComplete", "https://github.com/splitwise/TokenAutoComplete/", "Apache License, Version 2.0"))
 
         fun start(context: Context) {
             val intent = Intent(context, AboutActivity::class.java)
             context.startActivity(intent)
         }
     }
 }
+
+public class LibraryFragment : Fragment() {
```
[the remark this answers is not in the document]
williamvds: Converted to `RecyclerView` in force-pushed a1195f5
I used fragments because I knew they were a thing. I'm new to Android development - thanks for pointing me to that

app/ui/src/main/java/com/fsck/k9/ui/settings/AboutActivity.kt
```diff
@@ -21,10 +31,45 @@ class AboutActivity : K9Activity() {
...
-        webView = findViewById(R.id.about_view)
 
-        val aboutHtml = buildHtml()
-        webView.loadDataWithBaseURL("file:///android_res/drawable/", aboutHtml, "text/html", "utf-8", null)
+        findViewById<TextView>(R.id.version).text = getVersionNumber()
+
+        val year = Integer.toString(Calendar.getInstance().get(Calendar.YEAR))
+        findViewById<TextView>(R.id.copyright).text = getString(R.string.app_copyright_fmt, year, year)
+
+        findViewById<View>(R.id.authorsLayout).setOnClickListener {
```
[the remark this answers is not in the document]
williamvds: Bindings are now used as of 1853964

app/ui/src/main/res/values/strings.xml
```diff
@@ -83,7 +88,7 @@ Please submit bug reports, contribute new features and ask questions at
     <string name="delete_messages_label">Delete Emails</string>
     <string name="delete_messages_desc">Allows this application to delete your Emails.</string>
 
-    <string name="about_title_fmt">About <xliff:g id="app_name">%s</xliff:g></string>
```
wiktor-k: Is there a reason `%s` was removed? I guess some people may want to release a customized version of K-9 mail with a different name that is easy to adjust.

app/ui/src/main/res/values-zh-rTW/strings.xml
```diff
@@ -148,7 +148,7 @@
   <string name="special_mailbox_name_archive_fmt"><xliff:g id="folder">%s</xliff:g>（歸檔）</string>
   <string name="special_mailbox_name_spam_fmt"><xliff:g id="folder">%s</xliff:g>（垃圾郵件）</string>
   <string name="send_failure_subject">部分郵件沒有成功寄送</string>
-  <string name="debug_version_fmt">版本：<xliff:g id="version">%s</xliff:g></string>
+  <string name="version">版本：<xliff:g id="version">%s</xliff:g></string>
```
wiktor-k: Interpolation symbols are not removed here as they are in english version resulting in "Version: %s" (where "Version" is localized and `%s` is printed verbatim).

I'm not entirely sure how should these strings be handled w.r.t. Transifex.

wiktor-k:
Maybe the "Version" field should also be clickable opening the built-in change-log? I know the changelog is in the overflow menu but both Authors and License are clickable, that'd make it a consistent behavior.

Valodim:
Might as well get rid of the overflow menu, then?

app/ui/src/main/res/values/strings.xml
williamvds: A few other strings appear to use "K-9 Mail" instead of string formatting
I thought I'd try to fix all of those in one go later (eg using [xml entities](https://stackoverflow.com/a/39870268))
Would it be preferable to keep them as formatted strings for now?

app/ui/src/main/res/values-zh-rTW/strings.xml
williamvds: Thanks for spotting, fixed in b9724b4
I've never used Transifex before, though [the docs](https://docs.transifex.com/formats/android-xml/) suggest the files can be re-imported

williamvds:
Version field now opens changelog, overflow menu removed (2473a57)

wiktor-k:
This looks very good on my device :+1:. I'll take a better look at the source tomorrow to get this merged as soon as possible.

Thank you!

app/ui/src/main/res/values/strings.xml
wiktor-k: If you're planning to address it later I think it's OK to use "K-9 Mail" now, at least this will be consistent :)

app/ui/src/main/java/com/fsck/k9/ui/settings/AboutActivity.kt
```diff
@@ -1,42 +1,75 @@
...
 import de.cketti.library.changelog.ChangeLog
+import kotlinx.android.synthetic.main.about.*
+import kotlinx.android.synthetic.main.about_library.view.*
+
 import timber.log.Timber
-import java.util.Calendar
 
-class AboutActivity : K9Activity() {
-    private lateinit var webView: WebView
+private data class Library(val name: String, val URL: String, val license: String = "")
```
wiktor-k: Hmm... as far as I can see all libraries have licenses so I think we can drop the default (` = ""`).

williamvds: Done in 7b15649

wiktor-k:
Merged, thanks for your work @williamvds! It finally looks modern now! :)

PR #3851, opened 2018-12-27
Material paper drawer background

ByteHamster:
The background even works well with yellow. The account chip still looks strange when selecting yellow: restricting the account colors to not allow yellow will be done in another PR.


ByteHamster:
By the way, the image is not repeatable. It looks too symmetric if it's repeatable, in my opinion.

cketti:
@ByteHamster: Can you please create the PNGs for the different density buckets we support for regular drawables (mdpi tp xxhdpi).

ByteHamster:
I think it would look good to have the sidebar menu go below the status bar like in most Google apps. What do you think about that?

cketti:
> I think it would look good to have the sidebar menu go below the status bar like in most Google apps. What do you think about that?

:+1:

Dafnik:
No most google apps don't do this. It was a thing when the first version of Material design was released.

ByteHamster:
Hi @Dafnik, not sure if we are talking about the same thing. The Google apps I use (Calendar, Gmail, Play store, Play console, Translate) all display the sidebar below the status bar. Only Google Maps does not display the sidebar below the status bar.

screenshot_20181229-234822_google_play_store-picsay

Dafnik:
Oh I'm sorry, I thought you talked about the toolbar. My fault 👍

images/drawer_header_background_generate.sh
```diff
@@ -0,0 +1,11 @@
...
+
+function generate() {
+    inkscape "drawer_header_background.svg" -export-area-page --without-gui -e "../app/ui/src/main/res/drawable-$1/drawer_header_background.png" -w $2 -h $3
+}
+
+generate mdpi     384 216
+generate hdpi     576 324
+generate xhdpi    768 432
+generate xxhdpi  1152 648
+#generate xxxhdpi 1728 972
```
wiktor-k: Why is this commented out? Not used?

ByteHamster: cketty told me to support mdpi tp xxhdpi. Just in case we want to support xxxhdpi sometimes, there sizes are already here.

wiktor-k: Err, sounds legit. And it's not commented-out because the file generated would be big or for what reason?

cketti: xxxhdpi is ridiculously dense and really should only be used for launcher icons because "some app launchers display your app icon as much as 25% larger than what's called for by the device's density bucket". (https://developer.android.com/training/multiscreen/screendensities)

See also https://developer.android.com/guide/topics/resources/providing-resources

cketti:
:+1:

PR #3928, opened 2019-02-27
Koin upgrade

pueffl:
Migration to Koin 1.0.2


pueffl:
Not based correctly on master

PR #4010, opened 2019-04-06
Flat icons in message list

ByteHamster:
Replaced icons with flat versions. Icon shadows are not used in material design.

Before/After:
image a

Before/After:


cketti:
Nice change. 

The icons should probably be a bit smaller and lighter (in the light theme; respectively a bit darker in the dark theme). I like how the paperclip icon in the Gmail app matches the size of the date text.
I don't like the angled shape of the reply/forward icons. Maybe use the shape of the reply action icon?

ByteHamster:
Updated the icons.

image

Before / After:

cketti:
Very nice :100: Thanks!

PR #4102, opened 2019-07-10
* STAR to flag messages now uses the color of the respective account.

robin523790:
- Instead of light blue which was difficult to see with dark theme.
- Uses the same shape and color as AccountsAdapter (list of all accounts).

Accounts overview (unchanged) --- has stars in account colors
01_accountlist_unchanged

Message list (before changes) --- Hard to see which messages are flagged.
02_messagelist_before_changes

Message lists (after changes):
03_messagelist-all-accounts_changed
04_messagelist-blue-account
05_messagelist-green-account
06_messagelist-red-account

Sorry, no screenshot of the MessageHeader view. There also the star was light blue, now is in the same color&shape as in the lists.

### DIFF
--- a/app/ui/src/main/java/com/fsck/k9/fragment/MessageListAdapter.java
+++ b/app/ui/src/main/java/com/fsck/k9/fragment/MessageListAdapter.java
@@ -209,7 +209,11 @@ public class MessageListAdapter extends CursorAdapter {
             holder.selected.setChecked(selected);
         }
         if (fragment.stars) {
-            holder.flagged.setChecked(flagged);
+            final int iAttr = flagged ? R.attr.iconActionFlag : R.attr.iconActionUnflag;
+            final TypedArray attrArray = holder.flagged.getContext().getTheme().obtainStyledAttributes(new int[] { iAttr });
+            final int attributeResourceId = attrArray.getResourceId(0, 0);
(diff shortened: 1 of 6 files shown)


cketti:
I don't like the stars using the account color. I think we should do it similar to how Gmail does it. Use a star outline for unstarred messages and a filled star for starred messages. We can use the Material Design icons `star` and `star_border` for this.

robin523790:
If we use `star` and `star_border`, I'd have to be theme-aware. Like this:

gray_darktheme  gray_lighttheme


If we do what Gmail does, the icon would always be tinted yellow:

yellow_darktheme  yellow_lighttheme

I've implemented both. Tell me what you prefer, and I'll update the PR accordingly.

_In either case, the accounts overview would remain unchanged, still using the account color. (Cp. first screenshot above)_

cketti:
I think it's hard to find a shade of yellow that works equally well for both the light and the dark theme. So I'd like to go with a theme-aware version. Your first version looks nice. Let's go with that.

robin523790:
Done. Thank you (and everyone else who contributed) for your efforts with this project!!

cketti:
Thanks for updating the pull request. However, I'd prefer to keep the old mechanism of using a `CheckBox` with a style that references a selector drawable to do all of the work.

cketti:
See #4124

PR #4104, opened 2019-07-11
refactor / Decouple MessageListAdapter

bhm:
This should decouple the MessageListAdapter from MessageListFragment.

This is  a basic refactor to decouple MessageListFragment further, and move MessageListFragment to MVVM down the line


bhm:
Should be on my remote first thing.

PR #4165, opened 2019-08-15
Update strings.xml

viliam-durina:
Fix a typo in Slovak translation.


ByteHamster:
Translations are managed on Transifex: https://www.transifex.com/projects/p/k9mail/

Changes in the source files will be overwritten when pulling translations from Transifex.

viliam-durina:
Thanks, fixed the translation on Transfiex

PR #4178, opened 2019-08-31
Folder manager

ByteHamster:
See #3750

- Changed activity title to 'Manage folders'
- Added icons to FolderList
- Removed color chip

Before/After:

### DIFF
--- a/app/ui/src/main/res/layout/folder_list_item.xml
+++ b/app/ui/src/main/res/layout/folder_list_item.xml
@@ -1,28 +1,32 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    android:id="@+id/folder_list_item_layout"
-    android:layout_width="match_parent"
-    android:layout_height="wrap_content"
-    android:gravity="center_vertical"
-    android:minHeight="?android:attr/listPreferredItemHeight"
-    android:orientation="horizontal">
+        xmlns:android="http://schemas.android.com/apk/res/android"
+        xmlns:tools="http://schemas.android.com/tools"
+        android:id="@+id/folder_list_item_layout"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:gravity="center_vertical"
+        android:minHeight="56dp"
+        android:orientation="horizontal">
 
-    <View
-        android:id="@+id/chip"
-        android:layout_width="8dp"
-        android:layout_height="match_parent"
-        android:layout_marginEnd="8dp"
-        android:background="#ff336699" />
+    <ImageView
+            android:id="@+id/folder_icon"
+            android:layout_width="24dp"
+            android:layout_height="24dp"
+            android:layout_marginStart="16dp"
+            android:layout_marginEnd="32dp"
+            android:src="@drawable/ic_drafts_folder_light"/>
 
     <TextView
-        android:id="@+id/folder_name"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:padding="8dp"
-        android:textAppearance="?android:attr/textAppearanceLarge"
-        android:textColor="?android:attr/textColorPrimary"
-        tools:text="Folder name" />
+            android:id="@+id/folder_name"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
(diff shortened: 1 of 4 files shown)


app/ui/src/main/java/com/fsck/k9/ui/folders/FolderIconLoader.kt
```diff
@@ -0,0 +1,60 @@
+package com.fsck.k9.ui.folders
+
+import android.content.res.Resources
+import android.util.TypedValue
+import com.fsck.k9.mail.Folder
+import com.fsck.k9.mailstore.FolderType
+import com.fsck.k9.ui.R
+
+class FolderIconLoader(private val theme: Resources.Theme) {
```
cketti: This class isn't loading icons when its public methods are used. `FolderIconProvider` would probably be a better name.

app/ui/src/main/java/com/fsck/k9/ui/folders/FolderIconLoader.kt
```diff
@@ -0,0 +1,60 @@
...
+    private var iconFolderInboxResId: Int = 0
```
cketti: All of these can be read-only properties (`val`).

app/ui/src/main/java/com/fsck/k9/ui/folders/FolderIconLoader.kt
```diff
@@ -0,0 +1,60 @@
...
+    private var iconFolderOutboxResId: Int = 0
+    private var iconFolderSentResId: Int = 0
+    private var iconFolderTrashResId: Int = 0
+    private var iconFolderDraftsResId: Int = 0
+    private var iconFolderArchiveResId: Int = 0
+    private var iconFolderSpamResId: Int = 0
+    public var iconFolderResId: Int = 0
```
cketti: The `public` modifier can be omitted in Kotlin because it's the default.

app/ui/src/main/java/com/fsck/k9/ui/folders/FolderIconLoader.kt
```diff
@@ -0,0 +1,60 @@
...
+        FolderType.SENT -> iconFolderSentResId
+        FolderType.TRASH -> iconFolderTrashResId
+        FolderType.DRAFTS -> iconFolderDraftsResId
+        FolderType.ARCHIVE -> iconFolderArchiveResId
+        FolderType.SPAM -> iconFolderSpamResId
+        else -> iconFolderResId
+    }
+
+    public fun getFolderIcon(type: Folder.FolderType): Int = when (type) {
+        Folder.FolderType.INBOX -> iconFolderInboxResId
```
cketti: I suggest using an import of `import com.fsck.k9.mail.Folder.FolderType as LegacyFolderType` and replacing all occurrences of `Folder.FolderType` in this file with `LegacyFolderType`. I find this makes the code a little bit easier to read.

app/ui/src/main/res/layout/folder_list_item.xml
```diff
@@ -7,21 +7,22 @@
     android:layout_height="wrap_content"
     android:gravity="center_vertical"
     android:minHeight="?android:attr/listPreferredItemHeight"
-    android:orientation="horizontal">
+    android:orientation="horizontal"
+    android:padding="8dp">
```
cketti: The padding moves the folder icon too far to the right. You'll have to reduce the margin for the `ImageView` in order for the folder names to line up with the toolbar title.

ByteHamster:
Updated PR with the requested changes. When adding icons to the "Move message to folder" screen, too, the management and moving screen will look very similar. @cketti Do you think it could help to show a settings icon next to each entry on the "Manage folders" screen? I think displaying with full opacity looks a bit too strong but I am not sure if the light icon could be misread as disabled action.

cketti:
I'd leave out the settings icon. I'm sure we'll add other functionality to this screen later so it'll look different from the "choose folder" screen. For now it's fine if they look very similar.

ByteHamster:
Copied over your suggestion.

For the next PR adding icons to the "Move to folder" activity: do you prefer to copy the adapter and layout or can I re-use the one from the manage folders screen?

cketti:
> For the next PR adding icons to the "Move to folder" activity: do you prefer to copy the adapter and layout or can I re-use the one from the manage folders screen?

I think they'll show slightly different data soon. Having different copies of the adapters and layouts will make things easier in the future.

cketti:
Thanks :+1:

PR #4411, opened 2019-12-27
add background color to adaptive icons

Bubu:
This adds a desaturated version of K9-Mails envelope color as the
adaptive icon background color.

This should look a lot better than a plain white background.

Also slightly enlarge the icon shape.

image

### DIFF
--- a/app/k9mail/src/main/AndroidManifest.xml
+++ b/app/k9mail/src/main/AndroidManifest.xml
@@ -48,7 +48,6 @@
         android:allowTaskReparenting="false"
         android:usesCleartextTraffic="true"
         android:icon="@drawable/ic_launcher"
-        android:roundIcon="@drawable/ic_launcher_round"
         android:label="@string/app_name"
         android:theme="@style/Theme.K9.Startup"
         android:resizeableActivity="true"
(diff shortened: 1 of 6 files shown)


ByteHamster:
Most apps seem to use white, so this would not be consistent. Obviously, everyone feels different about design. To be honest, I personally don't think that it actually looks better. I have neutral feelings about the purple debug version: I don't think it really looks better but also not worse. Anyway, the red release version now looks quite pink with this change. In my opinion, the red background could be a bit more desaturated.

Bubu:
In my experience a white background is the exception. But that obviously depends on the apps you use. (And it very rarely looks good, but that's ofc subjective.)

I actually agree that the release version could be a bit more desaturated. It's not easy to find a sweet spot as it tends to look slightly different on each device.

cketti:
> I actually agree that the release version could be a bit more desaturated. It's not easy to find a sweet spot as it tends to look slightly different on each device.

Let's try a bit more desaturated version then.

Bubu:
`#FFCCD6` corresponds to HSV 348,20,100. (where 20 is the saturation).
We could try `#FFDAE1` which is 15% saturation.

Oh wow, github has a color preview :mind_blown:
image

How to proceed? Does someone else want to play around with the colors? Do you want a new side-by-side screenshort with the changed color?

cketti:
Update the pull request and I'll include it in the next beta build. If it doesn't work out we can always switch back to a white background.

Bubu:
Done

cketti:
Thanks

jpopovitch:
Can this be re-done as a configurable option?   I think it's relevant to say that not everyone wants a white background, yet now everyone is forced to have a pink background.

PR #4700, opened 2020-04-26
Optional password export

hannesa2:
image

Because https://github.com/k9mail/k-9/pull/4696 was declined on less reflected doubts, maybe we can optional export the passwords

### DIFF
--- a/app/core/src/main/java/com/fsck/k9/preferences/SettingsExporter.kt
+++ b/app/core/src/main/java/com/fsck/k9/preferences/SettingsExporter.kt
@@ -29,10 +29,10 @@ class SettingsExporter(
     private val folderRepositoryManager: FolderRepositoryManager
 ) {
     @Throws(SettingsImportExportException::class)
-    fun exportToUri(includeGlobals: Boolean, accountUuids: Set<String>, uri: Uri) {
+    fun exportToUri(includeGlobals: Boolean, accountUuids: Set<String>, uri: Uri, withPassword: Boolean) {
         try {
             contentResolver.openOutputStream(uri)!!.use { outputStream ->
-                exportPreferences(outputStream, includeGlobals, accountUuids)
+                exportPreferences(outputStream, includeGlobals, accountUuids, withPassword)
             }
         } catch (e: Exception) {
             throw SettingsImportExportException(e)
@@ -40,7 +40,7 @@ class SettingsExporter(
     }
 
     @Throws(SettingsImportExportException::class)
-    fun exportPreferences(outputStream: OutputStream, includeGlobals: Boolean, accountUuids: Set<String>) {
+    fun exportPreferences(outputStream: OutputStream, includeGlobals: Boolean, accountUuids: Set<String>, withPassword: Boolean) {
         try {
             val serializer = Xml.newSerializer()
             serializer.setOutput(outputStream, "UTF-8")
@@ -68,7 +68,7 @@ class SettingsExporter(
             serializer.startTag(null, ACCOUNTS_ELEMENT)
             for (accountUuid in accountUuids) {
                 val account = preferences.getAccount(accountUuid)
-                writeAccount(serializer, account, prefs)
+                writeAccount(serializer, account, prefs, withPassword)
             }
             serializer.endTag(null, ACCOUNTS_ELEMENT)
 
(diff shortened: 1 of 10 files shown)


hannesa2:
But before this must be solved https://github.com/k9mail/k-9/issues/4698 (I know it's maybe related to https://github.com/k9mail/k-9/issues/3691)

When https://github.com/k9mail/k-9/issues/4698 or https://github.com/k9mail/k-9/issues/3691 is solved (not by me) I can finish my tests here

hannesa2:
image

First of all, you added code style to git 👍 

Sorry, but the outcome with ktLint is not good. 👎  
Ok, you force an organized import order, but both function 

image#

changes nothing **at all**.

I would recommend to remove this check in ktLint, or change codestyle to make it work automatically

cketti:
Currently IntelliJ IDEA/Android Studio can't be configured to define the import order for Kotlin files. AFAIK this will be fixed soon. In the meantime you can use the `ktlintFormat` task to reformat the files, see also https://github.com/k9mail/k-9/wiki/BuildingK9#checkfix-code-style

Thanks for including the screenshot of the modified export screen. I don't quite like the idea of simply including a "Passwords" entry in the list. The option to include passwords when exporting accounts feels more like a modifier to the selection, not like something that belongs in the list of objects to export. 
I'd also like to include some warning text that clearly states that this will include passwords in clear text, so users need to be careful where they save the resulting file.

I currently don't know how a good UI that satisfies these points could look like. 
@jancborchardt: Do you maybe have a suggestion?

hannesa2:
Now, the user knows about the consequence

image

About the UI: you can change it later, when you see a better place. Maybe a separator is fine ?

hannesa2:
The password import works, but a folder sync is missing and only `Outbox` is shown.
image

After sync folders, it works properly.

@cketti I need your advise. Please can you point me to the place and what's to do then, to do this automatic after import

cketti:
Enabling an account via `AccountActivator` will trigger the necessary actions:

https://github.com/k9mail/k-9/blob/664e444d791c58f0ea8104908883334fa58e8d77/app/ui/src/main/java/com/fsck/k9/ui/settings/import/SettingsImportViewModel.kt#L203

https://github.com/k9mail/k-9/blob/664e444d791c58f0ea8104908883334fa58e8d77/app/ui/src/main/java/com/fsck/k9/ui/settings/import/AccountActivator.kt#L19-L29

jancborchardt:
> Now, the user knows about the consequence
> 
> image
> 
> About the UI: you can change it later, when you see a better place. Maybe a separator is fine ?

In addition to this wording, I would say this should be unchecked by default. What do you think?

hannesa2:
> In addition to this wording, I would say this should be unchecked by default. What do you think?

After fix build after project restructure, I'm a little bit tired about maintain this PR.

I see it as a nice feature to prevent from password-hell after device change. 
@jancborchardt The user (especially me) want to get rid of password enter, and your suggestion incapacitated the user again. 
Why this maximum prevent user from well working stuff ?

hannesa2:
@cketti  Any change to get this merged ?
A rename, changed default.... could be made later on

cketti:
I was thinking more along the lines of something like this:

k9mail_mockup_settings_export_with_passwords

Yes, the option would be unchecked by default. 
Yes, this looks scary. 
Yes, this doesn't make things easier for the average user. 

All of that is intentional. Exporting plain text passwords is dangerous. The only reason I'm even considering this is because someone else is willing to do the work and it's much more work to add a way to support exporting passwords in a secure way (that probably means encrypting all of them using a master password). This way at least users that are knowledgeable enough and accept the risks have the option of exporting passwords.

Why all the fuss? I believe that it is our job as developers to keep the user's data safe. The average user is not a tech expert and doesn't have the knowledge required to make a proper risk assessment. And they shouldn't have to. It is our job. 
Sometimes that also means not adding a feature that would greatly improve the user experience if it would also increase the risk beyond acceptable levels. Of course the acceptable level of risk is highly subjective. With K-9 Mail we rather err on the side of safety. But in my experience this is the norm rather than the exception in the free software community.

Merging an incomplete feature and then finishing things later is totally acceptable in some situations. Adding something dangerous and then adding the safety switch later (maybe? worst case is I'll have to do the work myself; on a feature I don't think is a great idea in the first place) is not something I'm keen on doing.

hannesa2:
Sorry, all is just to make thing not/bad working

hannesa2:
There is nothing incomplete, simply added like other switches before.
If someone has additional wishes, he can do it.

PR #4767, opened 2020-05-15
Lay groundwork for optional in/out server discovery

livmackintosh:
Components to update (including tests):
- [x]  ConnectionSettingsDiscovery interface
- [x] SrvServiceDiscovery
- [x] ProviderXmlDiscovery*
- [x] ThunderbirdDiscovery

* This appears to be designed such that there is one SMTP and one IMAP server for each email provider. The current implementation hasn't really been changed to support multiple SMTP ot multiple IMAP servers.

Addresses: https://github.com/k9mail/k-9/issues/4759

### DIFF
--- a/app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/ConnectionSettingsDiscovery.kt
+++ b/app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/ConnectionSettingsDiscovery.kt
@@ -1,5 +1,25 @@
 package com.fsck.k9.autodiscovery
 
+import com.fsck.k9.mail.AuthType
+import com.fsck.k9.mail.ConnectionSecurity
+
 interface ConnectionSettingsDiscovery {
-    fun discover(email: String): ConnectionSettings?
+    fun discover(email: String, target: DiscoveryTarget): DiscoveryResults?
 }
+
+enum class DiscoveryTarget(val outgoing: Boolean, val incoming: Boolean) {
+    OUTGOING(true, false),
+    INCOMING(false, true),
+    INCOMING_AND_OUTGOING(true, true)
+}
+
+data class DiscoveryResults(val incoming: List<DiscoveredServerSettings>, val outgoing: List<DiscoveredServerSettings>)
+
+data class DiscoveredServerSettings(
+    val protocol: String,
+    val host: String,
+    val port: Int,
+    val security: ConnectionSecurity,
+    val authType: AuthType?,
+    val username: String?
+)
--- a/app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/providersxml/ProvidersXmlDiscovery.kt
+++ b/app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/providersxml/ProvidersXmlDiscovery.kt
@@ -1,8 +1,10 @@
 package com.fsck.k9.autodiscovery.providersxml
 
 import android.content.res.XmlResourceParser
-import com.fsck.k9.autodiscovery.ConnectionSettings
 import com.fsck.k9.autodiscovery.ConnectionSettingsDiscovery
+import com.fsck.k9.autodiscovery.DiscoveredServerSettings
+import com.fsck.k9.autodiscovery.DiscoveryResults
+import com.fsck.k9.autodiscovery.DiscoveryTarget
 import com.fsck.k9.backend.BackendManager
 import com.fsck.k9.helper.EmailHelper
 import com.fsck.k9.helper.UrlEncodingHelper
@@ -16,11 +18,7 @@ class ProvidersXmlDiscovery(
     private val xmlProvider: ProvidersXmlProvider
 ) : ConnectionSettingsDiscovery {
 
-    override fun discover(email: String): ConnectionSettings? {
-        return discover(email, outgoing = true, incoming = true)
-    }
-
-    private fun discover(email: String, outgoing: Boolean, incoming: Boolean): ConnectionSettings? {
+    override fun discover(email: String, target: DiscoveryTarget): DiscoveryResults? {
         val password = ""
 
         val user = EmailHelper.getLocalPartFromEmailAddress(email) ?: return null
@@ -38,7 +36,16 @@ class ProvidersXmlDiscovery(
             val incomingUri = with(URI(provider.incomingUriTemplate)) {
                 URI(scheme, "$incomingUserUrlEncoded:$password", host, port, null, null, null).toString()
             }
-            val incomingSettings = backendManager.decodeStoreUri(incomingUri)
+            val incomingSettings = backendManager.decodeStoreUri(incomingUri).let {
+                listOf(DiscoveredServerSettings(
+                    it.type,
+                    it.host,
+                    it.port,
+                    it.connectionSecurity,
+                    it.authenticationType,
+                    email
+                ))
+            }
 
             val outgoingUserUrlEncoded = provider.outgoingUsernameTemplate
                     ?.replace("\$email", emailUrlEncoded)
@@ -48,9 +55,18 @@ class ProvidersXmlDiscovery(
             val outgoingUri = with(URI(provider.outgoingUriTemplate)) {
                 URI(scheme, outgoingUserInfo, host, port, null, null, null).toString()
             }
-            val outgoingSettings = backendManager.decodeTransportUri(outgoingUri)
+            val outgoingSettings = backendManager.decodeTransportUri(outgoingUri).let {
+                listOf(DiscoveredServerSettings(
+                    it.type,
+                    it.host,
+                    it.port,
+                    it.connectionSecurity,
+                    it.authenticationType,
+                    email
+                ))
+            }
 
-            return ConnectionSettings(incomingSettings, outgoingSettings)
+            return DiscoveryResults(incomingSettings, outgoingSettings)
         } catch (use: URISyntaxException) {
             return null
         }
--- a/app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/providersxml/ProvidersXmlDiscovery.kt
+++ b/app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/providersxml/ProvidersXmlDiscovery.kt
@@ -17,6 +17,10 @@ class ProvidersXmlDiscovery(
 ) : ConnectionSettingsDiscovery {
 
     override fun discover(email: String): ConnectionSettings? {
+        return discover(email, outgoing = true, incoming = true)
+    }
+
+    private fun discover(email: String, outgoing: Boolean, incoming: Boolean): ConnectionSettings? {
         val password = ""
 
         val user = EmailHelper.getLocalPartFromEmailAddress(email) ?: return null
--- a/app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/providersxml/ProvidersXmlDiscovery.kt
+++ b/app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/providersxml/ProvidersXmlDiscovery.kt
@@ -43,7 +43,7 @@ class ProvidersXmlDiscovery(
                     it.port,
                     it.connectionSecurity,
                     it.authenticationType,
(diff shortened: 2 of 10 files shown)


app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/providersxml/ProvidersXmlDiscovery.kt
```diff
@@ -17,6 +17,10 @@ class ProvidersXmlDiscovery(
 ) : ConnectionSettingsDiscovery {
 
     override fun discover(email: String): ConnectionSettings? {
+        return discover(email, outgoing = true, incoming = true)
```
cketti, line 20: I suggest replacing the parameters `incoming` and `outgoing` with an enum class, so that the invalid state where both arguments are `false` will be prevented by the type system.

app/ui/src/main/java/com/fsck/k9/activity/setup/AccountSetupBasics.java
```diff
@@ -263,6 +266,34 @@ private void finishAutoSetup(ConnectionSettings connectionSettings) {
         AccountSetupCheckSettings.actionCheckSettings(this, mAccount, CheckDirection.INCOMING);
     }
 
+    private ConnectionSettings providersXmlDiscoveryDiscover(String email, DiscoveryTarget discoveryTarget) {
```
livmackintosh, line 269: I don't really want this to exist for long in it's current form because it is making assumptions about the implementation of ProvidersXmlDiscovery.discover() but it's a stepping stone.

app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/ConnectionSettingsDiscovery.kt
```diff
@@ -1,5 +1,25 @@
 package com.fsck.k9.autodiscovery
 
+import com.fsck.k9.mail.AuthType
+import com.fsck.k9.mail.ConnectionSecurity
+
 interface ConnectionSettingsDiscovery {
-    fun discover(email: String): ConnectionSettings?
+    fun discover(email: String, target: DiscoveryTarget = DiscoveryTarget.INCOMING_AND_OUTGOING): DiscoveryResults?
```
livmackintosh, line 7: This is pandering to the test a little bit. I should remove this and specify target in test(s).

app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/srvrecords/SrvServiceDiscovery.kt
```diff
@@ -1,49 +1,42 @@
...
-                outgoingService.host,
-                outgoingService.port,
-                outgoingService.security,
-                AuthType.PLAIN,
-                email,
-                null,
-                null
-            )
-        )
+        // val pickMailService = compareBy<MailService> { it.priority }.thenByDescending { it.security }
```
livmackintosh, line 17: Something similar to this can be used in AccountSetupBasic?

app/autodiscovery/src/test/java/com/fsck/k9/autodiscovery/providersxml/ProvidersXmlDiscoveryTest.kt
```diff
@@ -28,23 +28,24 @@ class ProvidersXmlDiscoveryTest : RobolectricTest() {
         val connectionSettings = providersXmlDiscovery.discover("user@gmail.com")
 
         assertThat(connectionSettings).isNotNull()
-        with(connectionSettings!!.incoming) {
+        with(connectionSettings!!.incoming!!.first()) {
```
livmackintosh, line 31: Tests need updating to reflect type change. This is assuming there is only one result, we could just assert this.

app/autodiscovery/src/test/java/com/fsck/k9/autodiscovery/srvrecords/SrvServiceDiscoveryTest.kt
```diff
@@ -17,9 +22,10 @@ class SrvServiceDiscoveryTest : RobolectricTest() {
         val srvServiceDiscovery = SrvServiceDiscovery(srvResolver)
         val result = srvServiceDiscovery.discover("test@example.com")
 
-        assertNull(result)
+        assertEquals(DiscoveryResults(listOf(), listOf()), result)
     }
 
+    @Ignore
```
livmackintosh, line 28: Need to update these tests.

app/autodiscovery/src/test/java/com/fsck/k9/autodiscovery/thunderbird/ThunderbirdAutoconfigTest.kt
```diff
@@ -41,12 +41,12 @@ class ThunderbirdAutoconfigTest : RobolectricTest() {
         assertThat(connectionSettings).isNotNull()
         assertThat(connectionSettings!!.incoming).isNotNull()
         assertThat(connectionSettings.outgoing).isNotNull()
-        with(connectionSettings.incoming) {
-            assertThat(host).isEqualTo("imap.googlemail.com")
-            assertThat(port).isEqualTo(993)
-            assertThat(username).isEqualTo("test@metacode.biz")
+        with(connectionSettings.incoming?.first()) {
```
livmackintosh, line 44: Same as ProviderXmlDiscovery

app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/thunderbird/ThunderbirdAutoconfigParser.kt
```diff
@@ -16,37 +16,32 @@ import org.xmlpull.v1.XmlPullParserFactory
...
-        var incoming: ServerSettings? = null
-        var outgoing: ServerSettings? = null
-
+    fun parseSettings(stream: InputStream, email: String): DiscoveryResults? {
         val factory = XmlPullParserFactory.newInstance()
         val xpp = factory.newPullParser()
 
         xpp.setInput(InputStreamReader(stream))
 
+        val incomingServers = mutableListOf<DiscoveredServerSettings>()
```
livmackintosh, line 25: Not so keen on this very imperative implementation - it feels quite "noisy". Would much prefer a functional approach but I can revisit that near the end if feasible.

livmackintosh:
I think this is more or less complete in a pinch but there are a few areas that I've highlighted that may benefit from some improvements. I'd be interested to get your take.

app/ui/src/main/java/com/fsck/k9/activity/setup/AccountSetupBasics.java
cketti: I don't think we should try to modify the current setup UI to work with the plan proposed in #4721. It's better to create a new Activity. 
Since I expect it to be quite a while before the new setup experience is complete, I think it'd be good to support both UIs at the same time. We'll use the old UI for releases until the new one is ready. I can set this up next week.
But that also means this code will be around for a while. We should make sure it works like it did before, as best as we can.

app/ui/src/main/java/com/fsck/k9/activity/setup/AccountSetupBasics.java
```diff
@@ -263,6 +266,34 @@ private void finishAutoSetup(ConnectionSettings connectionSettings) {
...
+        DiscoveryResults discoveryResults = providersXmlDiscovery.discover(email, DiscoveryTarget.INCOMING_AND_OUTGOING);
+        if (discoveryResults == null || (discoveryResults.getIncoming().size() != 1 || discoveryResults.getOutgoing().size() != 1)) {
```
cketti, line 271: IIRC the old implementation used the first result. So this should only return `null` when there's no incoming or outgoing server at all.

app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/ConnectionSettingsDiscovery.kt
```diff
@@ -1,5 +1,25 @@
 package com.fsck.k9.autodiscovery
 
+import com.fsck.k9.mail.AuthType
+import com.fsck.k9.mail.ConnectionSecurity
+
 interface ConnectionSettingsDiscovery {
-    fun discover(email: String): ConnectionSettings?
+    fun discover(email: String, target: DiscoveryTarget): DiscoveryResults?
```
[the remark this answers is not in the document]
livmackintosh: I suppose there is a loose connection between null relating to a failure in parsing the email or getting garbage response but if we do care about such handling we probably should use something equivalent to a Maybe/Either type but for now I'll remove it.

app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/ConnectionSettingsDiscovery.kt
```diff
@@ -1,5 +1,25 @@
...
+    OUTGOING(true, false),
+    INCOMING(false, true),
+    INCOMING_AND_OUTGOING(true, true)
 }
+
+data class DiscoveryResults(val incoming: List<DiscoveredServerSettings>, val outgoing: List<DiscoveredServerSettings>)
+
+data class DiscoveredServerSettings(
+    val protocol: String,
+    val host: String?,
```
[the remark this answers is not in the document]
livmackintosh: ThunderbirdAutoconfigParser returns a type with nullable host value.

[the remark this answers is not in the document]
livmackintosh: In practice, I'm not sure exactly where it would be null though.

app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/ConnectionSettingsDiscovery.kt
```diff
@@ -1,5 +1,25 @@
...
+    val port: Int,
+    val security: ConnectionSecurity?,
```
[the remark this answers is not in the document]
livmackintosh: Same reason as host

app/autodiscovery/src/test/java/com/fsck/k9/autodiscovery/srvrecords/SrvServiceDiscoveryTest.kt
```diff
@@ -79,10 +85,56 @@ class SrvServiceDiscoveryTest : RobolectricTest() {
...
+                host = "smtp.example.com", port = 465, srvType = SrvType.SUBMISSIONS,
+                security = ConnectionSecurity.SSL_TLS_REQUIRED, priority = 0)
+        ))
+        val srvServiceDiscovery = SrvServiceDiscovery(srvResolver)
+        val result = srvServiceDiscovery.discover("test@example.com", DiscoveryTarget.OUTGOING)
+        verify(srvResolver).lookup("example.com", SrvType.SUBMISSIONS)
+        verify(srvResolver).lookup("example.com", SrvType.SUBMISSION)
+        verifyNoMoreInteractions(srvResolver)
+        assertEquals(1, result!!.outgoing.size)
+        assertEquals(listOf<DiscoveredServerSettings>(), result.incoming)
```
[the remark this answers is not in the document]
livmackintosh: :grimacing:

app/autodiscovery/src/main/java/com/fsck/k9/autodiscovery/ConnectionSettingsDiscovery.kt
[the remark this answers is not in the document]
cketti: We can't use values where `host` or `security` are `null`. Can you please change this. 
For now it's okay to use the `!!`-method in `ThunderbirdAutoconfigParser` (it's already used for `port`). We have to make another pass over that class at some point anyway.

livmackintosh:
@cketti I'll do that now. thanks

livmackintosh:
@cketti should be good now - was a bit of pain :sweat_smile:

cketti:
Thanks

PR #4897, opened 2020-07-30
Message list: hide footer view if empty

starshipcoder:
Hide the footer view in message list if empty

Before
device-2020-07-30-104008

After
device-2020-07-30-103825

### DIFF
--- a/app/ui/legacy/src/main/java/com/fsck/k9/fragment/MessageListFragment.java
+++ b/app/ui/legacy/src/main/java/com/fsck/k9/fragment/MessageListFragment.java
@@ -1106,14 +1106,18 @@ public class MessageListFragment extends Fragment implements OnItemClickListener
             return;
         }
 
-        FooterViewHolder holder = (FooterViewHolder) footerView.getTag();
-
-        if (text != null) {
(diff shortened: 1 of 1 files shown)


szaimen:
I mean great if the footer is gone when the message list is empty but why was it needed at all in the first place?
(I would remove it completely)

cketti:
Thanks a lot for your contribution :heart: 

Sorry for the delay in getting this merged.

@szaimen: The footer view is used as a button to fetch more messages from the server ("Load up to X more").

szaimen:
> @szaimen: The footer view is used as a button to fetch more messages from the server ("Load up to X more").

Ah okay, thanks for the clarification!

Could the footer also get hidden if all mails in a specific folder are loaded?
(I can of course open a new issue for that, if you want)
image

cketti:
@szaimen: This change will hide the footer whenever it would have been displayed as an empty view. So your case is covered by this as well.

szaimen:
> @szaimen: This change will hide the footer whenever it would have been displayed as an empty view. So your case is covered by this as well.

Ah okay great! Thanks for for this!

PR #4909, opened 2020-08-13
Closes #4127 by changing sha-1 to sha-512

### LINKED ISSUE #4127: Display Sha512 Fingerprints for Server-Certs
### Expected behavior
If my mailserver uses a self-signed cert, k9 should display the sha512 fingerprint of the cert

### Actual behavior
I only see sha1 fingerpint

### Environment
K-9 Mail version: 5.600
Android version: 8.1

GNUDimarik:

### DIFF
--- a/app/ui/legacy/src/main/java/com/fsck/k9/activity/setup/AccountSetupCheckSettings.java
+++ b/app/ui/legacy/src/main/java/com/fsck/k9/activity/setup/AccountSetupCheckSettings.java
@@ -166,9 +166,9 @@ public class AccountSetupCheckSettings extends K9Activity implements OnClickList
 
                 mProgressBar.setIndeterminate(false);
                 StringBuilder chainInfo = new StringBuilder(100);
-                MessageDigest sha1 = null;
+                MessageDigest sha512 = null;
                 try {
-                    sha1 = MessageDigest.getInstance("SHA-1");
+                    sha512 = MessageDigest.getInstance("SHA-512");
                 } catch (NoSuchAlgorithmException e) {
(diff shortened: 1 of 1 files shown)


GNUDimarik:
I had no chance to test it because it could not to find JMAP server for my emails ... I trying to find out why ...

GNUDimarik:
Thanks for update.
Once i able to test it i'll add the PR.
Btw how do you currently test it during development? What do you use with
jmap support?

пт, 14 авг. 2020 г., 16:49 cketti <notifications@github.com>:

cketti:
This code is unrelated to JMAP. It is only used for IMAP, POP3, SMTP and maybe WebDAV servers.

GNUDimarik:
Right. But i can't loging to the app at all... Finding out ...

cketti:
Make sure to launch `k9mail.app.k9mail` in Android Studio, **not** `k9mail.app.k9mail-jmap`.

GNUDimarik:
Thanks.
Now it's better. Will continue working then

GNUDimarik:
> Thanks for looking to contribute to K-9 Mail ❤️
> 
> Browsers show the SHA-1 and SHA-256 fingerprint of certificates. We should also do that.
> 
> I don't mind if we show the SHA-512 fingerprint as well. But only showing the SHA-512 value (when nobody else seems to do that) is not an option. So this PR should be changed to show the SHA-512 value in addition to the SHA-1 value. Ideally, you'd also add the SHA-256 value.

So you mean to put 3 different hashes there ideally. RIght? PLease confirm

PR #5122, opened 2021-02-05
convert unified inbox from a virtual account to a virtual folder

toppk:
This is only in the UI.  This initial implementation leaves unified inbox implemented as a SearchAccount, but otherwise ignored as
an account, and is treated as a special drawer entry by the UI (not as a folder).

This addresses:  https://github.com/k9mail/k-9/issues/5020

I do think that we may want to rethink usability around default folder, as people may want to prefer unified/normal inbox as default.  I do think that the behavior implemented here to switch to normal inbox when changing profiles makes sense.  There is already a ticket around default folders here:   https://github.com/k9mail/k-9/issues/4471

Just submitting this for early feedback.  I want to get unread message counts display working, and perhaps understand why there are all these createUnifiedInboxAccount() calls all over the place.

### DIFF
--- a/app/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
+++ b/app/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
@@ -377,6 +377,11 @@ open class MessageList :
             search!!.addAllowedFolder(folderId)
         }
 
+        // edgecase: someone disabled unified view, while looking at unified view
+        if ((search != null) && (search!!.id == SearchAccount.UNIFIED_INBOX) && !K9.isShowUnifiedInbox) {
+            search = null
+        }
+
         if (search == null) {
             val accountUuid = intent.getStringExtra("account")
             if (accountUuid != null) {
@@ -1421,6 +1426,8 @@ open class MessageList :
         val drawer = drawer ?: return
         when {
             singleFolderMode -> drawer.selectFolder(search!!.folderIds[0])
+            // deselect the view since we're still looking at unified mailbox
+            search!!.id == SearchAccount.UNIFIED_INBOX && !K9.isShowUnifiedInbox -> drawer.deselect()
             search!!.id == SearchAccount.UNIFIED_INBOX -> drawer.selectUnifiedInbox()
             else -> drawer.deselect()
         }
--- a/app/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
+++ b/app/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
@@ -394,11 +394,10 @@ open class MessageList :
                 search!!.addAccountUuid(accountUuid)
                 search!!.addAllowedFolder(folderId)
             } else {
+                account = preferences.defaultAccount
                 if (K9.isShowUnifiedInbox) {
-                    account = null
                     search = SearchAccount.createUnifiedInboxAccount().relatedSearch
                 } else {
-                    account = preferences.defaultAccount
                     search = LocalSearch()
                     search!!.addAccountUuid(account!!.uuid)
                     val folderId = defaultFolderProvider.getDefaultFolder(account!!)
@@ -522,8 +521,6 @@ open class MessageList :
     }
 
     fun openUnifiedInbox() {
-        account = null
-        drawer!!.selectUnifiedInbox()
         actionDisplaySearch(this, SearchAccount.createUnifiedInboxAccount().relatedSearch, false, false)
     }
 
@@ -1406,9 +1403,7 @@ open class MessageList :
         this.search = search
         singleFolderMode = false
 
-        if (search!!.searchAllAccounts()) {
-            account = null
-        } else {
+        if (!search!!.searchAllAccounts()) {
             val accountUuids = search.accountUuids
             if (accountUuids.size == 1) {
                 account = preferences.getAccount(accountUuids[0])
--- a/app/ui/legacy/src/main/java/com/fsck/k9/ui/K9Drawer.kt
+++ b/app/ui/legacy/src/main/java/com/fsck/k9/ui/K9Drawer.kt
@@ -2,7 +2,6 @@ package com.fsck.k9.ui
 
 import android.content.Context
 import android.content.res.Resources
-import android.graphics.Color
 import android.graphics.PorterDuff
 import android.net.Uri
 import android.os.Bundle
@@ -35,6 +34,7 @@ import com.mikepenz.materialdrawer.AccountHeaderBuilder
 import com.mikepenz.materialdrawer.Drawer
 import com.mikepenz.materialdrawer.Drawer.OnDrawerItemClickListener
 import com.mikepenz.materialdrawer.DrawerBuilder
+import com.mikepenz.materialdrawer.model.DividerDrawerItem
 import com.mikepenz.materialdrawer.model.PrimaryDrawerItem
 import com.mikepenz.materialdrawer.model.ProfileDrawerItem
 import com.mikepenz.materialdrawer.model.interfaces.IDrawerItem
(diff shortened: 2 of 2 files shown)


app/ui/legacy/src/main/java/com/fsck/k9/activity/AccountList.java
```diff
@@ -70,25 +70,13 @@ public void onItemClick(AdapterView<?> parent, View view, int position, long id)
     public void populateListView(List<Account> realAccounts) {
         List<BaseAccount> accounts = new ArrayList<>();
 
-        if (displaySpecialAccounts() && !K9.isHideSpecialAccounts()) {
```
cketti, line 73: Hm, I didn't consider screens based on `AccountList` when creating #5020. For now we should probably keep the current behavior.

app/core/src/main/java/com/fsck/k9/K9.kt
```diff
@@ -227,7 +227,7 @@ object K9 : EarlyInit {
     var isUseVolumeKeysForListNavigation = false
 
     @JvmStatic
-    var isHideSpecialAccounts = false
+    var isHideUnifiedInbox = false
```
cketti, line 230: If we're renaming this anyway, let's make it `isShowUnifiedInbox`. If find that makes for more readable code.

app/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
```diff
@@ -1406,9 +1403,7 @@ open class MessageList :
         this.search = search
         singleFolderMode = false
 
-        if (search!!.searchAllAccounts()) {
-            account = null
-        } else {
+        if (! search!!.searchAllAccounts()) {
```
cketti, line 1406: Nit: I'm surprised ktlint doesn't flag this. There shouldn't be a space between `!` and `search…`.

app/ui/legacy/src/main/res/values-bg/strings.xml
```diff
@@ -653,7 +653,7 @@ K-9 Mail е мощен, безплатен имейл клиент за Андр
   <string name="volume_navigation_title">Навигация с клавиш за звука</string>
   <string name="volume_navigation_message">В преглед на съобщението</string>
   <string name="volume_navigation_list">В списъка със съобщения</string>
-  <string name="hide_special_accounts_title">Скрий обща входяща кутия</string>
+  <string name="hide_unified_inbox_title">Скрий обща входяща кутия</string>
```
cketti, line 656: Renaming string keys is tricky. For translations we use [Transifex](https://www.transifex.com/k-9/k9mail). Changing the key of a string is equivalent to removing a string and adding a new one. So all translations for the old key on Transifex will be dropped and the new string will be initially untranslated.
The next time we pull translations from Transifex all changes in `values-*/strings.xml` will be overwritten. I could pull in Translations from Transifex, update with your changes, and then push to Transifex and hope there haven't been any changes in the meantime. But I generally try to avoid that.

I think it's best to live with the outdated name for now. Maybe add a TODO comment to the main strings file stating that the key should be renamed. Once we have enough cases like this we can do a mass rename, giving prior notice to translators.

cketti (changes requested):
I think this is the way to go :+1:

Have you tried adding a divider line between "Unified Inbox" and regular folders? I think that separation might be helpful. But I haven't tried adding it myself. I might not like the idea anymore once I see it :)

app/ui/legacy/src/main/java/com/fsck/k9/activity/AccountList.java
toppk: For clarification, you want to keep the unified Inbox account in accounts list object, you just want me to find another way of keeping it out of the UI.  I think I can do that.

cketti: No, just don't touch these screens. Really, #5020 is about the drawer only.

toppk: I think i'm not writing precisely.  You want to maintain accountlist behavior as it was prior to this patch, but still want to remove the unified inbox account from the "drawer" account selector spinner.  you don't want to impact any other "screens" that may be using accountlist at this time.  did I get it right?

cketti: I think "account list" vs. `AccountList` was tripping me up. So yes, change the drawer, don't touch anything related to the `AccountList` class, i.e. the `ChooseAccount` and `LauncherShortcuts` activities.

toppk: now I see what I did.  I just assumed that AccountList.accounts were somehow making it's way over to the drawer, but the injection is of some other list of accounts.   it seems that specialaccounts is a somewhat incomplete metaphor.  perhaps I'm mistaken.

app/core/src/main/java/com/fsck/k9/K9.kt
toppk: done

toppk:
I've added a horizontal rule.  In one way it makes it less jarring that the second item ("Inbox") is by default selected when changing profiles, but in another way it disturbs the metaphor that unified inbox is just another folder.  Let me know which way you prefer.

Whatever happens, I'd like to squash the commits before you merge.  the final diff is quite a bit tidier then the commits currently suggest.  I've left it as is for now so as not to mess up the review.

cketti:
I love the name change to `K9.isShowUnifiedInbox`, but it distracts from the functional change. I created PR #5126 to do just the name change. Do you mind having a quick look and then rebasing this branch once the other PR has been merged?

toppk:
in the future I would prefer to merge my PR, and then refactor/tidy as you see fit afterwards.  this was a bit harder all the way around. 

There is a potentially showstopper issue, with this though, when you have selected unified inbox (not from startup) and then turn the setting off, the account is messagelist is empty, therefore forcing the drawer to be funny.  That was okay for the old drawer since it kinda sorta looked correct, and you need to select a profile/account before you can click on a folder and things are okay.

But with this patch, if you are on unified inbox and you disable show unified inbox, you are presented the drawer with all folder choices (minus unified inbox), but again, without an account, so selecting a folder without first setting a profile/account will crash.

I'd hold off merging until that is addressed.  Not a new issue, but the crash is a new possibility.

toppk:
This is my understanding of the issue (I'm not really an android developer):  If you have selected "unified inbox" and then turned off the setting,   MessageList.onCreate is called, which tries to reverse engineer the account based on the folder/search that is being displayed.  since it is a search all accounts search, there's no luck there.   It seems the best thing to do is to switch to default folder (inbox) for the account that was selected before the setting was flipped.  Does that sound right?

toppk:
Regarding the bug, I avoided the crasher by selecting the default account and deselecting the folder.

If you had more than one account configured, you can see this by:

1. enable unified inbox (if not enabled)
2. switch to secondary account
3. select unified inbox
4. open drawer, go to settings and turn unified inbox view off
5. when you return, the drawer will show the primary account profile without any folder highlighted, and the messagelist will still show unified inbox

I probably would have preferred going to default folder of the account that was last selected, but that solution is beyond the scope of this PR (and of my current knowledge of the codebase and framework).  I find the behavior in the PR somewhat intuitive.  One thing to note is that if you aren't viewing unified inbox, the drawer will reflect the currently listed folder/profile accurately.

On a related issue I was able to trigger a race condition by flipping not selecting "unified inbox" off, turning off   "show unified inbox" and  very quickly returning to the drawer.  This calls MessageList.onCreate/decodeExtras, and it is occationally clearly executing as if expected data isn't there.   I suspect synchronization is needed.  that decodeExtras is a beast and probably could use refactoring at some point.  I don't think this a blocker for  this PR.

cketti (approved):
Yes, `MessageList` is a mess that needs to be fixed. The workaround for the crash is good enough for now.

Thanks :+1:

app/ui/legacy/src/main/java/com/fsck/k9/ui/K9Drawer.kt
```diff
@@ -258,11 +231,26 @@ class K9Drawer(private val parent: MessageList, savedInstanceState: Bundle?) : K
...
+        if (K9.isShowUnifiedInbox) {
+            val drawerItem = PrimaryDrawerItem()
+                .withIcon(R.drawable.ic_inbox_multiple)
+                .withIdentifier(DRAWER_ID_UNIFIED_INBOX)
+                .withSelectedColor(selectedColor)
+                .withSelectedTextColor(accentColor)
+                .withName(R.string.integrated_inbox_title)
+            drawer.addItem(drawerItem)
+            drawer.addItem(DividerDrawerItem())
+            if (unifiedInboxSelected == true) {
```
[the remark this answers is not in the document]
toppk: i blame the IDE for that one.  it made me do it.  I should have restarted the IDE :)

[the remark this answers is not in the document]
cketti: It's possible that the type of `unifiedInboxSelected` was `Boolean?` at the time you created the statement. In that case you can't use `if (unifiedInboxSelected) …` but have to write it like `if (unifiedInboxSelected == true) …` to acknowledge the fact that `unifiedInboxSelected` might be `null`.

PR #5498, opened 2021-08-01
Button bar in navigation drawer

ByteHamster:
Work in progress. This PR is just to indicate that I am working on it. I did not pay attention to the code style yet.

ToDo:
- [x] Use nice icon for folder manager
- [x] Use nice icon for "refresh all" that users can tell apart from "refresh account"
- [x] Update content description text
- [x] Fix code style


cketti:
The refresh action is probably the most important one. How about we move it to the left and add "check mail" as text to its right? The remaining actions could then be aligned to the right edge of the drawer.

ByteHamster:
The old version also just showed an icon without text, didn't it? I think users will understand the icon without text. I was somehow missing the "settings" text more when trying this. Is there a chance that there will be more icons later? If so, I would stick to icons only

cketti:
> The old version also just showed an icon without text, didn't it? I think users will understand the icon without text.

I'm sure they would. However, Icon + Text would accomplish these things:
- large click target
- separate the one action from the two navigation options
- spread out the elements so the "toolbar" doesn't look that empty

> I was somehow missing the "settings" text more when trying this.

I know what you mean. It feels like one step forward (refresh button), but two steps back. Maybe we need to come up with a different plan :cry:

ByteHamster:
> spread out the elements so the "toolbar" doesn't look that empty

I think we will find more icons to display there. Users are already asking for a way to search for folders. Maybe there could be an icon for that, too.

> separate the one action from the two navigation options

We can also do that by placing the action on the left:



> Maybe we need to come up with a different plan

I think it's rather important to get the refresh button out (people having trouble discovering pull-to-refresh), so if you don't have a better idea, I would suggest to do it like in this PR.

---

Still thinking about how to do a "refresh all accounts" icon. Some ideas:

sfadschm:
I like this implementation in general.
Text labels for the bottom icons can be spared if the icons are well chosen. Thus I would vote for the first option with the "all" tag.
The other icons can have ambiguous meanings.

While I like the option, I fear that people who do not get the "pull to refresh" action will jump to requesting a second refresh button to refresh the currently open account.

cketti:
> Still thinking about how to do a "refresh all accounts" icon. 

The circle, but with two arrowheads right next to each other?

ByteHamster:
> The circle, but with two arrowheads right next to each other?

Made a rather crappy drawing but yeah, I think that could work.

ByteHamster:
The "manage folders" icon is from https://materialdesignicons.com/icon/folder-cog. Do we need to add credit for that?

cketti:
> The "manage folders" icon is from https://materialdesignicons.com/icon/folder-cog. Do we need to add credit for that?

Whether we have to or not, it would certainly be nice if we did. How about an "Icons" section in the *About* screen? We could also add the creator of the current [app icon](https://github.com/k9mail/k-9/pull/2082). (In the future I also want to add a list of contributors, translators, etc.)

To me it's not clear under which license the icon was published. But all of these look acceptable:
https://github.com/Templarian/MaterialDesign/blob/master/LICENSE

ByteHamster:
Should that about screen be done in this PR?

cketti:
Separate PR is probably best. If you don't want to or don't have the time to do it yourself, please create an issue so we remember to give credit to the icon author.

ByteHamster:
I think I'm ready for review here :)

> If you don't want to or don't have the time to do it yourself, please create an issue so we remember to give credit to the creator of the icon.

I will create an issue when this is merged.

app/ui/legacy/src/main/res/values/strings.xml
```diff
@@ -135,6 +135,8 @@ Please submit bug reports, contribute new features and ask questions at
     <string name="check_mail_action">Check mail</string>
     <string name="send_messages_action">Send messages</string>
     <string name="refresh_folders_action">Refresh folder list</string>
+    <string name="refresh_account">Look for new mail</string>
```
cketti, line 138: We used "Check mail" in the past (see `check_mail_action`). I think we should stick to that wording.

cketti (commented):
I don't know yet how to feel about this change. It makes the UI feel more cluttered. 

But we're at the beginning of a new development cycle and this is the time to try out new things. So I say we merge this and see how it feels in practice.

cketti:
Thanks

Issue #5614, opened 2021-08-24
Navigation drawer: Move Unified Inbox from "Folder View" to "Account View"


andrewfg:
The **Unified Inbox** drawer item should **not** be displayed in the navigation drawer **Folder View**, but rather it should be displayed in its **Account View** -- see mock up below.

Reasons are:
- Unified Inbox (although it is technically a folder rather than an account) is not a specific folder for any one single account
- Unified Inbox really has the character of an Account itself; i.e. an Account having only a single Inbox folder within it.

(Note: In #5519 we still have a pending issue to add both starred and unread counts to the Unified Inbox display).

![image](https://user-images.githubusercontent.com/893994/130484922-831f5f77-8e97-453c-9573-cb693a4e951c.png)

![image](https://user-images.githubusercontent.com/893994/130486338-5e1a4e7b-611b-48cb-82b6-65d5950f54a5.png)

ByteHamster:
The unified inbox was originally developed as a "special account". Opposite issue requesting it to be a folder instead of an account: https://github.com/k9mail/k-9/issues/5020

PR #5616, opened 2021-08-25
Unified Inbox: add starred & unread message counts; move to Account View

andrewfg:
This PR does two things..
1. Adds the starred and unread message counts to the Unified Inbox
2. Moves the Unified Inbox from Folder View to Account View

Note: according to https://github.com/k9mail/k-9/issues/5614 other people don't like item 2. above. However I am making the PR anyway so you can play with it. _**And in the worst case, you can revert item 2. but keep item 1.**_

image


andrewfg:
=> second attempt https://github.com/k9mail/k-9/pull/5623

PR #5637, opened 2021-09-01
update providers to add 2 of the domains offered by Mailo

abawany:
Updated `app/autodiscovery/providersxml/src/main/res/xml/providers.xml` to add details for Mailo.com.


abawany:
Mistake in the domains section - sorry.

PR #5645, opened 2021-09-04
Make unread icon look less like a compose button

ByteHamster:
The icons seem to cause quite a bit of confusion to users.

grafik

### DIFF
--- a/app/ui/legacy/src/main/res/drawable/ic_mark_new.xml
+++ b/app/ui/legacy/src/main/res/drawable/ic_mark_new.xml
@@ -1,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:tint="?attr/colorControlNormal"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path
+        android:fillColor="@android:color/white"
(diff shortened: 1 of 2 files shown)


cketti:
Just looking at the two icons you're proposing, it's not clear to me what's "mark as read" and what's "mark as unread". And that's with the context that it has to be one of the two. I think those two icons would increase confusion, not lower it.

I understand the envelope icon comes with a learning curve for users switching from 5.600. But every other email app I checked is using the icons we're using now for read/unread. I don't see a compelling reason to deviate from that. Users will adapt.

One thing we could do is to use a closed envelope with a dot for "mark as unread". This is what Outlook does:

outlook__mark_as_unread

ByteHamster:
Hmm, I thought a check mark would be pretty clear. So maybe check mark for "mark as read" and dot for "mark as unread"?

cketti:
It's an envelope/letter metaphor. If the envelope is open, you've read the letter. A check mark doesn't fit the metaphor.

ByteHamster:
> It's an envelope/letter metaphor. 

Yeah but the metaphor doesn't seem to work for many users. Anyway, I changed the icon to one with a dot. Let's see if that already helps.

grafik

cketti:
How about a less pronounced dot?

image

[envelope_dot_4.zip](https://github.com/k9mail/k-9/files/7124469/envelope_dot_4.zip)

cketti:
After playing around a bit with different dot variants I'm not sure it's a good idea anymore. But let's just go with your variant for a while and see how it feels.

PR #5646, opened 2021-09-05
Added support for left/right swiping between emails

andrewfg:
Added support for left/right swiping between emails in message view.

### DIFF
--- a/app/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
+++ b/app/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
@@ -87,7 +87,8 @@ open class MessageList :
     MessageViewFragmentListener,
     FragmentManager.OnBackStackChangedListener,
     OnSwitchCompleteListener,
-    PermissionUiHelper {
+    PermissionUiHelper,
+    SwipeCatcher {
 
     private val recentChangesViewModel: RecentChangesViewModel by viewModel()
 
@@ -136,16 +137,6 @@ open class MessageList :
     private lateinit var recentChangesSnackbar: Snackbar
 
     private val coroutineScope: CoroutineScope = CoroutineScope(Dispatchers.Default)
-    private val onSwipe =
-        SwipeCatcher { swipeToLeft ->
-            coroutineScope.launch {
-                if (swipeToLeft) {
-                    showNextMessage()
-                } else {
-                    showPreviousMessage()
-                }
-            }
-        }
 
     public override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
@@ -1252,7 +1243,7 @@ open class MessageList :
             }
 
             val fragment = MessageViewFragment.newInstance(messageReference)
-            fragment.setSwipeCatcher(onSwipe)
+            fragment.setSwipeCatcher(this)
             val fragmentTransaction = supportFragmentManager.beginTransaction()
             fragmentTransaction.replace(R.id.message_view_container, fragment, FRAGMENT_TAG_MESSAGE_VIEW)
             fragmentTransaction.commit()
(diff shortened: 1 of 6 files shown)


cketti:
We already had an implementation similar to this. It wasn't great. See #585.

andrewfg:
Thanks for the feedback.
So what do you want me to do?

andrewfg:
To answer my question above, I guess you want me to add a view left/right slide animation? And/or a haptic 'bump' when the transition occurs?

andrewfg:
Added visual animation and haptic feedback

ByteHamster:
I think the idea is to use a ViewPager. That's pretty hard to implement, though, because of K-9's current code structure.

andrewfg:
Indeed a ViewPager is pretty hard to implement :( ..
.. however I think that my solution with TranslateAnimation actually looks quite good; => perhaps you could build it and look at it yourself.

cketti:
The important part is the user experience. When moving the finger the "surface" underneath it needs to move with the finger; just like scrolling a list. This is complicated by the fact that WebView can also scroll horizontally if the content is wider than the available screen width, e.g. because it was zoomed in. I suggest you check out the Gmail app for how it should feel like.

andrewfg:
See screen recording below..

https://user-images.githubusercontent.com/893994/132246868-f515126d-30cb-43b6-afa0-8a66a7a37723.mp4

andrewfg:
> This is complicated by the fact that WebView can also scroll horizontally

Yeah. I am still thinking about how to handle that..

andrewfg:
PS I think the animation is still a bit 'lumpy' but after using it for some time I already have some ideas to make it smoother...

cketti:
It doesn't matter how smooth the animations are. A change that doesn't implement the UX outlined above (and in #585) has no chance of getting merged.

andrewfg:
Okay, I have done a few more things today..
- Made big improvements in the swipe animations (see movie below)
- Disable the respective swipe direction if already at the start or end of the message list
- Disable swipe capture if the touch movement is on the WebView and it is zoomed

https://user-images.githubusercontent.com/893994/132386770-cbedc9f5-99a2-4663-8a2b-904f9b535762.mp4

andrewfg:
> Disable swipe capture if the touch movement is on the WebView and it is zoomed

Next I will try to add code to re-enable the swipe capture again if the touch movement is on the WebView and it is zoomed AND it is at its scroller left / right limit respectively.

ByteHamster:
Please don't invest a lot of your time into an animation based method. It won't get merged, as cketti wrote. Unless you basically implement a ViewPager manually, it will never feel as smooth as a real ViewPager.

See the following gif (from another app) for example. You cannot achieve that kind of natural feeling with any kind of plain animation.
device-2021-09-07-222959 mp4

andrewfg:
I am not going to write a ViewPager since that is almost a total rewrite. But I will continue with my current approach because I think I can make it work, and look quite good. Obviously its up to you guys to decide whether to accept a solution now that is good enough, kr wajt forever

andrewfg:
Okay, I have finished my coding, so its ready for your review.

Please excuse me if I am wrong, but I got the impression that you guys already formed a negative opinion of this PR, and if so, I do kindly ask that you actually build this version and try it physically on your phone so that you get a proper impression of the look & feel, and the functionality.

Many thanks in advance.

cketti:
I tested this PR on a real device. It doesn't feel like it should. Like I wrote before, I don't want to merge a change that doesn't get the UX right.

andrewfg:
@cketti thanks for the feedback; I tried to implement everything mentioned in #585, so I am not sure what is missing. Can you please be more specific about what you don't like, and want you want done differently?

(Or are you just saying that you will NEVER accept any PR (from me) that tries to address this user need?)

cketti:
Have a look at one of the many apps that support swiping between pages. I mentioned Gmail. ByteHamster even posted a video. That's the behavior that needs to be matched. See also https://material.io/design/interaction/gestures.html

andrewfg:
Ok I will study your proposals. But just one more question: we are trying to swipe between two Views; one is the actual current message View on the screen, and the other is the View of the next/previous message. It is rather easy to have the current message View to tracking the user's touch motions and animate its DIS-appearance (it is just a matter of tuning the dynamics). However it is much more difficult to animate the (potentential) appearance of the next/previous message view; since we have to pre-create and display it in anticipation that the swipe action might actually be completed. So in this PR I was just animating the disappearance of the old view and just letting the new one pop in. If you want the new View to be animated as well, then I am not sure if it is possible within the current architecture. Please advise.

ByteHamster:
> However it is much more difficult to animate the (potentential) appearance of the next/previous message view; since we have to pre-create and display it in anticipation

That's exactly the problem of doing it with a manual animation. When using a ViewPager with fragments, you get that automatically. Like I said earlier, adding the ViewPager won't be easy because it needs some code refactoring (no longer keep a reference to a message fragment instance in MessageList) but that refactoring will be a positive thing, too.

andrewfg:
> that refactoring will be a positive thing, too.

Ha. ;)

PR #6662, opened 2023-02-13
Change root build files to Kotlin

wmontwe:
In order to apply new build configuration and harmonize the setup, using Kotlin instead of Groovy simplifies the implementation efforts.


cketti (approved):
LGTM

PR #6674, opened 2023-02-15
Add compose support

wmontwe:
Add Jetpack compose support with a dedicated build plugin and simple demo app.

Jetpack Compose is not enabled yet for the main app as it has impact on the build memory pressure. We need to increase the `gradle.properties` allowance to 2g. Also more testing is needed to see if there are issues with API 31 as target SDK for Android and if we need to upgrade first.


cketti (approved):
LGTM :+1:

PR #6714, opened 2023-03-02
Add core common modules

wmontwe:
Add `:core:common` module for JVM only components and `:core:android:common` module for components that have a hard Android dependency


cketti (approved):
LGTM

PR #6757, opened 2023-03-14
Change truth testing library to legacy

wmontwe:
Change truth testing library to legacy and only allow usage in old modules. 

This changes some small modules to [assertk](https://github.com/willowtreeapps/assertk) already.


cketti (approved):
LGTM

PR #6763, opened 2023-03-16
Change compose build plugins

wmontwe:
Change app and library compose build plugins to share dependencies and add compose jUnit test dependencies. 

Also ensure that Gradle extensions are internal to avoid leaking into the rest of the project.


cketti (approved):
LGTM

PR #6766, opened 2023-03-16
Fix editor config import rule for aliases

wmontwe:
The editor config rule had an error that didn't move alias imports to the end of the import statement when formatted in Android Studio. This caused issues with `ktlint` as it correctly expects them at the end.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,7 @@ indent_style = space
 insert_final_newline = true
 
 [*.{kt,kts}]
-ij_kotlin_imports_layout = *,^*
+ij_kotlin_imports_layout = *,^
 ij_kotlin_allow_trailing_comma = true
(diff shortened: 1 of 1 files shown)


cketti (approved):
I'm too lazy to look up the syntax for this setting. I trust you that this is doing the right thing :smile:

wmontwe:
This is what I found:

```
ij_kotlin_imports_layout
# Defines the imports layout. The layout can be composed by the following symbols:
# "*" - wildcard. There must be at least one entry of a single wildcard to match all other imports. Matches anything after a specified symbol/import as well.
# "|" - blank line. Supports only single blank lines between imports. No blank line is allowed in the beginning or end of the layout.
# "^" - alias import, e.g. "^android.*" will match all android alias imports, "^" will match all other alias imports.
```

PR #6803, opened 2023-04-04
Add feature preview app

wmontwe:
This adds a new app module to showcase new app features and starts with adding onboarding as first showcase. It includes a Composable navigation setup and a splashscreen. 

Issues with the onboarding logo background have been fixed too and button styles are aligned with current design.


app-feature-preview/src/main/res/mipmap-xxxhdpi/ic_launcher.png
cketti, line 1: On older Android versions apps typically use app icons without a background/shape drawn around them. We're using this icon for the main app: https://github.com/thundernest/k-9/blob/1a302a2dd34169d1c0b568363c8a52e8ed06b988/app/ui/legacy/src/main/res/mipmap-xxxhdpi/icon.png

For non-production apps it's probably also enough to include the xxxhdpi-variant of the app icon in the repository.

core/ui/compose/theme/src/main/java/app/k9mail/core/ui/compose/theme/color/SystemBarsColor.kt
```diff
@@ -0,0 +1,26 @@
...
+fun SetSystemBarsColor(
+    color: Color,
+) {
+    val systemUiController = rememberSystemUiController()
+    val isDarkColor = color.luminance() < MAX_DARK_COLOR_LUMINANCE
+
+    DisposableEffect(Unit) {
+        systemUiController.setSystemBarsColor(
+            color = color,
+            darkIcons = !isDarkColor,
```
cketti, line 21: It took me a moment to understand this. Maybe `isDarkColor` should be `isDarkBackgroundColor`?

feature/onboarding/src/main/kotlin/net/thunderbird/feature/onboarding/OnboardingContent.kt
```diff
@@ -28,19 +29,21 @@ import app.k9mail.core.ui.compose.theme.K9Theme
 import app.k9mail.core.ui.compose.theme.MainTheme
 import app.k9mail.core.ui.compose.theme.ThunderbirdTheme
 
+private val COLOR_LOGO_BACKGROUND = Color(color = 0xFFEEEEEE)
```
cketti, line 32: Shouldn't this be part of the theme?

app-feature-preview/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
+    <background android:drawable="@color/ic_launcher_background"/>
+    <foreground android:drawable="@drawable/ic_launcher_logo"/>
+    <monochrome android:drawable="@drawable/ic_launcher_logo" />
```
cketti, line 5: `@drawable/ic_launcher_logo` is not a monochrome icon.

app-feature-preview/src/main/res/mipmap-anydpi-v26/ic_launcher_round.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
+    <background android:drawable="@color/ic_launcher_background"/>
+    <foreground android:drawable="@drawable/ic_launcher_logo"/>
+    <monochrome android:drawable="@drawable/ic_launcher_logo" />
```
cketti, line 5: `@drawable/ic_launcher_logo` is not a monochrome icon.

cketti:
The padding should probably be set on the content inside the scrollable container.

image

app-feature-preview/src/main/res/mipmap-anydpi-v26/ic_launcher_round.xml
wmontwe: See above

app-feature-preview/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
wmontwe: The issue with the current monochrome vector is that it fails for API 21-23. Not shure why not for the legacy project.

wmontwe: Lint then complains about a missing monochrome icon

cketti: Seems fair, we don't have one :smile: Pretending that we do, just to make Lint happy, seems wrong. Let's just suppress the warning for test apps.

core/ui/compose/theme/src/main/java/app/k9mail/core/ui/compose/theme/color/SystemBarsColor.kt
wmontwe: But the `SystemUiController` uses color as terminology and we could cut the discussion here as it applies the luminance check by default anyway. So this check could go.

cketti (approved):

core/ui/compose/theme/build.gradle.kts
```diff
@@ -9,5 +9,6 @@ android {
 
 dependencies {
     api(projects.core.ui.compose.common)
+    api(libs.accompanist.systemuicontroller)
```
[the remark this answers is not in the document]
wmontwe: It was suggested by a dependency check, but I actually don't see a reason why it should be `api`.

PR #6889, opened 2023-05-09
Fix NPE on devices that don't support widgets

IlyaBizyaev:
Turns out, `AppWidgetManager.getInstance(context)` can return null:

* https://github.com/chromium/chromium/blob/08033c4b62f48e46e66f97be3a39c99ab9163402/chrome/android/java/src/org/chromium/chrome/browser/quickactionsearchwidget/QuickActionSearchWidgetProvider.java#L289
* https://github.com/tasks/tasks/blob/d25f859385e8301ae274835297f151125f7cefb7/app/src/main/java/org/tasks/widget/AppWidgetManager.kt#L18

I've tested that this change fixes K-9 crash on the device.

### DIFF
--- a/app/ui/message-list-widget/src/main/java/app/k9mail/ui/widget/list/MessageListWidgetManager.kt
+++ b/app/ui/message-list-widget/src/main/java/app/k9mail/ui/widget/list/MessageListWidgetManager.kt
@@ -23,6 +23,9 @@ class MessageListWidgetManager(
 
     fun init() {
         appWidgetManager = AppWidgetManager.getInstance(context)
+        if (appWidgetManager == null) {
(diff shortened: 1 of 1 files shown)


wmontwe (changes requested):
Thanks for fixing the issue. Indeed the `AppWidgetManager.getInstance(context)` is just a wrapper of `context.getSystemService(Context.APPWIDGET_SERVICE)` which is nullable.

Could you please add the following to the init block to document that no AppWidgetManager is present?

```
if (appWidgetManager == null){
    Timber.v("Message list widget is not supported on this device.")
}
```

IlyaBizyaev:
Done :)

wmontwe (approved):
LGTM and thank your for your support!

PR #6890, opened 2023-05-10
Add account setup atoms and molecules

wmontwe:
This adds atoms and molecules to the design system, that are needed for the account setup screens.

Atoms:
- Changed `Text` to extend with color and annotated string properties
- Added `TextFieldOutlinedEmailAddress` with keyboard options set to email input
- Added `Icon` to wrap Material Icon
- Added `LoadingView` and `ErrorView`

Molecules:
- Added `EmailAddressInput` to combine `TextFieldOutlinedEmailAddress` with error messages
- Added `PasswordInput` to combine `TextFieldOutlinedPassword` with error messages

It also adds `Icons` to the theme to wrap material icons.

All new views have been added to the catalog app. It is growing in size and needs a makeover to better section it's content in the future.


cketti (approved):
LGTM

PR #6915, opened 2023-05-23
Add `Icon` and `ButtonIcon` atoms and `Icons` to the theme

wmontwe:
This adds `Icon` and `ButtonIcon` atoms to the design system accompanied by a selection of Material icons made available as `Icons` in the theme.


cketti (approved):
LGTM

PR #6930, opened 2023-05-26
Change catalog and feature app to use new insets api

wmontwe:
With Compose 1.2 window insets could be handled by using modifiers. This changes the setup of the Catalog and Feature apps to use these. See [WindowInsets and IME Animations](https://developer.android.com/jetpack/compose/migrate/other-considerations#ime-animations)

This also removes the TopAppBar flickering when used with a Scaffold on my Samsung device.


cketti (approved):
LGTM

But this doesn't quite match what the main app is doing. I believe we're currently always using a navigation bar with a solid color.

PR #6945, opened 2023-05-31
Add account setup options validation

wmontwe:
This changes the validaton approach a bit by using a ValidationError interface istead of Exceptions. This gives more freedom to handle error states as sealed interfaces or classes for exhaustive handling.

Updates AccountSetupOptions to use validation on submit.


cketti (approved):
LGTM

PR #6972, opened 2023-06-09
Inconsistent recipient names when composing a message

gitstart:
fixes https://github.com/thundernest/k-9/issues/6816
### PR Description
This PR ensures that the recipient's name is consistent when composing a message regardless of whether they are found using a nickname or not

### Steps to reproduce

1. Have a contact with the nickname different from the name.
2. Create a new message in K9
3. Start typing in "To"

### Issue loom
https://www.loom.com/share/429b833c1d8c46c4aeeaaae1902f0459

### Fix loom
https://www.loom.com/share/554b780adeb14b0aa0e4b3579b17ec55

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.

### DIFF
--- a/app/ui/legacy/src/main/java/com/fsck/k9/activity/compose/RecipientLoader.java
+++ b/app/ui/legacy/src/main/java/com/fsck/k9/activity/compose/RecipientLoader.java
@@ -72,7 +72,6 @@ public class RecipientLoader extends AsyncTaskLoader<List<Recipient>> {
 
     private static final String[] PROJECTION_NICKNAME = {
             ContactsContract.Data.CONTACT_ID,
-            ContactsContract.CommonDataKinds.Nickname.NAME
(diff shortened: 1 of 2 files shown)


gitstart:
> This change has the desired effect, but it obscures the intended behavior when reading the code.
> 
> Currently, the intended behavior is to use the nickname in the outgoing message when the recipient was added using the nickname. `fillContactDataFromCursor()` has a parameter `prefilledName` to accomplish that. When changing the behavior, that mechanism should be removed entirely, i.e. the `prefilledName` parameter should be removed. Then `PROJECTION_NICKNAME` can be reduced to just fetch the contact ID.

Thank you for reviewing PR @cketti, 
We have attended to the comment. Please take a look.

cketti (approved):

PR #6973, opened 2023-06-12
Change input to allow booleans with error

wmontwe:
We need to display an error in case a switch or checkbox state is required. This introduces a BooleanInputField and wraps CheckboxInput and SwitchInput with an error. 

This also decouples shapes from Material Theme to be used in downstream modules.

Text composables now also support text alignment.

Changes Buttons to allow custom color.


cketti (approved):
LGTM

PR #7009, opened 2023-06-20
Change from BuildConfig to OAuthProviderSettingsFactory

wmontwe:
Change away from BuildConfig properties to factories to create different OAuthProviderSettings per build type.

This is a proposal on top of #7007 for further discussion.


wmontwe:
This needs further changes

PR #7038, opened 2023-06-29
Add launch import settings

wmontwe:
This adds the ability to launch import settings from the new onboarding.

Depends on: #7037


cketti (approved):
The line `findNavController().popBackStack()` in `SettingsImportFragment` should to be changed to `findNavController()?.popBackStack()` because `FragmentLauncherActivity` doesn't initialize a `NavController`.

wmontwe:
> The line `findNavController().popBackStack()` in `SettingsImportFragment` should to be changed to `findNavController()?.popBackStack()` because `FragmentLauncherActivity` doesn't initialize a `NavController`.

I had a look at the underlying implementation, it might throw an `IllegalStateException`.  Added a wrapper.

PR #7076, opened 2023-07-19
Change WizardNavigationBar to stateful

wmontwe:
Change WizardNavigationBar to enable next and back button based on given state


cketti (approved):
LGTM

PR #7088, opened 2023-07-26
Add ValidateServerSettings use case

wmontwe:
This is supposed to replace `CheckIncomingServerConfig` and `CheckOutgoingServerConfig` use cases as they are not needed anymore with the separate account validation step.


cketti (approved):
LGTM

PR #7104, opened 2023-07-31
Change SignInWithGoogleButton to fit to our button style

wmontwe:
Also added outline for better contrast in light mode and removed the loading indicator as it's not needed.

#### Old:

old

#### New:

new

Depends on: #7102


cketti (approved):
LGTM

PR #7144, opened 2023-08-29
Change line break rule to avoid issues with contributions from Windows

wmontwe:
This follows the recommendations from this excellent atricle: [CRLF vs. LF: Normalizing Line Endings in Git](https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/)


wmontwe (commented):
👍

PR #7152, opened 2023-09-01
Recipient address <xxx@gmail.com/> is not a valid RFC-5321 address.

gitstart:
fixes https://github.com/thundernest/k-9/issues/7112

### PR Description
This PR adds extra check to the recipient email(s) and intercept sending an email if the email address is wrong

### Issue video
https://www.loom.com/share/52a9937220cb4b1fa36a0108332bba6e
### Demo video
https://www.loom.com/share/45e42fdcba0c4e7595153442b8b84aec

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.


app/ui/legacy/src/main/java/com/fsck/k9/activity/compose/RecipientMvpView.kt
```diff
@@ -265,6 +265,10 @@ class RecipientMvpView(private val activity: MessageCompose) : View.OnFocusChang
         return toView.hasUncompletedText()
     }
 
+    fun recipientToIsValid(): Boolean {
+        return toView.hasValidEmail(toView.text.toString())
```
Goooler, line 269: Need to check CC and BCC as well.

Gus1004 (changes requested):
K

app/ui/legacy/src/main/java/com/fsck/k9/view/RecipientSelectView.java
```diff
@@ -61,6 +62,7 @@ public class RecipientSelectView extends TokenCompleteTextView<Recipient> implem
     private static final int LOADER_ID_ALTERNATES = 1;
 
 
+    private EmailAddressValidator validator = new EmailAddressValidator();
```
wmontwe, line 65: Could you please change this to depdency injection and it could also be set to final?

```suggestion
    private final EmailAddressValidator validator = DI.get(EmailAddressValidator.class);
```

app/ui/legacy/src/main/java/com/fsck/k9/activity/compose/RecipientPresenter.kt
```diff
@@ -130,6 +130,21 @@ class RecipientPresenter(
             return true
         }
 
+        if (!recipientMvpView.recipientToIsValid()) {
```
wmontwe, line 133: The order of the checks should be rearranged as follows:

Place the following check after the condition that verifies if `toAddresses`, `ccAddresses`, and `bccAddresses` are empty:

```kotlin
if (toAddresses.isEmpty() && ccAddresses.isEmpty() && bccAddresses.isEmpty()) {
    recipientMvpView.showNoRecipientsError()
    return true
}
```

Doing so ensures that the user is informed to add at least one recipient, rather than receiving a misleading message about invalid input in the fields.

wmontwe (changes requested):
Thanks for the PR! Great work, but there are a few changes needed. 

Please update when you can.

PR #7156, opened 2023-09-05
Add account validation module

wmontwe:
This moves the server settings validation to it's own module for use in the account edit feature.


wmontwe (commented):
👍

PR #7169, opened 2023-09-13
Refactor account server settings

wmontwe:
This is renaming the `account:server:config` module to `accountsserver:settings` for consistency. It applies the same pattern also for it's class names.


wmontwe (commented):
👍

PR #7171, opened 2023-09-13
Move email validation to account setup

wmontwe:
As part of the account edit feature. This moves the email validatio to the account setup module.


wmontwe (commented):
👍

PR #7172, opened 2023-09-13
Change injection for server validation

wmontwe:
This changes named injection to typed injection for the server validation to avoid issues with dependency resolution. Part of the edit feature efforts


wmontwe (commented):
:+1:

PR #7173, opened 2023-09-14
Fix gradle properties to avoid out of memory

wmontwe:
According to [#19750](https://github.com/gradle/gradle/issues/19750) the Gradle Daemon could disapear or run out of memory when `org.gradle.jvmargs` has been altered in the `gradle.properties`. This causes the deamon to drop all default values leaving `XX:MaxMetaspaceSize` unbound which consumes more and more memory.

This should fix the issues until [#19750](https://github.com/gradle/gradle/issues/19750) is fixed.

Here are the default values used by the Gradle Daemon: [DaemonParameters](https://github.com/gradle/gradle/blob/master/subprojects/launcher/src/main/java/org/gradle/launcher/daemon/configuration/DaemonParameters.java). 

For our project the default memory settings have been multiplied by 4.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,9 +9,13 @@ android.useAndroidX=true
 android.enableJetifier=false
 android.nonTransitiveRClass=true
 
+# Gradle
 ## Ensure important default jvmargs aren't overwritten. See https://github.com/gradle/gradle/issues/19750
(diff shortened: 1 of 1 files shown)


wmontwe (commented):
:+1:

PR #7174, opened 2023-09-15
Add account updater

wmontwe:
This adds an AccountUpdater to update current account information in the legacy code base. This will be used in the account edit feature.

Also adds the `feature:account:edit` module.


wmontwe (commented):
👍

PR #7193, opened 2023-09-25
Remove legacy account setup and onboarding, part 1

wmontwe:
This is the 1st iteration of removing the legacy account setup und onboarding code that is not used by other parts of the app.


cketti (approved):
LGTM

PR #7223, opened 2023-10-06
Fix signingConfig hack

wmontwe:
This removes the hack for F-Droid and replaces it by a simplyfied assignment. This should work as the receiving `signingConfig` is nullable and `signingConfigs.findByName("release")` is either null or returns the release config.
Having a one liner allows F-Droid to strip the `signingConfig` line without tripping detekt or spotless afterwards.

From what I discovered: The troublesome code tried to fix another issue that arrised due to the use of `signingConfigs.getByName("release")`, which throws an `UnknownDomainObjectException`. But with `findByName("release")` no issue.

### DIFF
--- a/app/k9mail/build.gradle.kts
+++ b/app/k9mail/build.gradle.kts
@@ -83,11 +83,7 @@ android {
 
     buildTypes {
         release {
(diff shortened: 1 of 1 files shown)


cketti (approved):
:+1:

PR #7231, opened 2023-10-10
Change Gradle config to accommodate for latest changes

wmontwe:
Minor modification to comply with the latest Gradle version. The key change involves relocating the `build-plugin` to `pluginManagement` within `settings.gradle.kts`. This resolves the issue where the `build-plugin` should exclusively be added to the Gradle plugins and not leak into the whole project.


cketti (approved):
LGTM

PR #7269, opened 2023-10-19
Disable Gradle configuration cache

wmontwe:
This disables Gradle configuration cache as Spotless doesn't support it properly.

I fails with the following message and requires user action to remove current cache in case it got stale. This happens a lot if you switch between different branches of the project.

```
Spotless JVM-local cache is stale. Regenerate the cache with
    rm -rf .gradle/configuration-cache
```


cketti (approved):
:disappointed:

PR #7279, opened 2023-10-21
Refactor MailToTest class to Kotlin

adamszewe:
## Description:
Refactor the `com.fsck.k9.helper.MailToTest` class from Java to Kotlin.

## Changes:
* refactor the `com.fsck.k9.helper.MailToTest` class from Java to Kotlin
* use assertk matchers as required by the [latest ADR](https://github.com/thundernest/k-9/blob/main/docs/architecture/adr/0003-switch-test-assertions-from-truth-to-assertk.md)


cketti (changes requested):
Thanks for taking the time to contribute to K-9 Mail :+1:

I appreciate that you're trying to reduce the amount of Java code in this project. However, the Kotlin code generated by the automatic conversion isn't great. Also, for this code in particular I was planning to switch to `androidx.core.net.MailTo`. I don't think it makes sense to convert this class to Kotlin right now.

Please change your pull request to only convert `MailToTest` from Java to Kotlin. Create a commit that only renames the file from `.java` to `.kt`. Then a separate commit for the code conversion. That way it's easier to track the Git history of the file. Android Studio supports this by showing a "Extra commit for .java > .kt renames" checkbox in the commit dialog.

Read other tests to learn how to avoid the `!!` operator and how to compare lists.

In general, converting existing code to Kotlin is probably not the best way to start contributing to K-9 Mail. It usually involves a lot of code to review and touches on a lot of code style questions. The past has shown it's also rather easy to introduce NullPointerExceptions.

Instead of converting code to Kotlin, try fixing bugs. Usually figuring out what went wrong is the hard part. Often the fix itself doesn't involve a lot of code and is thus easy to review.

adamszewe:
Hello and thank you for your feedback 😄 
I have updated the PR as requested and the description to match.

> In general, converting existing code to Kotlin is probably not the best way to start contributing to K-9 Mail. It usually involves a lot of code to review and touches on a lot of code style questions. The past has shown it's also rather easy to introduce NullPointerExceptions.
> 
> Instead of converting code to Kotlin, try fixing bugs. Usually figuring out what went wrong is the hard part. Often the fix itself doesn't involve a lot of code and is thus easy to review.

Understood 😄 
I'll seek to contribute in more meaningful ways next by tackling some bugs 🐛

cketti (approved):
Thanks :+1:

PR #7284, opened 2023-10-22
Set GroupAlertBehavior for Single (and Summary) Notifications

### LINKED ISSUE #7271: Notifications silent on Wear OS 4
### Checklist

### App version

6.603

Google Play

### Android version

14

### Device model

Pixel 8 Pro / Pixel Watch 2

### Steps to reproduce

1. Phone and watch paired
2. Receive an email

### Expected behavior

Notification appears as popup on watch with vibration and ringtone

### Actual behavior

Notification appears silent in background (on phone itself it vibrates and rings)

This is the only app where I noticed this behaviour. 

### Logs
[logs.txt](https://github.com/thundernest/k-9/files/13051966/logs.txt)

marcelmu:
To fix notifications on WearOS being handled silently an explicit `GroupAlertBehaviour` had to be set.

Fixes #7271


cketti (approved):
:+1:

PR #7318, opened 2023-11-02
Change repository links

wmontwe:
Change repository links from `thundernest/k-9` to `thunderbird/thunderbird-android`


cketti (approved):

williamdes:
by the way, I noticed this change since it broke my email filters. Do you have an announcement to link somewhere about this namespace change ?
Very cool that it is now named thunderbird !

PR #7384, opened 2023-11-28
Change ServerSettings to support loading state

wmontwe:
Solves: #7295


cketti (approved):
:+1:

In the future we might want to change `ContentLoadingErrorView` to optionally disable animations. It feels unnecessary to animate when the loading view is empty (like in this case).

PR #7389, opened 2023-11-29
Set button corners to material default values

bernhard-fp:
Set button corners to material default values

Reference: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material/material/src/commonMain/kotlin/androidx/compose/material/Shapes.kt;l=44-58;drc=696caaf1f75d83dc5fbe4ec30434021db31668e8


wmontwe (approved):
Thank you for your contribution, approved.

PR #7425, opened 2023-12-08
Change lint config to lower ExtraTranslation to warning level

wmontwe:
The feature app's build is having errors due to translated resource strings whose source has been removed. These are being flagged by lint. To address this, I've adjusted the configuration to downgrade these errors to warnings.


wmontwe:
Still investigating what's the source of the problem -> don't merge

cketti (approved):

PR #7426, opened 2023-12-08
Add `FixedScaleImage` to design system

wmontwe:
This is going to be used in the onboarding feature for the background image placement.

Depends on: #7425


cketti (approved):
LGTM

PR #7427, opened 2023-12-08
Add success onboarding screen

wmontwe:
This is a functional success screen, but the design is not ideal. To match the intended look, there needs to be another design session to address some issues with the assets available in Figma.

grafik

Depends on: #7425


feature/onboarding/success/src/main/kotlin/app/k9mail/feature/onboarding/success/ui/SuccessContent.kt
```diff
@@ -0,0 +1,91 @@
...
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import app.k9mail.core.ui.compose.designsystem.atom.button.Button
+import app.k9mail.core.ui.compose.designsystem.atom.text.TextBody1
+import app.k9mail.core.ui.compose.designsystem.atom.text.TextHeadline4
+import app.k9mail.core.ui.compose.theme.MainTheme
+import app.k9mail.feature.onboarding.success.R
+
+@Composable
+internal fun SuccessLogo(
```
Mukuljangir372, line 22: Instead of writing composables inside `SuccessContent`, It should come under `components` package (used for defining reusable composables). What do you think?

PR #7526, opened 2024-01-18
Fix AndroidAlarmManager 

wmontwe:
It uses `ContextCompat` and `ContextCompat.RECEIVER_NOT_EXPORTED` to work with API 34.

Using `ContextCompat` is recommended for all apps targeting API 33 and higher, see [Safer exporting of context-registered receivers](https://developer.android.com/about/versions/13/features#runtime-receivers). The downside is, that our current project setup needs to be tweaked to apply the proper permission for tests.

For now the affected tests are moved to the app module. I'll add a ticket to address the dependency setup to fix these.


cketti (approved):
LGTM

PR #7560, opened 2024-01-25
Fix lint complaining about naming

wmontwe:
Second pass on this lint issues, see #7546. Lint started to complain as soon I did a clean with ignore caches.

Added `lint-baseline.xml` file


wmontwe:
> We should probably configure Lint to ignore all errors related to translations. Weblate already flags these. Developers are the wrong recipients for these types of errors. Ignoring them via a baseline file will just mean we have to update the baseline all the time because translations are fairly dynamic.

That's a valid idea, but this is blocking my development now. So better to move that to a later point and do it properly then.

wmontwe:
And another task, figure out why this was able to slip through in the first place.

cketti (approved):

PR #7586, opened 2024-01-31
Dependency updates

wmontwe:
This will update Compose to v1.6 and all Compose Material dependencies are now also determined by the BOM version.

Sadly the new `TextField2` was pushed to v1.7 alpha.

Bumped:
- AndroidX Compose BOM 2023.10.1 -> 2024.01.00
- Android Gradle Plugin 8.2.1 -> 8.2.2
- Android Tools 31.2.1 -> 31.2.2
- KSP 1.9.22-1.0.16 -> 1.9.22-1.0.17
- Moshi 1.15.0 -> 1.15.1
- LeakCanary 2.9.1 -> 2.13

This also changed the dependency substitutions by removing unambiguous declarations. We should consider removing all of them, see #7587.


cketti (approved):
LGTM

PR #7631, opened 2024-02-12
Add `translation-cleaner` script

wmontwe:
This script deletes all languages that are below a given threshold.

Usage: 
- `translation-cleaner <weblate-token> [threshold]`
- defaults to 70%


wmontwe:
Close this in favor of #7638

PR #7681, opened 2024-02-29
Fix passwords validation to exclude linebreaks

### LINKED ISSUE #7670: Don't allow passwords containing line breaks
### Checklist

### App version

6.717

None

### Android version

X

### Device model

_No response_

### Steps to reproduce

Start configuring your icloud.com email

### Expected behavior

Password checks and it goes to next step

I did the setup on FairEmail at the same time and it worked. 

### Actual behavior

Server error: unmatch quote

### Logs

```
02-24 22:58:27.993 20341 10889 V RealImapConnection: conn72600286 <<< #null# [OK, [CAPABILITY, XAPPLEPUSHSERVICE, IMAP4, IMAP4rev1, SASL-IR, AUTH=ATOKEN, AUTH=PLAIN], (2409B40-f42d5d1bd981) mr36p00im-tygg11200301.me.com]
02-24 22:58:27.994 20341 10889 D RealImapConnection: Saving [IMAP4, AUTH=PLAIN, XAPPLEPUSHSERVICE, SASL-IR, AUTH=ATOKEN, IMAP4REV1] capabilities for conn72600286
02-24 22:58:27.995 20341 10889 V RealImapConnection: conn72600286>>> 1 AUTHENTICATE PLAIN 

02-24 22:58:28.166 20341 20341 D InsetsController: show(ime(), fromIme=true)
02-24 22:58:28.223 20341 10889 V RealImapConnection: conn72600286<<<#+# []
02-24 22:58:28.651 20341 10889 V ImapResponseParser: conn72600286<<<#1# [NO, [AUTHENTICATIONFAILED], Authentication Failed]


RealImapConnection: conn72600286>>> 2 LOGIN "myusername" "ykkl-xxx-xxxt-kkki
02-24 22:58:28.655 20341 10889 V RealImapConnection: 
02-24 22:58:28.655 20341 10889 V RealImapConnection: "
02-24 22:58:28.848 20341 10889 V ImapResponseParser: conn72600286<<<#2# [BAD, unmatch quote]
```

wmontwe:
Fixes #7670

### DIFF
--- a/feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/domain/usecase/ValidatePassword.kt
+++ b/feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/domain/usecase/ValidatePassword.kt
@@ -10,12 +10,17 @@ class ValidatePassword : UseCase.ValidatePassword {
     override fun execute(password: String): ValidationResult {
         return when {
             password.isBlank() -> ValidationResult.Failure(ValidatePasswordError.EmptyPassword)
-
+            password.contains(LINE_BREAK) -> ValidationResult.Failure(ValidatePasswordError.LinebreakInPassword)
             else -> ValidationResult.Success
         }
     }
 
     sealed interface ValidatePasswordError : ValidationError {
         data object EmptyPassword : ValidatePasswordError
+        data object LinebreakInPassword : ValidatePasswordError
+    }
+
+    private companion object {
+        const val LINE_BREAK = "\n"
     }
 }
--- a/feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/ui/common/mapper/ValidationErrorStringMapper.kt
+++ b/feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/ui/common/mapper/ValidationErrorStringMapper.kt
@@ -57,6 +57,10 @@ private fun ValidatePasswordError.toPasswordErrorString(resources: Resources): S
         ValidatePasswordError.EmptyPassword -> resources.getString(
             R.string.account_server_settings_validation_error_password_required,
         )
+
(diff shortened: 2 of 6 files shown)


feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/domain/usecase/ValidatePassword.kt
```diff
@@ -10,12 +10,17 @@ class ValidatePassword : UseCase.ValidatePassword {
...
         }
     }
 
     sealed interface ValidatePasswordError : ValidationError {
         data object EmptyPassword : ValidatePasswordError
+        data object LinebreakInPassword : ValidatePasswordError
+    }
+
+    private companion object {
+        const val LINE_BREAK = "\n"
```
williamdes, line 24: Is is possible to have a password that contains \r followed by a space for example?
I am unsure anyone could even copy paste this 🤔

wmontwe: I tried that but it seems to be converted to "\n".

wmontwe:
> The input field is limited to a single line. That makes editing text containing line breaks almost impossible.
> 
> I think it would make for a better user experience if the app removed all instances of `\r` and `\n` when text is pasted to the input field. That way text in the clipboard that accidentally includes line breaks at the start or end will just work as expected without the app having to bother the user to manually fix the problem.

That's not a good idea at the current state. The used `TextField` requires some effort to filter text on the fly and keep it in sync with it's visual representation. I would recommend doing that after `TextField2` is available. Especially as this ia a rare corner case. I think users would like to be informed that their password contains unwanted linebreaks. Especially as that could lead to all kind of undesired side effects. As opposed to just doing some magic and at other places it still fails.

PR #7764, opened 2024-04-19
Enable emailaddress parameter for autoconfig

Elsensee:
This matches the behaviour of Thunderbird on desktop.

Also, some ISPs rely on the parameter for returning the correct username.


ByteHamster:
This is deliberately disabled, see https://forum.k9mail.app/t/autoconfig-url-emailaddress-not-supported/8336

PR #7771, opened 2024-04-22
Remove Material 2 text atoms

wmontwe:
Remove remaining Material 2 text atoms and replace them by their Material 3 counterpart.

Adds Material 3 text atoms to catalog app and drops Material 2 dependency from the design system.

Depends on #7769


cketti (approved):
:+1:

PR #7775, opened 2024-04-23
Change features to theme 2 - part 1

wmontwe:
This changes the features to Theme 2 and moves previews to the `debug` build type.

Removes the `SuccessItem` as it is not in use anymore.

Existing fake implementation have been moved to `debug` if needed for previews. When used by tests they moved to `test` and to avoid complicated dependencies they could be doubled. In the future we could look into automatic fake generation, which would ease the use further.


cketti (approved):
LGTM

PR #7777, opened 2024-04-23
Change features to theme 2 - part 2

wmontwe:
Part 2 of changing the features to Theme 2 and move previews to the debug build type.

Existing fake implementation have been moved to debug if needed for previews. When used by tests they moved to test and to avoid complicated dependencies they could be doubled. In the future we could look into automatic fake generation, which would ease the use further.


cketti (approved):
LGTM

PR #7798, opened 2024-04-29
Change K9 theme to 0.2.0

wmontwe:
This changes the tonal colors to v0.2.0 of the K9 theme.

Changes:
- `StatusBar` and `NavigationBar` colors to use `surfaceContainer` or on old Android versions `surfaceInverse` for the light theme, when no dark icons available
- Change `Toolbar` to `surfaceContainer`
- Remove `MessageList` special handling for `StatusBar` and `Drawer` overlay to streamline the UI. Design is working on a Drawer overhaul that would affect how we continue here.
- Small style changes to `MessageCompose`

The changes are minimal. But looking at the current implementation further adjustments are needed. Especially when we change behavior. Design will do a screen by screen check and propose changes.


cketti (approved):
LGTM

PR #7821, opened 2024-05-08
Change Compose design system to use outlined icons

wmontwe:
The icons in the Compose design system are changed to their outlined version.


cketti (approved):
LGTM

PR #7828, opened 2024-05-08
Change more icons to Material 3 outlined versions

wmontwe:
Second part of the Material 3 migration to outlined icons.

There are more icons to change, that will be done in a follow up pull-request.


app/ui/legacy/src/main/java/com/fsck/k9/ui/account/AccountFallbackImageProvider.kt
```diff
@@ -5,14 +5,14 @@ import android.graphics.drawable.Drawable
...
-import com.fsck.k9.ui.R
+import app.k9mail.core.ui.legacy.designsystem.atom.icon.Icons
 
 /**
  * Provides a [Drawable] for the account using the account's color as background color.
  */
 class AccountFallbackImageProvider(private val context: Context) {
     fun getDrawable(color: Int): Drawable {
-        val drawable = ContextCompat.getDrawable(context, R.drawable.drawer_account_fallback)
+        val drawable = ContextCompat.getDrawable(context, Icons.Outlined.Person)
```
cketti, line 15: The icon doesn't work well in a circular image view. It looks like there should be additional padding and/or the content should be moved up a bit.

For the drawable to work with Material Drawer's `ProfileDrawerItem` it seems like the width and height need to be at least 56dp.

When using `Icons.Outlined.Person` the account switcher looks like this:
image

I suggest we keep the old drawable and defer changing this to after we have a new design for the drawer.

app/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListAdapter.kt
```diff
@@ -51,10 +51,10 @@ class MessageListAdapter internal constructor(
...
 ) : RecyclerView.Adapter<MessageListViewHolder>() {
 
-    private val forwardedIcon: Drawable = ResourcesCompat.getDrawable(res, R.drawable.ic_messagelist_forwarded, theme)!!
-    private val answeredIcon: Drawable = ResourcesCompat.getDrawable(res, R.drawable.ic_messagelist_answered, theme)!!
-    private val forwardedAnsweredIcon: Drawable = ResourcesCompat
-        .getDrawable(res, R.drawable.ic_messagelist_answered_forwarded, theme)!!
+    private val forwardedIcon: Drawable = ResourcesCompat.getDrawable(res, Icons.Outlined.Forward, theme)!!
+    private val answeredIcon: Drawable = ResourcesCompat.getDrawable(res, Icons.Outlined.Reply, theme)!!
+    private val forwardedAnsweredIcon: Drawable =
+        ResourcesCompat.getDrawable(res, Icons.Outlined.CompareArrows, theme)!!
```
cketti, line 57: These are not regular icons. They were (somewhat) optimized for being displayed in line with text in the message list.

|Before|After|
|-|-|
|image|image|

I think it would be best to keep the old icons for now and ask the design team to come up with new icons (and a color to use).

app/ui/legacy/src/main/java/com/fsck/k9/ui/account/AccountFallbackImageProvider.kt
wmontwe: Not ideal, but I fixed it for now.

app/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListAdapter.kt
wmontwe: They should be colored using a tint instead of coloring the icon directly as done previously. The icons themselfs need design another discussion. I'll undo the changes until there is a good replacement.

wmontwe: I had a look at the latest gmail app and they use the new icons in the detail screen and the old ones in the list. But the tint is the same as `onSurface`. I created a patch to align the icon.

grafik

cketti (approved):

PR #7871, opened 2024-05-27
Fix message list menu order

### LINKED ISSUE #7869: Selected Email Action Icon order different than Action Icon order when viewing Email
### Checklist

### App version

6.901

Google Play

### Android version

14

### Device model

SM-S911U1 (Galaxy S23)

### Steps to reproduce

By default when you select an email (or multiple emails) from the message list, you are presented with three actions in this order: trash, mark unread, and archive. When you view a message in the message reader, the default order is: archive, trash and mark unread.

These should be the same as it is hard to develop muscle memory considering they are in the same place and the same set of actions. I have definitely deleted messages I meant to archive and vice-versa because of the order change.

### Expected behavior

Make the default order the same.

### Actual behavior

Order is not the same, resulting in taking the wrong action based on muscle memory.

### Logs

_No response_

wmontwe:
Fixes #7869


cketti (approved):
Please use a more descriptive PR title.

PR #7872, opened 2024-05-27
Fix message list context menu color

wmontwe:
The message list uses action mode to create a toolbar overlay. This needs to be explicitly styled to pick up our Material 3 colors. Furthermore, the back icon is replaced with a close icon.


cketti:
|Before|After|
|-|-|
|image|image|

cketti (approved):

PR #7877, opened 2024-05-28
Change message list to material 3 part 1

wmontwe:
This removes customizations that don't work well with Material 3.


cketti:
I can see that we need to change the approach to properly support this functionality with Material 3. Removing the functionality now and adding it back using a new approach later runs the risk of simply forgetting about it. That's why I'd much prefer a pull request that replaces the mechanism, not remove it.

I think having different text colors for read and unread messages is an important part of the user experience. Especially in dark mode.

|Before|After|
|-|-|
|k9mail__message_list__different_text_colors|k9mail__message_list__same_text_color|
|k9mail__message_list_dark__different_text_colors|k9mail__message_list_dark__same_text_colors|

cketti (approved):
From a technical point of view the patch looks good to me.

However, I think it's a UX regression that should be fixed as soon as possible if this is merged as is.

wmontwe:
The UX will have some regressions as there is no direct replacement for all the customization done in the existing theme.

There are also a couple of XML layouts that either don't use Material components or are even applying AppCompat styles still. Will take some time to remove all of that to get a clear view on all the little things that are not working.

PR #7903, opened 2024-06-04
Add String resource mover CLI script

wmontwe:
This adds a cli script to move String declarations from one module to another. This helps to strip the legacy code base or do some refactoring that includes touching string resources.

Usage:

```bash
./scripts/resource-mover --from <source-module-path> --to <target-module-path> --keys <keys-to-move>
```

This is a preparation for the app name provider implementation, that requires string resources to be moved.

### DIFF
--- a/cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/ResourceMoverCli.kt
+++ b/cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/ResourceMoverCli.kt
@@ -1,0 +1,29 @@
+package net.thunderbird.cli.resource.mover
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.options.split
+
+class ResourceMoverCli(
+    private val stringResourceMover: StringResourceMover = StringResourceMover(),
+) : CliktCommand(
+    name = "resource-mover",
+    help = "Move string resources from one file to another",
+) {
+    private val from: String by option(
+        help = "Source module path",
+    ).required()
+
+    private val to: String by option(
+        help = "Target module path",
+    ).required()
+
+    private val keys: List<String> by option(
+        help = "Keys to move",
+    ).split(",").required()
+
+    override fun run() {
+        stringResourceMover.moveKeys(from, to, keys)
+    }
+}
--- a/cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/StringResourceMover.kt
+++ b/cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/StringResourceMover.kt
@@ -1,0 +1,186 @@
+package net.thunderbird.cli.resource.mover
+
+import java.io.File
+import kotlin.system.exitProcess
+
+@Suppress("TooManyFunctions")
+class StringResourceMover {
+
+    fun moveKeys(source: String, target: String, keys: List<String>) {
+        val sourcePath = File(source + RESOURCE_PATH)
+        val targetPath = File(target + RESOURCE_PATH)
+
+        if (!sourcePath.exists()) {
+            println("\nSource path does not exist: $sourcePath\n")
+            return
+        }
+
+        println("\nMoving keys $keys")
+        println("       from \"$sourcePath\" -> \"$targetPath\"\n")
+        for (key in keys) {
+            moveKey(sourcePath, targetPath, key)
+        }
+    }
+
+    private fun moveKey(sourcePath: File, targetPath: File, key: String) {
+        println("\nMoving key: $key\n")
+
+        sourcePath.walk()
+            .filter { it.name.startsWith(VALUES_PATH) }
+            .forEach { sourceDir ->
+                val sourceFile = sourceDir.resolve(STRING_RESOURCE_FILE_NAME)
+                if (sourceFile.exists()) {
+                    moveKeyDeclaration(sourceFile, targetPath, key)
+                }
+            }
+    }
+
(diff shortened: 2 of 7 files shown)


cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/ResourceMoverCli.kt
```diff
@@ -0,0 +1,32 @@
...
+
+class ResourceMoverCli(
+    private val stringResourceMover: StringResourceMover = StringResourceMover(),
+) : CliktCommand(
+    name = "resource-mover",
+    help = "Move string resources from one file to another",
+) {
+    private val from: String by option(
+        help = "Source module path",
+    ).required()
```
cketti, line 15: clikt could check most of the preconditions.

```suggestion
    private val from: File by option(help = "Source module path")
        .file(mustExist = true, canBeFile = false)
        .required()
```

cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/ResourceMoverCli.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    private val to: String by option(
+        help = "Target module path",
+    ).required()
```
cketti, line 19:
```suggestion
    private val to: File by option(help = "Target module path")
        .file(mustExist = false, canBeFile = false)
        .required()
```

cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/ResourceMoverCli.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    private val keys: String by option(
+        help = "Keys to move",
+    ).required()
```
cketti, line 23:
```suggestion
    private val keys: List<String> by option(help = "Keys to move")
        .split(",")
        .required()
```

cketti (approved):
The fact that source and target files are read multiple times for each key is making me a bit sad. But this is not mission critical code. It seems to work and performance is not an issue.

wmontwe:
> The fact that source and target files are read multiple times for each key is making me a bit sad. But this is not mission critical code. It seems to work and performance is not an issue.

What about skipping such comments? This script is intended to help with refactoring the project, not to be a high-performance tool and neither to win any prizes. I'm prioritizing getting a job done and simplicity over performance concern and premature optimizations. Especially looking at the code base, there are bigger fishes to fry first.

cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/ResourceMoverCli.kt
wmontwe: Thanks for deep diving into clikt's options, but I don't see a benefit doing so.

cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/ResourceMoverCli.kt
wmontwe: Thanks for deep diving into clikt's options, but I don't see a benefit doing so.

cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/ResourceMoverCli.kt
wmontwe: Thanks, this looks helpfull.

cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/StringResourceMover.kt
```diff
@@ -0,0 +1,186 @@
...
+        targetFile.writeText(TARGET_FILE_CONTENT)
+        println("Target file ${targetFile.path} created")
+    }
+
+    private fun printError(message: String) {
+        System.err.println("\n$message\n")
+    }
+
+    private companion object {
+        const val RESOURCE_PATH = "/src/main/res/"
```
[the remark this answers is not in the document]
wmontwe: I understand your concern about hardcoded path separators, but given our core team uses Unix-based systems, I think it's reasonable to prioritize compatibility with those platforms. This refactoring tool is intended for internal use and not designed for external contributions. 

If we want to enable contributions from Windows systems, we should consider validating the project's compatibility on that platform first. This would require having a Windows system available, which could be addressed in a separate issue with discussion on the necessary resources and effort required.

cli/resource-mover-cli/src/main/kotlin/net/thunderbird/cli/resource/mover/StringResourceMover.kt
```diff
@@ -0,0 +1,186 @@
...
+        val targetPath = File(target + RESOURCE_PATH)
+
+        if (!sourcePath.exists()) {
+            println("\nSource path does not exist: $sourcePath\n")
+            return
+        }
+
+        println("\nMoving keys $keys")
+        println("       from \"$sourcePath\" -> \"$targetPath\"\n")
+        for (key in keys.sorted()) {
```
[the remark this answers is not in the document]
wmontwe: We could keep the order in which the keys have been entered. This would allow reorganisation when executing this script.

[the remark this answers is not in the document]
cketti: Sorry if I was unclear. That's what I meant. Use the order specified on the command line (not "retain the original order in the source file"). That way entries can be reordered when moving, but they don't have to (if the order of keys on the command line matches the order in the source file).

Basically drop the `.sorted()`.

PR #7905, opened 2024-06-04
Fix shortcut widget

wmontwe:


cketti:
I updated the branch to use `ShortcutManagerCompat.createShortcutResultIntent()`. This seems to make shortcuts work on devices where they previously didn't.

However, the shortcuts now use the app icon to indicate which app the shortcut is coming from. So our shortcuts look like this right now:

image

I think we should use the account image instead. However, this is probably a larger effort. So I created issue #7906 to keep track of the enhancement.

Tested on:
- Android 14, Pixel 8
- Android 14 (CalyxOS), Pixel 3
- Android 13, Fairphone 4
- Android 14 (Google APIs), Emulator
- Android 10 (Google APIs), Emulator
- Android 5.0, Emulator

cketti:
@wmontwe: GitHub doesn't let me add you as a reviewer. Presumably because you've created the PR. Can you please have a look at my changes.

wmontwe:
@cketti I had a quick check and it doesn't seem to work on my Android 12 device.

cketti:
Thanks for checking. I'm downloading an Android 12 image for the emulator right now and will have a look. 

Does the old code without this change work on your Android 12 device? What flavor of Android is it running?

wmontwe:
The old code works on my device: Samsung Android 12 tested with OneUi Launcher and Nova Launcher.

cketti:
I wasn't able to reproduce this. API 31 emulator images worked fine for me. But using the old method on API 31 and older and the new method on API 32 and newer also worked in my tests. So I created PR #7907 using this approach.

wmontwe:
Closing this in favor of #7907

Issue #7924, opened 2024-06-11
Add app name provider


wmontwe:
As part of #7807 we need a mechanism to dynamically retrieve the localized application’s name depending on the build application, specifically for K-9 Mail or Thunderbird for Android. The app name should be injectable into string resources using XLIFF placeholders to allow localization.

#### Components:

- **AppNameProvider Interface:** Interface to abstract the dynamic app name retrieval.
- **AppNameProvider Implementation:** Concrete implementations of the AppNameProvider for K-9 Mail (K9) and Thunderbird for Android (Tb).
- **XLIFF Placeholder:** Defined XLIFF placeholder to inject the app name into string resources.
- **Dependency injection:** A Koin module that sets up the dependency injection.

### Tasks

- [x] Implement AppNameProvider #7923 
- [x] Replace existing use of app names

PR #7942, opened 2024-06-13
Replace app name usage - Part 1

wmontwe:
Follow up on #7924. The hard coded app name is replaced by using the new `AppNameProvider`, part 1.


cketti (approved):
There's a lot of passing `appNameProvider` or `appName` along. We could reduce this by only using Koin to inject `AppNameProvider` in places where it's actually used. The downside is that we'd need to configure Koin for all previews that display one of those composables. But we could make this part of e.g. `PreviewWithTheme` (using `koinPreview`). That way we also wouldn't have to touch many previews.

PR #8022, opened 2024-07-23
Change K9-Mail signing to new setup

wmontwe:
This changes the signing configuration to allow easier addition of different signing configs for our different app variants. Keystores are now located in the `.signing` directory, accompanied by `signing.properties` files that provide the necessary keys. This avoids having to declare global properties for all the different variants.

This is the first iteration to change signing for K-9 Mail. Thunderbird for Android will follow shortly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,10 @@ out/
 *.jks
 *.keystore
 
+# Signing files
+.signing/
+*.signing.properties
+
 # IDEA/Android Studio ignores
 *iml
 .idea/*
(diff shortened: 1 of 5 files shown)


cketti (approved):
LGTM :+1:

I remember that F-Droid's build script is trying to strip signing configs, but don't remember how it does that exactly. Please make sure to test that this change works with F-Droid.

wmontwe:
It strips the signing config from the build script:

- `^[\t ]*signingConfig\s*[= ]\s*[^ ]*$` matches `signingConfig = signingConfigs.getByType(SigningType.K9_RELEASE)`

PR #8029, opened 2024-07-30
Change OAuth configuration

wmontwe:
This moves the OAuth configuration from the gradle script to the factory implementation. OAuth setup for AOL, Gmail, Office365 and Yahoo are added to Thunderbird. See #8030

Added a note about propper OAuth setup for forks of this project.


cketti (approved):
LGTM

PR #8054, opened 2024-08-06
Split legacy - part 3

wmontwe:
Part 3 of the legacy split, see previous pull-request: #8051


kewisch (approved):
I guess you found out what coreResourceProvider is for? I am curious :)

wmontwe:
> I guess you found out what coreResourceProvider is for? I am curious :)

From my current understand it looks like a workaround to provide translatable strings accross multiple modules.

PR #8056, opened 2024-08-07
Split legacy - part 4

wmontwe:
Last Part 4 of the legacy split, see previous pull-request: #8054


legacy/core/src/main/java/com/fsck/k9/controller/DefaultMessageCountsProvider.kt
```diff
@@ -47,6 +41,7 @@ internal class DefaultMessageCountsProvider(
         return MessageCounts(unreadCount, starredCount)
     }
 
+    @Suppress("TooGenericExceptionCaught")
```
kewisch, line 44: Do we have a strategy to remove some of these suppressions in the future? Maybe some good first bugs to file?

kewisch (approved):

legacy/core/src/main/java/com/fsck/k9/controller/DefaultMessageCountsProvider.kt
wmontwe: This specific warning is generated by our detekt code quality tool, which is desired to be reported in general. But there are cased where it is wrong or it's hard to provide the exact exception. Especially as our legacy code is designed to rely on `try/catch` constructs quite a lot. This is something we need to address when we touch that code base, but it's not easy to do.

Issue #7807, opened 2024-05-03
[EPIC] White Label Architecture


wmontwe:
We will be implementing a modular white-label architecture to improve the development process for our Thunderbird for Android and K-9 Mail applications. This new structure will allow us to configure brand-specific details either during build or at runtime while maintaining a consistent core functionality. The proposed (#7805) architecture consists of application modules and a common module as follows:

**Application Modules:**

1. Thunderbird for Android: `app-thunderbird`
2. K-9 Mail: `app-k9mail`

These modules will contain app-specific implementations, configurations, resources, and startup logic. They will only depend on the `app-common` module for shared functionalities and selectively integrate other modules as needed to configure app-specific functionality.

**App Common Module:**

1. `app-common`: This module will act as the central integration point for shared code among the applications. It will contain the core functionality, shared resources, and configurations that are common to both apps.

### Tasks

- [x] Create `app-thunderbird` and `app-k9mail` modules for application-specific code and configurations and include them in the project structure
  - #7494
  - #7779
- [x] Create a `app-common` module containing shared functionality, resources, and configurations for both applications. #8000
- [x] Create an app name provider #7924
- [ ] Migrate code from the existing structure to the new modules while ensuring minimal disruption to ongoing development.
- [x] Migrate app widgets to work with both apps #7933
- [ ] Provide necessary documentation and resources for developers to help them get started with the new architecture.
- [x] Test the new architecture thoroughly to ensure it meets our performance, scalability, and security requirements.

kewisch:
@wmontwe wondering if we can close this one, I see there are two tasks left. If they still need to be done let's maybe resolve them for after the release or decide against them? This would help with closure.

Issue #6617, opened 2023-02-01
[EPIC] Material You/Material 3 Design


seniorm0ment:
Would be very nice to see the app with a more modern feel.
My suggestion is to bring it up to par with Material You design.

Also in the actual screen that displays the mail you are reading ,that screen feels very inconsistent and outdated to the other menus already. Especially if viewed as plaintext only.

Wave6677:
A material you coat of paint over K-9 Mail would be lovely, especially with the Thunderbird integration coming up

MXC48:
yes, this will allow for better integration into the Android system

joshuajeschek:
This would be so great!

rezazarchi:
It should be noted that Material 3 is dependent on migration of the entire UI to Jetpack Compose, which constitutes a substantial amount of work. It seems that this constitutes a long-term plan for K-9, although it is acknowledged that there are some tasks involving rewriting some parts of the application (#6974, #6975).

cketti:
What makes you think the entire UI needs to be migrated to Jetpack Compose?

rezazarchi:
> What makes you think the entire UI needs to be migrated to Jetpack Compose?

I recently came across information in Google documents indicating that Material 3 can be applied with Jetpack Compose using the MaterialTheme but didn't find any official information about Material 3 in the XML way. some parameters like dynamic color scheme and other modern Material You UI components are available in compose library.

W1zardK1ng:
waiting for the Material You version of k9 mail.

Cwpute:
I'd like to say i'm not interested in a Material You design for this app, as Material You uses more space on the screen for the same amount of information.
For users like me who have a phone with a not so big screen, this is not good news.

flexagoon:
@Cwpute k9 has a compact view, and nothing in M3 says you can't have a compact view in a M3 app

tamirat67:
Hello dear certification issue not possible to me, it say warning ⚠️
invalid certificate ohhh

cketti:
> Hello dear certification issue not possible to me, it say warning ⚠️ invalid certificate ohhh

@tamirat67: Please use the [forum](https://forum.k9mail.app/) for support requests.

S0ulf3re:
I would also love to see material you as well

rezazarchi:
> I would also love to see material you as well

[...]

PR #8060, opened 2024-08-08
Remove legacy UI theme

wmontwe:
Part of the Material 3 #6617 and White Label Architecture #7807

This removes the remains of the old theming setup and moves it to app module level. Now app themes are truly separated.

To resolve issues with the attribute declarations, a helper theme `Theme.Legacy.Test` has been added.


cketti (approved):
LGTM

Issue #8059, opened 2024-08-08
Update Material 3 color sheme


wmontwe:
Design created an updated Material 3 color scheme for K-9 Mail and Thunderbird for Android. This should improve the overall look and contrasts of the apps. While the color changes fix some of the existing issues, there are still some areas that require further refinement.

Part of: #6617

PR #8062, opened 2024-08-08
Change Material 3 color scheme

wmontwe:
This applies the new color schemes for K-9 Mail and Thunderbird for Android.

Solves #8059 and depends on #8060


cketti (approved):
LGTM

PR #8067, opened 2024-08-14
Fix Thunderbird app icons

wmontwe:
Change to mipmap app icons as the app icon requires gradients which are only support from API 24+


cketti (approved):
LGTM

Issue #8118, opened 2024-09-06
Add the drawer account folder list


wmontwe:
Implement the folder list highlighted in this design:

![FolderDrawer_01](https://github.com/user-attachments/assets/454e297d-c991-45be-a521-9a647b1a8a84)

PR #8131, opened 2024-09-10
Add drawer folder list - Part 2

wmontwe:
Second part of the drawer folder list implementation #8118. The direct dependencies, like `AccountsViewModel` and `FoldersViewModel`, have been moved to the feature module. The UI related functionality has been separated from the `FolderRepository` and moved into `DisplayFolderRepository`, to clarify responsibilities. Related data models have been moved out of legacy to where they are actually used.

This introduces a new `:core:mail:folder:api` module for folder specific declarations.


wmontwe:
> LGTM
> 
> Side note: The classes in a feature module that are not part of the exposed "API" could use the `internal` visibility modifier.

Right, I'll do that in a follow up pull-request

PR #8153, opened 2024-09-19
Add basic feature setup for funding and Google Play billing support

### LINKED ISSUE #8152: Add payment permission to Thunderbird Daily app
To be able to start settting up the Google Play Store for funding, we need to upload an app with the billing library added. We do this for the Thunderbird Daily as a first step.

wmontwe:
This sets up a feature for funding support and adds no-op and googleplay support. Implementation is only an on-off flag and the Google Play billing library dependency enabled for Thunderbird Daily.

Resolves  #8152

### DIFF
--- a/app-k9mail/build.gradle.kts
+++ b/app-k9mail/build.gradle.kts
@@ -24,6 +24,7 @@ dependencies {
     implementation(projects.feature.widget.shortcut)
     implementation(projects.feature.widget.unread)
     implementation(projects.feature.telemetry.noop)
+    implementation(projects.feature.funding.noop)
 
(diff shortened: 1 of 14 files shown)


kewisch (approved):

kewisch:
Would we need to add `<uses-permission android:name="com.android.vending.BILLING" />` somewhere?

wmontwe:
> Would we need to add `<uses-permission android:name="com.android.vending.BILLING" />` somewhere?

If we package the library with the app it automatically adds the billing permission, so it's available for TB Daily.

PR #8166, opened 2024-09-24
Change drawer sync calls

### LINKED ISSUE #8146: Improve drawer sync mechanism
During review of #8143 it was stated that the `SyncMail` use case could be used wrongly and needs improvement. 

As the `SyncMail` implementation is basically a copy paste from the existing code, it does not distinguish between syncing a single account vs all accounts and at what point in time either of them is appropriate to use. 

It would be good to improve the api to minimize misunderstandings in the future. Take note that the underlying `MessagingController.checkMail()` could benefit from such an improvement too.

See [comment](https://github.com/thunderbird/thunderbird-android/pull/8143/files/29bada814b06f528cbc9c4964af8b2503ca9a5af#r1761876643) for more details.

wmontwe:
Resolves #8146

Add a syn all accounts button to the side rail.

Depends on #8165


cketti:
image

It would be nice if the icons aligned with the "Manage folders" and "Hide accounts" icons. I confirmed with layout inspector (`build.gradle.kts` has to be changed to not exclude `"META-INF/*.version"`) that it's close, but not quite the same y-coordinates.

cketti (approved):

wmontwe:
This is an layouting issue as the column is offset by 1 dp caused by the devider. I added an uneven spacer to compensate this. The LazyColumn is not affected by that.

cketti:
I don't think the divider is the issue. 

The `SettingListItem`s in the drawer are 56dp in height by default. The icon is vertically centered.
To horizontally align the icons, the `SettingItem` should have the same height and vertically center the icons. 

This will fail to align the icons when the text in `SettingItem` wraps to multiple lines. But it should work for the default case where it's just one line.

wmontwe:
The outer column is placed at an uneven y position and that's causing the alignment issue.

cketti:
Both views are aligned to the bottom. It's just a matter of replicating the layout behavior. See https://github.com/cketti/k-9/commit/95ac9eda4b178578602b7003dee4878f5861f518

PR #8171, opened 2024-09-25
Add issues permissions for handling needinfo requests

kewisch:
I see the needinfo job is failing due to missing permissions. I've not needed any permissions for this action before, but possibly they are locked down a bit due to recent releng work. This adds the issues permission. It is a stab in the dark, I don't know if it will work, but it is the area of all octokit API client calls that occurs in the action.

https://github.com/thunderbird/thunderbird-android/actions/runs/11004526905/job/30665933452

### DIFF
--- a/.github/workflows/needinfo-remove.yml
+++ b/.github/workflows/needinfo-remove.yml
@@ -12,6 +12,9 @@ jobs:
     if: |
       github.event.comment.author_association != 'OWNER' &&
       github.event.comment.author_association != 'COLLABORATOR'
+    permissions:
+      issues: write
+      pull-requests: write
     steps:
       - name: Remove needinfo label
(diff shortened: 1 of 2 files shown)


wmontwe:
Looking at the [GitHub Issues Api](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-issues), it requires `issues:write` permission. I'm not aware which exact operation are used by the action, but some of them might need multiple permissions.

wmontwe (approved):

kewisch:
@wmontwe I also added the pull-requests permission now, the API endpoints used say they require that permission as well. Let me know if that causes trouble.

wmontwe:
Fine with me

PR #8172, opened 2024-09-26
Fix a11y warnings in accounts and settings

kewisch:
I believe this fixes the a11y warnings in the play console. I suspect there may be an easier way to merge checkbox and text field, simply using mergeDescendants on its own didn't do it though.

### DIFF
--- a/core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/molecule/input/CheckboxInput.kt
+++ b/core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/molecule/input/CheckboxInput.kt
@@ -1,13 +1,14 @@
 package app.k9mail.core.ui.compose.designsystem.molecule.input
 
-import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.selection.toggleable
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.Role
 import app.k9mail.core.ui.compose.designsystem.atom.Checkbox
 import app.k9mail.core.ui.compose.designsystem.atom.text.TextBodyLarge
 import app.k9mail.core.ui.compose.theme2.MainTheme
@@ -29,13 +30,17 @@ fun CheckboxInput(
         Row(
             modifier = Modifier
                 .fillMaxWidth()
-                .clickable { onCheckedChange(!checked) },
+                .toggleable(
+                    value = checked,
+                    role = Role.Checkbox,
+                    onValueChange = { onCheckedChange(!checked) },
+                ),
             verticalAlignment = Alignment.CenterVertically,
             horizontalArrangement = Arrangement.spacedBy(MainTheme.spacings.half),
         ) {
             Checkbox(
                 checked = checked,
-                onCheckedChange = onCheckedChange,
+                onCheckedChange = null,
             )
             TextBodyLarge(text = text)
         }
--- a/feature/settings/import/src/main/kotlin/app/k9mail/feature/settings/import/ui/SettingsImportScreen.kt
+++ b/feature/settings/import/src/main/kotlin/app/k9mail/feature/settings/import/ui/SettingsImportScreen.kt
@@ -31,6 +31,7 @@ fun SettingsImportScreen(
                     ButtonIcon(
                         onClick = onBack,
                         imageVector = Icons.Outlined.ArrowBack,
+                        contentDescription = stringResource(androidx.appcompat.R.string.abc_action_bar_up_description),
                     )
                 },
             )
--- a/core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/atom/Checkbox.kt
+++ b/core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/atom/Checkbox.kt
@@ -7,7 +7,7 @@ import androidx.compose.material3.Checkbox as Material3Checkbox
 @Composable
 fun Checkbox(
     checked: Boolean,
-    onCheckedChange: (Boolean) -> Unit,
+    onCheckedChange: ((Boolean) -> Unit)?,
     modifier: Modifier = Modifier,
     enabled: Boolean = true,
 ) {


feature/settings/import/src/main/kotlin/app/k9mail/feature/settings/import/ui/SettingsImportScreen.kt
```diff
@@ -31,6 +31,7 @@ fun SettingsImportScreen(
                     ButtonIcon(
                         onClick = onBack,
                         imageVector = Icons.Outlined.ArrowBack,
+                        contentDescription = stringResource(R.string.settings_import_back_button),
```
cketti, line 34: For consistency we should use the string that is (automatically) used in all other screens.
```suggestion
                        contentDescription = stringResource(androidx.appcompat.R.string.abc_action_bar_up_description),
```

core/ui/compose/designsystem/src/main/res/values/strings.xml
```diff
@@ -5,4 +5,6 @@
     <string name="designsystem_molecule_email_address_input_label">Email address</string>
     <string name="designsystem_molecule_password_input_label">Password</string>
     <string name="designsystem_molecule_error_view_button_retry">Retry</string>
+    <string name="designsystem_molecule_a11y_checked">%1$s, checked</string>
+    <string name="designsystem_molecule_a11y_unchecked">%1$s, not checked</string>
```
cketti, line 9: I think we can increase the accessibility of our code by avoiding abbreviations whenever possible :smile: 

In similar cases we are using the term "content description". Maybe `designsystem_molecule_checkbox_content_description_checked` and `designsystem_molecule_checkbox_content_description_unchecked`? (If we need this at all)

feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/ui/incoming/content/IncomingFormItems.kt
```diff
@@ -32,7 +33,11 @@ internal fun LazyListScope.incomingFormItems(
     resources: Resources,
 ) {
     item {
-        Spacer(modifier = Modifier.requiredHeight(MainTheme.sizes.smaller))
+        Spacer(
+            modifier = Modifier
+                .requiredHeight(MainTheme.sizes.smaller)
+                .clearAndSetSemantics { },
```
cketti, line 39: Is this really necessary? It feels odd that a warning would be generated for a `Spacer` with a missing(?) content description. If here, why not in the other places where a `Spacer` is used?

core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/molecule/input/CheckboxInput.kt
```diff
@@ -29,7 +42,11 @@ fun CheckboxInput(
         Row(
             modifier = Modifier
                 .fillMaxWidth()
-                .clickable { onCheckedChange(!checked) },
+                .clickable { onCheckedChange(!checked) }
+                .semantics(mergeDescendants = true) {}
+                .clearAndSetSemantics {
+                    contentDescription = combinedContentDescription
+                },
```
cketti, line 49: Looks like we could use `.toggleable(value = checked) { onCheckedChange(!checked) }` instead.

See https://stackoverflow.com/a/78059984

We should also use a `Checkbox` with `onCheckedChange` set to `null` so the checkbox is not individually selectable using Talkback.

cketti:
To automatically fix issues found by spotless, run `./gradlew spotlessApply`.

feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/ui/incoming/content/IncomingFormItems.kt
kewisch: I'm not completely sure, I might have targeted the wrong element. Here is the screenshot from the pre-launch report. It looks a bit like the server name field, but that one provides a text description?

image

feature/settings/import/src/main/kotlin/app/k9mail/feature/settings/import/ui/SettingsImportScreen.kt
kewisch: Ah nice, I was looking for this, I couldn't imagine there isn't already a string for it.

core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/molecule/input/CheckboxInput.kt
kewisch: Looks like this approach is successful, and if I use mergeDescendents it will also work well without a combined description.

kewisch: Hmm with this approach talkback reads "checked, detected icon, checkmark, use compression, detected icon, checkmark, checkbox."

I don't know where it is getting icon from, and why it is repeating checkmark/checkbox so often. I guess I'd need to clear semantics somewhere. 

I'll upload the new version in a bit as an intermediate.

core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/molecule/input/CheckboxInput.kt
```diff
@@ -29,13 +31,18 @@ fun CheckboxInput(
         Row(
             modifier = Modifier
                 .fillMaxWidth()
-                .clickable { onCheckedChange(!checked) },
+                .semantics(mergeDescendants = true) {}
```
cketti, line 34: This doesn't seem to be necessary for it to work on my test device. Does it make any difference for you?

core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/molecule/input/CheckboxInput.kt
kewisch: I'm going to categorize this as "better than before", we can do more work here with some targeted a11y work.

core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/molecule/input/CheckboxInput.kt
kewisch: It did at some point, but now I don't see a difference.

kewisch:
@cketti This should be ready for review now

cketti (approved):
LGTM

Issue #8162, opened 2024-09-24
Integrate Google Play Billing library


wmontwe:
[Integrate Google Play Billing Library](https://developer.android.com/google/play/billing/integrate#kts)

PR #8188, opened 2024-09-30
Add funding dependencies

wmontwe:
Add the Kotlin KTX dependency for the biiling library and KotlinX Serialization to support type safe navigation and update the bundled version to be compatible with Kotlin 2.0.20 as part of #8162.

Cleaned up the build-plugin and Gradle setup.


cketti (approved):
LGTM

Issue #8163, opened 2024-09-24
Implement subscription ui flow


wmontwe:
Design:

![Support -_ Choose amount v01 1](https://github.com/user-attachments/assets/bcf35dc3-ac5c-4ffa-a8d5-530918771509)

MillionsToOne:
Will this also be enabled for first time installs like the desktop version of Thunderbird does? And will it pop up regularly or will it just be there?

kewisch:
We haven't worked out the cadence yet, but the first time install might be a bit too early. Folks need to have a chance to use the app before they make this decision. On desktop it works because there is more screen space and the message does not get in the way.

PR #8212, opened 2024-10-01
Change FeatureLauncherActivity to launch targets

wmontwe:
Change FeatureLauncherActivity to depend on a target for managing launches.

Part of #8163


cketti (approved):
What problem is this trying to solve? To me it looks like more ceremony without any clear benefit.

That being said, I don't see any harm in merging this either.

wmontwe:
Easier setup of new routes, see next pull-request

PR #8216, opened 2024-10-01
Don't remove needinfo label for MEMBER authors

kewisch:
This is driving me crazy. Apparently MEMBER is missing, Github sees me as a member of the org rather than an owner or collaborator, despite direct access. So let's not immediately remove labels when members take action.


wmontwe (approved):
LGTM

PR #8337, opened 2024-10-15
Add releaser keys to automation setup and fix formatting

kewisch:
Applied black formatting and added releaser key setup


kewisch:
I'm going to go ahead and merge this so we can get the K-9 build going, please do review after the fact and I am happy to make changes.

PR #8344, opened 2024-10-16
Release Automation Improvements

kewisch:
* Speed up printVersionInfo by using configure-on-demand
* Allow passing repository and branch to render-notes
* Make sure changelog files are not overwritten in render-notes (catches us using the wrong version code)
* Print-only mode for local testing with render-notes
* Always release to internal testing track so we can promote from there


wmontwe (approved):
👍

coreycb (approved):

PR #8350, opened 2024-10-17
Fix color roles calculation

wmontwe:
Debug variant includes the Material 3 library, but release not. I moved the `ColorRoles` to the `theme` module where it has access to the `MaterialColors` class for all variants.


cketti (approved):
LGTM

PR #8374, opened 2024-10-20
Render notes with long and short form

coreycb:


coreycb:
Tested with:
python ./scripts/render-notes.py -r "coreycb/thunderbird-notes" -p changelog -- ${APPLICATION_ID} ${FULL_VERSION_NAME} ${VERSION_CODE} ${GITHUB_NOTES_FILE} | tee -a $GITHUB_STEP_SUMMARY

coreycb:
@kewisch This is ready for review.

.github/workflows/shippable_builds.yml
```diff
@@ -204,17 +205,25 @@ jobs:
...
           mkdir -p ./app-metadata/${APPLICATION_ID}/en-US/changelogs
-          python ./scripts/render-notes.py ${APPLICATION_ID} ${FULL_VERSION_NAME} ${VERSION_CODE} | tee -a $GITHUB_STEP_SUMMARY
+          GITHUB_NOTES_FILE="$(mktemp -d)/long-notes.txt"
+          python ./scripts/render-notes.py -p changelog -- ${APPLICATION_ID} ${FULL_VERSION_NAME} ${VERSION_CODE} ${GITHUB_NOTES_FILE} | tee -a $GITHUB_STEP_SUMMARY
           echo "</pre>" | tee -a $GITHUB_STEP_SUMMARY
+          if [[ "$APP_NAME" == "thunderbird" ]]; then
+            echo "thunderbird_github_notes<<EOF" >> $GITHUB_OUTPUT
+            cat $GITHUB_NOTES_FILE >> $GITHUB_OUTPUT
+            echo "EOF" >> $GITHUB_OUTPUT
+          fi
```
kewisch, line 226: Both K-9 Mail and Thunderbird for Android have GitHub Releases, so they both have long form notes. There are a few items that will be in the Thunderbird release notes that won't be in the K-9 Mail release notes, I don't think we talked about how to differentiate in the notes yaml.

.github/workflows/shippable_builds.yml
```diff
@@ -572,6 +581,21 @@ jobs:
...
+          APPLICATION_LABEL: ${{ steps.pkginfo.outputs.APPLICATION_LABEL }}
+          FULL_VERSION_NAME: ${{ steps.pkginfo.outputs.FULL_VERSION_NAME }}
+          VERSION_CODE: ${{ steps.pkginfo.outputs.VERSION_CODE }}
+          THUNDERBIRD_GITHUB_NOTES: ${{ needs.release_commit.outputs.thunderbird_github_notes }}
+        run: |
+          if [[ "$APP_NAME" == "thunderbird" ]]; then
+            echo "<h2>${APPLICATION_LABEL} ${FULL_VERSION_NAME} GitHub Draft Release Notes (${VERSION_CODE})</h2><pre>" | tee $GITHUB_STEP_SUMMARY
+            echo "${THUNDERBIRD_GITHUB_NOTES}" | sed -z 's/\n\n/\n/g' | tee -a $GITHUB_STEP_SUMMARY
+            echo "</pre>" | tee -a $GITHUB_STEP_SUMMARY
+          fi
```
kewisch, line 597: For the step summary, let's rather show them early. We can do this together with the short form notes.

At this point in the process, we instead need to add them to the publish_gh step below instead so they get added to the github release.

scripts/render-notes.py
```diff
@@ -77,25 +94,35 @@ def render_notes(
...
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        "--print",
-        "-p",
+        "--print-only",
+        "-po",
         action="store_true",
         help="Only print the processed release notes",
     )
```
kewisch, line 118: This was mainly for local debugging, I think one of the arguments is sufficient.

scripts/render-notes.py
```diff
@@ -77,25 +94,35 @@ def render_notes(
...
+    parser.add_argument(
+        "--print",
+        "-p",
+        nargs="+",
+        default=[],
```
kewisch, line 123: Adding `choices` here would be good if we're using the identifiers.

.github/workflows/shippable_builds.yml
kewisch: What you'll need to do is extend the Get release sha step to output `app_release_notes` based on the appname, and then use this in the publish_gh step.

kewisch (changes requested):
Thanks for handing this in. I have a few comments that will need changes. Get in touch before you get started though, depending on how far along we are with the beta release I might have to work on this beforehand.

.github/workflows/shippable_builds.yml
coreycb: Ok. I wasn't sure if we were ready to publish the notes as they are rendered now, or if you wanted to just print them out for now and fine-tune manually.

kewisch:
Ported this to #8380

PR #8410, opened 2024-10-24
Encourage upgrading Thunderbird for QR code import

kewisch:
@cketti I give up, I can't believe it is so much extra work just to add a hyperlink. It looks like we'd need to have code that uses buildAnnotatedString with index positions to identify the hyperlink and then call openURI, which would make sense as a re-usable component.

Could you take a look at this one? I suspect it will be quick for you.

### DIFF
--- a/feature/onboarding/migration/thunderbird/build.gradle.kts
+++ b/feature/onboarding/migration/thunderbird/build.gradle.kts
@@ -13,5 +13,7 @@ dependencies {
     implementation(projects.core.ui.compose.designsystem)
     implementation(projects.feature.account.common)
 
+    implementation(libs.timber)
+
     testImplementation(projects.core.ui.compose.testing)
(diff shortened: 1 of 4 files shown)


cketti:
It's hard to make part of flowing text clickable while respecting the minimum size recommended to make it easy to tap using a finger. Presumably that's why there's no built-in support for that.

I'd recommend making the whole text clickable.

|Default|Finger down|
|-|-|
|image|image|

Would that work? I could update the PR accordingly.

kewisch:
Yes I was thinking the same, making the whole text clickable but underlining and making the learn more text look like a hyperlink would be great.

Appreciate you taking over the pull request.

cketti:
@kewisch: What's the URL for the link?

cketti:
image

PR #8411, opened 2024-10-25
Add support for daily build version codes

coreycb:


coreycb:
@kewisch I think this is pretty close. Can you take a look and let me know what you think? Things I can think of: 1) it may need a symlink update in build_unsigned after it pulls from git again, 2) Publish to Play Store step got skipped in my test run, 3) "Get release sha and notes" is getting garbage for the sha so I think I need to pass in the sha of HEAD.

I had to go with years since 2023 instead of years since 2024, because a leading zero caused failures. I think, for example, 02990040 might get interpreted as octal number.

.github/workflows/shippable_builds.yml
```diff
@@ -663,20 +703,23 @@ jobs:
...
+            echo "${!app_relnotes_name}" >> $GITHUB_OUTPUT
+            echo "EOF" >> $GITHUB_OUTPUT
+          fi
 
           cat $GITHUB_OUTPUT
 
       - name: Publish to GitHub Releases
         id: publish_gh
-        if: ${{ contains(matrix.releaseTarget, 'github') }}
+        if: "${{ contains(matrix.releaseTarget, 'github') && contains(fromJSON('[\"beta\", \"release\"]'), needs.dump_config.outputs.releaseType) }}"
```
kewisch, line 722: Daily targets won't be releasing to github, do we need the extra check here?

.github/workflows/shippable_builds.yml
```diff
@@ -697,11 +740,14 @@ jobs:
           APPLICATION_ID: ${{ steps.pkginfo.outputs.APPLICATION_ID }}
           REPO: ${{ github.repository }}
           APP_SHA: ${{ steps.shanotes.outputs.app_sha }}
+          RELEASE_TYPE: ${{ vars.RELEASE_TYPE }}
         run: |
           # r0adkll/upload-google-play expects the release notes in a different structure
-          FILEPATH=app-metadata/${APPLICATION_ID}/en-US/changelogs/${VERSION_CODE}.txt
           mkdir whatsnew
-          wget -O whatsnew/whatsnew-en-US https://raw.githubusercontent.com/${REPO}/${APP_SHA}/${FILEPATH}
+          if [[ "$RELEASE_TYPE" = "beta" || "$RELEASE_TYPE" = "release" ]]; then
```
kewisch, line 747: For daily we should have some release notes that indicate the sha it was built from, and a direct link to the workflow run. Any other info you think would be relevant? 

While you are here, can you add `?token=<timestamp here>` to the end of the URL? Otherwise github may do some caching as I noticed in render-notes

.github/workflows/shippable_builds.yml
```diff
@@ -697,11 +740,14 @@ jobs:
...
+            FILEPATH=app-metadata/${APPLICATION_ID}/en-US/changelogs/${VERSION_CODE}.txt
+            wget -O whatsnew/whatsnew-en-US https://raw.githubusercontent.com/${REPO}/${APP_SHA}/${FILEPATH}
+          fi
 
       - name: Publish to Google Play
         id: publish_play
```
kewisch, line 753: While you are here, could you adjust the release name for the play store upload to be `${{ versionCode }} (${{ versionName }})`? This makes it more consistent with the auto-generated name.

.github/workflows/shippable_builds.yml
```diff
@@ -409,9 +438,20 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
         with:
-          cache-disabled: ${{ needs.dump_config.outputs.releaseType == 'beta' || needs.dump_config.outputs.releaseType == 'release' }}
+          cache-disabled: "${{ contains(fromJSON('[\"daily\", \"beta\", \"release\"]'), needs.dump_config.outputs.releaseType) }}"
           add-job-summary: on-failure
 
+      - name: Get Version Code for Daily
```
kewisch, line 444: Get -> Set

.github/workflows/shippable_builds.yml
```diff
@@ -381,13 +405,18 @@ jobs:
...
           APP_NAME: ${{ matrix.appName }}
         run: |
-          case "${APP_NAME}" in
-            thunderbird) APP_SHA=$THUNDERBIRD_SHA ;;
-            k9mail) APP_SHA=$K9MAIL_SHA ;;
-            *) APP_SHA=$GITHUB_SHA ;;
-          esac
+          if [[ "$RELEASE_TYPE" = "daily" ]]; then
+            APP_SHA=$GITHUB_SHA
+          else
```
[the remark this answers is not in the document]
kewisch: Optionally you can also change this code to use the bash indirection like I did in the other location and missed here.

.github/workflows/shippable_builds.yml
```diff
@@ -224,21 +224,45 @@ jobs:
...
+            VERSION_FILE="$(pwd)/version.properties"
+            if [[ -e "$VERSION_FILE" ]]; then
+              while IFS="=" read -r key value; do
+                if [[ "${key}" = "NEW_VERSION_CODE" ]]; then
+                  NEW_VERSION_CODE="${value}"
+                fi
+              done < $VERSION_FILE
+            else
+              # Daily uses date-based calculation for version code in yDDDHHmm format.
+              # y: Years since 2023 (e.g. 2024 is 1)
```
kewisch, line 248: We can do 0 for 2024

.github/workflows/shippable_builds.yml
```diff
@@ -224,21 +224,45 @@ jobs:
...
+              # DDD: Day of the year (3 digits, zero-padded)
+              # HH: Hour of the day (2 digits, 00-23)
+              # mm: Minute of the hour (2 digits)
+              year=$(( $(date +"%Y") - 2023 ))
+              day=$(printf "%03d" $(date +"%j"))
```
kewisch, line 253: Is %j always zero padded? The docs seem to suggest so

.github/workflows/shippable_builds.yml
kewisch: If you chose 2023 to avoid needing to strip leading zeros, then I am fine using 2023 as the base.

.github/workflows/shippable_builds.yml
```diff
@@ -224,21 +224,45 @@ jobs:
...
+              time=$(date +"%H%M")
+              NEW_VERSION_CODE="${year}${day}${time}"
```
kewisch, line 255: Can you add a safety check here so that the final number is not higher than what we'd expect at the end of 2025? I want to avoid a future error accidentally burning the version code, if we reach 2100000000 then we are cooked and have to throw away the package id. Same check with beta/release would make sense, assuming monthly releases and weekly betas.

.github/workflows/shippable_builds.yml
```diff
@@ -224,21 +224,45 @@ jobs:
...
         env:
           APP_NAME: ${{ matrix.appName }}
           OLD_VERSION_CODE: ${{ steps.appinfo.outputs.VERSION_CODE }}
+          RELEASE_TYPE: ${{ vars.RELEASE_TYPE }}
         run: |
-          NEW_VERSION_CODE=$(($OLD_VERSION_CODE + 1))
+          if [[ "$RELEASE_TYPE" = "beta" || "$RELEASE_TYPE" = "release" ]]; then
+            NEW_VERSION_CODE=$(($OLD_VERSION_CODE + 1))
+          elif [[ "$RELEASE_TYPE" = "daily" ]]; then
+            VERSION_FILE="$(pwd)/version.properties"
```
kewisch, line 239: Does this require another PR? I don't think we have this file yet

kewisch (changes requested):

kewisch:
> @kewisch I think this is pretty close. Can you take a look and let me know what you think? Things I can think of: 1) it may need a symlink update in build_unsigned after it pulls from git again, 2) Publish to Play Store step got skipped in my test run, 3) "Get release sha and notes" is getting garbage for the sha so I think I need to pass in the sha of HEAD.
> 
> I had to go with years since 2023 instead of years since 2024, because a leading zero caused failures. I think, for example, 02990040 might get interpreted as octal number.

Sorry, missed this comment while reviewing, was on mobile. Going with 2023 is of course fine. The symlink update is only really necessary for f-droid, so we don't need to do that on daily. Play store publish, let's be careful on that and only do so when we know everything else works. If you have it enabled in your environment, you might want to comment out the play store upload target in setup_release_automation and update the thunderbird_daily environment.

Re head sha, you can get that from github.sha or GITHUB_SHA. In beta/release we need to use the commit sha, but for daily we can use the one the build was started on.

.github/workflows/shippable_builds.yml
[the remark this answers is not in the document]
coreycb: I didn't follow what you meant by `|| github.sha`.

Here's what I currently have:
```
        run: |
          APP_SHA=${APP_NAME^^}_SHA
          if [[ -z "$APP_SHA" ]]; then
            APP_SHA=$GITHUB_SHA
          fi
          echo "app_sha=$APP_SHA" >> $GITHUB_OUTPUT
```
          cat $GITHUB_OUTPUT

.github/workflows/shippable_builds.yml
coreycb: I'm only using that file in order to have parallel matrix jobs use the same version code, which may not be necessary.

.github/workflows/shippable_builds.yml
coreycb: Yes, looks good:
$ date -d "2025-12-31 23:59" +"%j"
365
$ date -d "2025-01-01 23:59" +"%j"
001

.github/workflows/shippable_builds.yml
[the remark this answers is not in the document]
coreycb: Updated to:
APP_SHA_NAME=${APP_NAME^^}_SHA
APP_SHA="${!APP_SHA_NAME:-$GITHUB_SHA}"
echo "app_sha=${APP_SHA}" >> $GITHUB_OUTPUT

coreycb:
I believe all comments have been addressed. This is ready to review again.

kewisch (changes requested):

.github/workflows/shippable_builds.yml
```diff
@@ -224,21 +225,56 @@ jobs:
...
+                fi
+              done < $VERSION_FILE
+            else
+              NEW_VERSION_CODE=$(date_version_code "now")
+              echo "new_version_code=${NEW_VERSION_CODE}" > $VERSION_FILE
+            fi
+          fi
+          # Ensure version code is not higher than max version code
+          MAX_VERSION_DATE="2025-12-31 23:59"
+          MAX_VERSION_CODE=$(date_version_code "$MAX_VERSION_DATE")
```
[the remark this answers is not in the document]
coreycb: Re version.properties: Ah, yeah of course..  maybe an upload artifact would be the right approach. I'll try that out.

Re version code: Well, hopefully that code isn't wrong! But yes, I can hardcode it instead.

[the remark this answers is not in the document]
kewisch: Re version.properties - If you calculate the date in an early job that is not parallel, you should be able to use that going forward, dump_environment would work for that.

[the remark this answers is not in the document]
kewisch: And if you could add a check for release/beta that the version code is not higher than 30, we should be good until end of 2025.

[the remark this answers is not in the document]
coreycb: I'm testing the new commits I've pushed. I'll let you know when they're ready for review.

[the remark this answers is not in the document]
coreycb: This is ready for review again. I'm using upload-artifact and download-artifact to store version.properties for parallel matrix jobs. It's not atomic but it should be good enough if there's a race. I had to use "continue-on-error: true" for both actions because they fail if the file exists (upload-artifact) or doesn't exist (download-artifact). I don't love it but it seems to work.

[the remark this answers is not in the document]
coreycb: Another option that would work better is to introduce another job before release_commit and let release_commit use the output from it. I just don't know if we want another job.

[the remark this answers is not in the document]
coreycb: Sorry, I just saw your latest comments. I'll rework with your comments.

.github/workflows/shippable_builds.yml
kewisch: That's fair, let's make it a month earlier.

.github/workflows/shippable_builds.yml
kewisch: Not sure if this is still coming, but if it is always padded, we don't need the printf.

coreycb:
@kewisch this should be ready to review when you get some time

.github/workflows/shippable_builds.yml
```diff
@@ -694,14 +746,25 @@ jobs:
...
-          wget -O whatsnew/whatsnew-en-US https://raw.githubusercontent.com/${REPO}/${APP_SHA}/${FILEPATH}
+          if [[ "$RELEASE_TYPE" = "beta" || "$RELEASE_TYPE" = "release" ]]; then
+            FILEPATH=app-metadata/${APPLICATION_ID}/en-US/changelogs/${VERSION_CODE}.txt
+            wget -O whatsnew/whatsnew-en-US https://raw.githubusercontent.com/${REPO}/${APP_SHA}/${FILEPATH}?token=$(date +%s)
+          elif [[ "$RELEASE_TYPE" = "daily" ]]; then
+            COMMIT_URL="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/commit/${GITHUB_SHA}"
+            WORKFLOW_URL="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+            echo "${APP_NAME^} ${VERSION_NAME} (${VERSION_CODE}) Daily Build" | tee whatsnew/whatsnew-en-US
+            echo "- Built from GitHub commit: ${COMMIT_URL}" | tee -a whatsnew/whatsnew-en-US
+            echo "- Built with GitHub workflow: ${WORKFLOW_URL}" | tee -a whatsnew/whatsnew-en-US
```
kewisch, line 766:
```suggestion
            sed 's/^[[:space:]]*//' > whatsnew/whatsnew-en-US <<EOF
              ${APP_NAME^} ${VERSION_NAME} (${VERSION_CODE}) Daily Build
              The Daily version is an unstable testing and development platform, make sure you back up important data regularly!
              If you find any issues, get in touch at https://github.com/thunderbird/thunderbird-android/issues/ and find us on Matrix at https://matrix.to/#/#tb-android-dev:mozilla.org
              Be prepared to debug the code in Android Studio :-)
              
              - Built from GitHub commit: ${COMMIT_URL}
              - Built with GitHub workflow: ${WORKFLOW_URL}
            EOF
```

.github/workflows/shippable_builds.yml
```diff
@@ -409,9 +446,20 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
         with:
-          cache-disabled: ${{ needs.dump_config.outputs.releaseType == 'beta' || needs.dump_config.outputs.releaseType == 'release' }}
+          cache-disabled: "${{ contains(fromJSON('[\"beta\", \"release\"]'), needs.dump_config.outputs.releaseType) }}"
           add-job-summary: on-failure
 
+      - name: Set Version Code for Daily
+        shell: bash
+        env:
```
kewisch, line 454: Shouldn't this be daily only?
```suggestion
      - name: Set Version Code for Daily
        if: ${{ needs.dump_config.outputs.releaseType == 'daily' }}
        shell: bash
        env:
```

kewisch (approved):
r+ with the following comments considered, I've added suggestions to make it easier to include.

PR #8441, opened 2024-10-30
Remove app

jfx2006:
Split the publish_release job into pkginfo, publish_release_github, and publish_release_playstore jobs in anticipation of adding a publish_ftpmo job which requires different permissions.

This also discontinues the use of `actions/create-github-app-token@v1` for publish_github and replaces it with `contents: 'write'` permission.

There are no associated changes to the environments to go with this change.

### DIFF
--- a/.github/workflows/shippable_builds.yml
+++ b/.github/workflows/shippable_builds.yml
@@ -551,7 +551,7 @@ jobs:
             🟡 [${{ needs.dump_config.outputs.releaseType }} build ${{ github.run_number}}](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})
             waiting for publish approval (triggered by ${{ needs.notify_build_start.outputs.actorLink }})
 
-  pre_hold:
+  publish_hold:
     # This is a holding job meant to require approval before proceeding with the publishing jobs below
     # The environment has a deployment protection rule requiring approval from a set of named reviewers
     # before proceeding.
@@ -569,7 +569,7 @@ jobs:
   pkginfo:
     # Generate pkginfo used by the publish jobs and the release notes
     name: Generate PkgInfo
-    needs: [pre_hold, dump_config, release_commit]
+    needs: [publish_hold, dump_config, release_commit]
     if: ${{ !failure() && !cancelled() }} # Run if previous step is skipped
     runs-on: ubuntu-latest
     strategy:
@@ -679,7 +679,7 @@ jobs:
 
   publish_release_github:
     name: Publish Release Github
-    needs: [pre_hold, dump_config, release_commit, pkginfo]
+    needs: [publish_hold, dump_config, release_commit, pkginfo]
     if: ${{ !failure() && !cancelled() }} # Run if previous step is skipped
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/shippable_builds.yml
```diff
@@ -566,20 +566,24 @@ jobs:
...
-      k9mail_release_url: ${{ steps.summary.outputs.k9mail_release_url }}
-      thunderbird_full_version_name: ${{ steps.summary.outputs.thunderbird_full_version_name }}
-      k9mail_full_version_name: ${{ steps.summary.outputs.k9mail_full_version_name }}
+      TAG_NAME: ${{ steps.pkginfo.outputs.TAG_NAME }}
+      FULL_VERSION_NAME: ${{ steps.pkginfo.outputs.FULL_VERSION_NAME }}
+      VERSION_NAME: ${{ steps.pkginfo.outputs.VERSION_NAME }}
+      VERSION_CODE: ${{ steps.pkginfo.outputs.VERSION_CODE }}
+      APPLICATION_ID: ${{ steps.pkginfo.outputs.APPLICATION_ID }}
+      app_sha: ${{ steps.shanotes.outputs.app_sha }}
+      app_github_notes: ${{ steps.shanotes.outputs.app_github_notes }}
```
kewisch, line 585: Let's use a consistent format with outputs

.github/workflows/shippable_builds.yml
```diff
@@ -674,29 +657,93 @@ jobs:
 
           cat $GITHUB_OUTPUT
 
+      - name: Rename release assets
```
kewisch, line 660: Renaming them was mainly to make sure we have sensible names in the steps such as release assets and artifact names. I don't think we should have a separate artifact to differentiate renamed vs not renamed.

.github/workflows/shippable_builds.yml
```diff
@@ -709,26 +756,45 @@ jobs:
...
+          releaseName: ${{ needs.pkginfo.outputs.VERSION_NAME }}
           status: completed
           changesNotSentForReview: ${{ inputs.draftGooglePlay }}
           whatsNewDirectory: whatsnew
           releaseFiles: |
-            uploads/${{ steps.rename.outputs.PKG_FILE }}
+            uploads/${{ needs.pkginfo.outputs.PKG_FILE }}
 
+  publish_summary:
+    name: Publish Release Play Store
```
kewisch, line 769: Name needs changing

.github/workflows/shippable_builds.yml
kewisch: Also, if you are using a Matrix job, the outputs will overwrite each other, so you only get one random set of variables.

If we need all of these later, it might make sense that instead of picking them from the apk/aab, we generate them from the sources. We have most of this already in the appinfo step early on, we just need to put it into the right outputs.

.github/workflows/shippable_builds.yml
```diff
@@ -674,29 +657,93 @@ jobs:
...
           prerelease: ${{ env.RELEASE_TYPE != 'release' }}
           fail_on_unmatched_files: true
           files: |
-            uploads/${{ steps.rename.outputs.PKG_FILE }}
+            uploads/${{ needs.pkginfo.outputs.PKG_FILE }}
+
+  publish_release_playstore:
+    name: Publish Release Play Store
+    needs: [publish_hold, dump_config, release_commit, pkginfo]
+    if: ${{ !failure() && !cancelled() }} # Run if previous step is skipped
```
kewisch, line 722: We can skip this whole job if the skipGooglePlay input is set.

kewisch (changes requested):
Can you also update setup_release_automation to update the environment to how it needs to be set up? This will make testing and future setup easier.

A few early comments, I'll need to take a closer look at some point.

jfx2006:
> Are any other changes needed to setup_release_automation, or is there no change since you didn't change the environment?
There should not be any changes needed.

wmontwe:
@jfx2006 Is this still needed?

PR #8466, opened 2024-11-01
Bump version name and suffix to 9.0b0

coreycb:


kewisch:
For the record we are using b0 as this will bump to b1 on the first beta build for that version

PR #8467, opened 2024-11-01
Bump version name and suffix to 10.0a1

coreycb:


coreycb:
There are currently no daily builds for k9mail but I figured it may be worth while to keep the version name/suffix the same.

PR #8470, opened 2024-11-01
Add a print mode to setup_release_automation

kewisch:


scripts/setup_release_automation
```diff
@@ -467,44 +544,64 @@ def main():
...
+
+            for name, value in data["variables"].items():
+                if isinstance(value, dict) or isinstance(value, list):
+                    value = json.dumps(value)
 
-        for name, value in data["variables"].items():
-            if isinstance(value, dict) or isinstance(value, list):
-                value = json.dumps(value)
-            set_github_environment_variable(args.repo, name, value, environment_name)
+                else:
```
coreycb, line 571: Was the else statement intended here?

kewisch: Good catch, it was not!

Issue #8238, opened 2024-10-04
Show picture profile on account list


Vucumpra:
### Checklist

- [X] I have used the search function to see if someone else has already submitted the same feature request.
- [X] I will describe the problem with as much detail as possible.
- [X] This issue only contains a request for one single feature, **not** multiple (related) features.

### App version

8.0b1

### Problem you are trying to solve

I have three Gmail accounts, all displayed as [GM] :/

### Suggested solution

Show picture from address book

### Screenshots / Drawings / Technical details

![Screenshot_20241004-104245](https://github.com/user-attachments/assets/f614ea26-b91d-4156-98c7-21d50a02ec4b)

MillionsToOne:
Yeah. If no picture is set, maybe use the first letters of the names instead of the first 2 letters of the domain.

MrEAlderson:
![Screenshot_20241005_204648_Thunderbird Beta](https://github.com/user-attachments/assets/ec59a27c-8097-47ed-b42d-70271c78bec1)
It is no fun guessing which account is which - K9 at least had a drop down picker. The account picker is at its current stage unusable

Vucumpra:
Yes, in order, select picture from : 
1- Personal address book
2- OR Remote account
3- OR firsts 2 letters of ID

kewisch:
We originally planned to add a picture but went with this for time reasons. The picture could come from the contacts, or potentially setting one. 

The letters from the domain were a good compromise since in many cases folks will have eg a personal email on gmail, a work email, and maybe another email from a different provider. We decided against using parts of the name, because it is more likely to be the same if used by the same person. Having multiple separate email accounts at the same domain might be more of an edge case.

That is not to say it should stay this way. If someone wants to work on this it would be a great project for someone with past Android development experience.

kewisch:
I am happy to share the original mockups if needed

marcRDZ:
> I am happy to share the original mockups if needed

Hi @kewisch can you show or send me them? I'd like to work on this awesome feature. :)

Julien-Delavisse:
Hello, I wonder why users prefer to see an image associated with an account, rather than ... the email address.
Can you tell us a little more about the choice to switch from the old design (e-mail address) to the new one (two letters ...)?
It may be “prettier”, but the user experience is much less pleasant.
Do you have several accounts and do you find it pleasant to guess which account is behind “AB” or “CD”? Not to mention the duplicates...
[...]

PR #8492, opened 2024-11-04
Extract chip initials from display name

### LINKED ISSUE #8381: Changing name of mail account does not change the two letters in account chip.
### Checklist

### App

Thunderbird for Android and K-9

### App version

8.0b4

Other

### Android version

Android 14

### Device model

Samsung Tab S9

### Steps to reproduce

1. Go to settings, select an account
2. In general settings of this account, change the account name, especially modify the first 2 letters

### Expected behavior

The 2 letters in account panel will follow that change. 

### Actual behavior

Nothing happens in account panel

### Logs

_No response_

marcRDZ:
Fixes #8381 

The chip abbreviation is extracted from display name different parts splitting by special characters if present or taking two first letters simply, giving more possible combinations by default and also allowing the user customizes it through account name on general settings.

I think we should take into consideration some of the suggestions shared on this and related issues to future enhancements like add a picture option #8238 or full name below chip https://github.com/thunderbird/thunderbird-android/issues/8381#issuecomment-2453493668

### DIFF
--- a/feature/navigation/drawer/src/main/kotlin/app/k9mail/feature/navigation/drawer/ui/account/AccountAvatar.kt
+++ b/feature/navigation/drawer/src/main/kotlin/app/k9mail/feature/navigation/drawer/ui/account/AccountAvatar.kt
@@ -50,7 +50,7 @@ internal fun AccountAvatar(
                     .border(2.dp, MainTheme.colors.surfaceContainerLowest, CircleShape),
             ) {
                 Placeholder(
-                    email = account.account.email,
+                    displayName = account.account.displayName,
                 )
                 // TODO: Add image loading
             }
@@ -64,11 +64,11 @@ internal fun AccountAvatar(
 
 @Composable
 private fun Placeholder(
-    email: String,
+    displayName: String,
     modifier: Modifier = Modifier,
 ) {
     TextTitleMedium(
-        text = extractDomainInitials(email).uppercase(),
+        text = extractNameInitials(displayName).uppercase(),
         modifier = modifier,
     )
 }
@@ -102,6 +102,6 @@ private fun UnreadBadge(
     }
 }
 
-private fun extractDomainInitials(email: String): String {
-    return email.split("@")[1].take(2)
+private fun extractNameInitials(displayName: String): String {
+    return displayName.take(2)
 }


kewisch:
Thanks for taking a stab at this so quickly! 

With the suggested approach I think we'll be fine for people who have e.g. multiple gmail addresses, but if I have firstname.lastname@example.com for home and firstname.lastname@thunderbird.net for work, then both would be FN if I understand correctly?

Have you seen https://github.com/thunderbird/thunderbird-android/issues/8381#issuecomment-2428881786 by chance? I've since discussed this with the team and there are some concerns this could be too magical, but I'd like to work out a short term solution. Let me bring this up again tomorrow and we'll come back with a suggestion.

marcRDZ:
Yeah, in that case both accounts will show FL but at least the user had the chance to modify it on account settings. 

Anyway, as I said it's only a practical approach. Add new features as custom pictures or other things would be really nice.

Thinking about that comment you've mentioned, if you set as default account name something different to the email address as it suggested, maybe the user realises that it can be customised clearly. In my first experience setting up accounts I didn't see that option until I'd been dealing some time with the app.

I will be looking for your suggestions, thanks!

wmontwe:
I think using the display name is a good first step, it defaults to the users email address until it is explicitly changed. But we're discussing possible solutions and will come back to your suggestion.

wmontwe:
Related to #8100

marcRDZ:
> I think using the display name is a good first step, it defaults to the users email address until it is explicitly changed. But we're discussing possible solutions and will come back to your suggestion.

I've simplified the change to this idea and I'll wait for your deliberations.

wmontwe (approved):
Thank you, we'll go with your changes and will add an additional migration step to cover existing users.

marcRDZ:
Great! Don't hesitate to count on me for other changes related. 
There are many nice features we could add taking into account suggestions shared on this issue.

Ted-T-east:
Below is an example of the current Android Thunderbird presentation.

    Also, shown is the current Android Blue mail presentation.

    Candidly, I don’t care if the underlying email account is Yahoo, Gmail, Proton, private server, etc.

    But, I do wish to click on an email account directly.

    Thus, it would be useful to display a user supplied two/three character identity.

    How?

    Add one more field to the basic account data, wherein the user can input two or three characters to more easily identify one account vs. another.

    For various reasons, essentially segmentation of various “classes” of emails. I have multiple accounts. A two or three character identifier would really assist me or anyone to move directly to a target account. Showing my email accounts with a two of three character acronym identifier is far better than multiple accounts displaying YA or GM with only a color to remind server, but NOT which email account.

    My email accounts and my identifying characters if the Thunderbird team would accept and implement the foregoing.

    FAM     Family
    FRN     Friends
    VET     Veteran
    PUR         Purchases
    MED       Medical
    NWS       Newsletters
    MST       Master
    MUS       Music
    WFE       Wife
    EVE        EVENTS
02 account mod for input
03 BlueMail
011 current and proposed

PR #8509, opened 2024-11-05
Add answered label when needinfo label was removed

wmontwe:
This should help to detect answered issues.


kewisch:
I was thinking about this today as well, I think this is a good solution. We'll need some guidelines on how to deal with "answered" labels. Thinking out loud here:

1) When the answer given allows us to confirm the bug, take immediate action, put it into planning, we remove the answered label and apply any other labels/milestones we might need
2) When the answer is inconclusive, we either ask for more info (remove answered, add needinfo), or we respond saying we don't have a good answer or whatnot, and remove the label. I'm not sure we can take any other actions?


Either way whoever is on issue duty should triage all "answered" issues and remove the label. We shouldn't end up with a bunch of issues that remain in the answered state, otherwise we'd end up triaging those again and again.

Did I miss anything?

kewisch (approved):

PR #8551, opened 2024-11-11
Fix drawer jumping account icons

wmontwe:
This change displays all accounts in the drawer account list to prevent any jarring transitions when switching between them. For easier readability, the selected account is now displayed smaller, to help differentiating it from unselected accounts.


Domilopment:
Hello! I hope it’s okay to leave feedback here. In addition to graying out the selected account, perhaps the account circle could also shrink slightly when selected, or even become a small empty placeholder circle to mark its position. Best regards!

kewisch:
I don't think we should grey it out, changing colors could confuse people more especially with the option for other accounts having that color as the standard color as you mention. Different size would work for me. I don't have a constructive better solution at the moment though, maybe @laurelterlesky has thoughts?

PR #8556, opened 2024-11-12
Update README

leoheitmannruiz:


wmontwe:
Thank you. Could you please provide the original image without further compression so that I can compare it to the current one in binary format?

wmontwe (approved):
We verified ourselfs, thank you 👍

PR #8557, opened 2024-11-12
plain-text -> plain text

leoheitmannruiz:
Should the following be update here or via Weblate?

https://github.com/thunderbird/thunderbird-android/blob/main/legacy/ui/legacy/src/main/res/values-en-rGB/strings.xml#L617


wmontwe (approved):
Thank you, en-GB is managed via Weblate and the translation could be changed once this pull-request is merged.

PR #8566, opened 2024-11-14
Enable funding for K-9 Mail Full

wmontwe:
This enables the funding feature for K-9 Mail.

Strings will be added in another pull-request, once they are ready.


cketti (approved):
It took me a while to understand why this involves a change to `TbFeatureFlagFactory`.

Removing the feature flag and switching the funding module for K-9 Mail are only distantly related changes. I think separating them into individual commits with their own commit messages would give readers a better understanding of what the changes are about.

PR #8580, opened 2024-11-18
Fix version filter for beta releases

kewisch:
README currently shows the release for both beta and actual release, this should fix it


wmontwe (approved):
👍

PR #8586, opened 2024-11-18
Render notes from the right directory for point release betas

kewisch:
@coreycb the `0b` logic won't work for 8.2b1, this should take care


kewisch:
Cancelling this as we need to name it 9.0b1

PR #8596, opened 2024-11-20
Fix MessageList colors for better contrast

wmontwe:
This removes the custom color attributes and replaces them by Material 3 colors. 

In result it's already better for the light theme, while the dark theme could be still tweaked to enhance the visual distinction between read and unread emails.

Resolves: [#9039](https://github.com/thunderbird/thunderbird-android/issues/9039)

Portrait:

Landscape:


wmontwe:
I removed this attributes as they should have been removed already before. They are just blocking the view on the actual issue. And yes, this has been discussed and agreed before with design. Just other things have been more important. But anyway.

wmontwe:
I adjusted the colors but I'm waiting for design feedback on some adjustments.

wmontwe:
I updated the colors to match the latest design, but this needs still a rework of the message list item and should be done as part of a specific epic.

kewisch (approved):

PR #8599, opened 2024-11-20
Add Thunderbird website hyperlink on contribution screen

### LINKED ISSUE #8430: When contributions are unavailable, link the website
When contributions are unavailable, it shows the text we devised. It does not however hyperlink thunderbird.net nor click to link to the website.

Also, there is a period missing at the end.

image

marcRDZ:
Fixes #8430 

I've followed this commit [c777ad5](https://github.com/thunderbird/thunderbird-android/commit/c777ad55ed82e69f1c0a6ee423621f891a60e52e) as a guide.
 - The string is fixed with punctuation mark at the end for all translations.
 - The whole text is clickable opening Thunderbird website.

I still doubt if I should hoist the click event and handle with a side effect as I've seen for other cases or it's ok on this way.
image


feature/funding/googleplay/src/main/kotlin/app/k9mail/feature/funding/googleplay/ui/contribution/ContributionList.kt
```diff
@@ -197,14 +209,33 @@ private fun ListEmptyView(
...
+        val annotatedString = annotatedStringResource(
+            id = R.string.funding_googleplay_contribution_list_empty_message,
+            argument = buildAnnotatedString {
+                withStyle(
+                    style = SpanStyle(
+                        color = MainTheme.colors.primary,
+                        textDecoration = TextDecoration.Underline,
+                    ),
+                ) {
+                    append(stringResource(R.string.funding_googleplay_thunderbird_website_domain))
```
wmontwe, line 222: The Compose APIs are changing constantly and I've seen there is a nicer way to achieve this now. Instead of adding a click listener on the whole text. You could use `withLink` to make just the link clickable.

```suggestion
                    withLink(
                        LinkAnnotation.Url(
                            url = stringResource(R.string.funding_googleplay_thunderbird_website_url),
                        ),
                    ) {
                        append(stringResource(R.string.funding_googleplay_thunderbird_website_domain))
                    }
```

wmontwe (changes requested):
Thanks, the direction is great. I just found a better way to support in-text links. 

Typicaly, we just change the source translations, but as this introduces a placeholder and Weblate doesn't handle that well. I think editing them here is the safer approach. Before merging this patch, we should pull the latest translations from Weblate and update this patch to ensure all changes are included.

Depends on #8608 <- merged

feature/funding/googleplay/src/main/res/values/constants.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
wmontwe, line 1: We enabled the funding feature for K-9 Mail. These links need to be updated in the K-9 Mail `constants.xml`.

feature/funding/googleplay/src/main/kotlin/app/k9mail/feature/funding/googleplay/ui/contribution/ContributionList.kt
marcRDZ: Great! I just was looking for something similar as we did with links and annotations before Compose but I implemented the other way to be aligned with other parts of the project.

feature/funding/googleplay/src/main/res/values/constants.xml
marcRDZ: Done

feature/funding/googleplay/src/main/kotlin/app/k9mail/feature/funding/googleplay/ui/contribution/ContributionList.kt
```diff
@@ -260,3 +294,20 @@ private fun ListErrorView(
         )
     }
 }
+
+private fun Context.launchThunderbirdWebsite() {
```
wmontwe, line 298: This is not necessary anymore, the url is handled by the system already.

marcRDZ: Fixed, Idk why but on my previous attempt it seems not working so I thought I had to handle through the listener. Now I tested again and it's opening the url properly.

wmontwe (approved):
Thank you, just updated two missing translations.

marcRDZ:
Don't mention it! I think I missed it merging translations on main rebase. Thanks for the quick fix.

PR #8609, opened 2024-11-26
Fix Thunderbird release 8.1's versioncode

kazushi3:
This changes TB 8.1's versioncode. 8.1's changelog is shown between 8.0b2 and 8.0b3 because it has incorrect versioncode.

### DIFF
--- a/app-thunderbird/src/main/res/raw/changelog_master.xml
+++ b/app-thunderbird/src/main/res/raw/changelog_master.xml
@@ -5,7 +5,7 @@
      Locale-specific versions are kept in res/raw-<locale qualifier>/changelog.xml.
 -->
 <changelog>
-    <release version="8.1" versioncode="5" date="2024-11-06">
+    <release version="8.1" versioncode="9" date="2024-11-06">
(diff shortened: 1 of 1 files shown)


kewisch:
It isn't that simple unfortunately. This was due to our release automation and merging not being 100% worked out. 8.1's version code is indeed 5, connected with the app net.thunderbird.android. The beta's versions are connected with the app net.thunderbird.android.beta.

I don't know how useful it would be to change the XML afterwards since the versions are already released. If we want to be verbose (and accurate) we could:

* Merge the notes from 8.0b5 to those of 8.1 with version code 5
* Merge the notes from 8.0b1 to 8.0b4 to a version 8.0 with version code 4

The other thing we could do is fall back to the version notes we prepared separately, which would be more in line with how automation works:

* Take all items with `short_notes` from https://github.com/thunderbird/thunderbird-notes/blob/master/android_release/8.0.yml and add them to the changelog xml under version 8.0 with version code 4
* Delete 8.0b1 - 8.0b5
* Keep 8.1 as it is

We were talking about this when we released 9.0b1, and decided to live with it for now, and do better in the future. If you still want to make a change, then the first approach would probably be closest to what is already visible in the notes.

kazushi3:
Oh...

I understood this is not easy and team has already discussed. I don't think I have better solution.
Thank you for your reply!

Issue #760, opened 2015-08-26
Enhancement: move message to another account


klonamy:
I would like to have the ability to move a message from one account to another like thunderbird.

Xeroin:
+1 on this.

kejriwals:
PLEASE, PLEASE add this feature ASAP. Only Samsung native email app has this feature for Android. Looking forward to hearing from you.  Thanks.

milankni:
Any progress on this feature?

It seems that the only alternative is MailDroid app, which is not open source, however. 

Any other suggestions? (Sorry for highjacking.)

kejriwals:
I have been using Maildroid for almost a year now. It works like a charm, including the feature of moving mails across accounts/folders. 

Best, 

Sandeep Kejriwal
www.SandeepKejriwal.com

Sent from my phone

-----Original Message-----

r-barnes:
Also hoping for this feature.

BennyLi:
I've been using k-9 for a while. The missing ability to move messages over to my personal archiving IMAP account is really disappointing me. 

Is there any reason the PR is not going to be merged?

corbolais:
Hi there,

First off, thank you for k9-mail. Great e-mail tool.

Over the years, I faced the need to move e-mail messages between accounts on my android phones. k9-mail does still not provide any means to achieve that. So one has to go at length and try various ways to get the messages transferred.

What is the status of this PR and why is this feature still not implemented in k9-mail?

cheers

vdbhb59:
Any update on this one pls? This is 7 years old, and is a really handy feature.

schniepp:
See more discussion [here](https://forum.k9mail.app/t/feature-request-move-emails-between-imap-accounts-servers/3753/5).

corbolais:
It's a PITA k-9 still does not support this useful feature. However, k-9 is a very nice and useable app. I'd like it to have this feature, it's so useful at times (actually regularly).

AleXSR700:
Is this feature being worked on?

vdbhb59:
Don't ask. Almost every comment has been marked as off-topic, even when it clearly is on this topic itself.

efa:
6.803 still miss this Thunderbird feature

ph1248:
Hi, I also would be glad to have this feature! Could you consider working on it? 

There may also be a connection to ticket #5709  which could be possibly solved alongside. 

Thanks in advancea and kind regards!

[...]

PR #8632, opened 2024-12-02
Add move/copy-to-account

mbalduccini:
Fixes issue #760.
The approach is fairly straightforward:
- I extended the relevant methods to take the destination account
- If the destination account is different from the source account, the message is copied over using `MessageStore.saveLocalMessage()`
- In case of a move, the source message is deleted
- Move/copy of threads is supported

It doesn't break any tests as far as I can tell. There are 11 tests that fail on my computer, and they are the same tests that fail in the main branch. Screenshots attached.

Screenshot_20241201_204310_Chrome
Screenshot_20241201_204256_Chrome
Screenshot_20241201_204219_Chrome


cketti:
The failing protocol tests might be because the tests were run using an (old?) JDK without an "unlimited cryptography policy". See https://www.java.com/en/configure_crypto.html

The failing tests in `QuoteDateFormatterTest` are also related to the Java version used to run the tests. #8651 should work around this, though.

In general it's probably best to use the same Java version our GitHub workflow is using. [Currrently](https://github.com/thunderbird/thunderbird-android/blob/e3526d50bae9ef06e25bc4de7cf51a70522b7f0e/.github/workflows/android.yml#L35) that's Java 17.

mbalduccini:
I switched to Java 17 (I was using 21) and things got a little better. The *ServerSettingsValidatorTest tests no longer fail. The RealImapConnectionTest test still fails as well as the tests in QuoteDateFormatterTest. I am attaching the screenshots.
@cketti I see that there are conflicts on three files. Would you like me to resolve them?

Screenshot_20241205_213356_Chrome
Screenshot_20241205_213326_Chrome
Screenshot_20241205_213303_Chrome

mbalduccini:
@cketti I merged your latest changes including #8651. All tests pass now.

cketti:
Thanks for working on adding support for moving messages from one account to another :heart: 

Unfortunately, the suggested change only works in the best case and doesn't support error handling and recovery at all. Mostly that's because the necessary infrastructure for that is not in place. I listed (some of) the things that need to be considered here: https://github.com/thunderbird/thunderbird-android/issues/760#issuecomment-2578013481

There's a lot of changes we need to make to `MessagingController` and how we handle pending commands before this feature can be implemented in a reliable way. I'm working on writing all of this down, so we can devise a plan to incrementally work towards getting there.

I don't think it would be reasonable to put all of the necessary changes in one pull request. So my recommendation is to close this pull request.

PR #8667, opened 2024-12-12
[CASA] Fix GitHub actions

wmontwe:
This follows the CASA requirement to pin GitHub actions to their Git hash. This should work with dependabot. 

I used [frizbee](https://github.com/stacklok/frizbee) to retrieve the hashes and validated against their latest versions.

Also changed `setup-gradle` to use the built-in validation and better cache control.

Resolves: [Supply Chain Attack - GitHub Actions](https://help.fluidattacks.com/portal/en/kb/articles/criteria-vulnerabilities-437)

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -25,8 +25,8 @@ jobs:
 
       - uses: actions/setup-java@v4
         with:
-          distribution: temurin
-          java-version: 17
+          distribution: 'temurin'
+          java-version: '17'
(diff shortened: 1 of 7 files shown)


coreycb (approved):
LGTM. I chatted with Wolf about whether we should document frizbee use somewhere or if this will be automated and he mentioned that it should be automated in the future, and he might add another GitHub action to ensure git hashes are used.

PR #8672, opened 2024-12-12
[CASA] Add validation workflows

wmontwe:
This adds workflows for validating the project:

- Gradle Wrapper validation
- [Scorecard](https://github.com/ossf/scorecard) to check for using best practises
- Script to check that GitHub actions are pinned to a Git hash

Workflow permissons are now by default `read-only` and further permissions need to be granted on job level.

### DIFF
--- a/.github/workflows/validate-workflows.yml
+++ b/.github/workflows/validate-workflows.yml
@@ -1,0 +1,31 @@
+name: "Validate Workflows"
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      dryRun:
+        type: boolean
+        description: Dry run
+        default: true
+      debug:
+        type: boolean
+        description: Debug mode
+        default: false
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  validate-workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Ensure GitHub actions are SHA pinned
+        env:
+          DRY_RUN: ${{ !inputs.dryRun && '--no-dry-run' || '' }}
+          DEBUG: ${{ inputs.debug && '--debug' || '' }}
+        run: |
+          bash scripts/validate-github-actions-pinned.sh $DRY_RUN $DEBUG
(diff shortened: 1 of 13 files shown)


.github/workflows/validate-workflows.yml
```diff
@@ -0,0 +1,17 @@
...
+permissions: read-all
+
+jobs:
+  validate-workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Ensure GitHub actions are SHA pinned
+        uses: zgosalvez/github-actions-ensure-sha-pinned-actions@5d6ac37a4cef8b8df67f482a8e384987766f0213 # v3.0.17
```
coreycb, line 17: I'm hesitant on using this action but it doesn't seem like there are (m)any alternatives. Have you audited the code?

wmontwe: I think we might need some rules for external actions, in this case the source is very basic. I'll rewrite it as bash script.

scripts/validate-github-actions-pinned.sh
```diff
@@ -0,0 +1,126 @@
...
+sha1_regex='^[a-f0-9]{40}$'
+sha256_regex='^[A-Fa-f0-9]{64}$'
+
+# Default values
+workflows_path=".github/workflows"
+dry_run=true
+debug=false
+action_has_error=false
+
+# Parse command-line arguments
```
coreycb, line 14: I think shift might not be doing anything in this case statement. You may need to use something like this for switch to work:

```
while [ "$#" -gt 0 ]; do
  case $1 in
```

wmontwe: Right, I didn't see that. Fixed

scripts/validate-github-actions-pinned.sh
```diff
@@ -0,0 +1,126 @@
...
+  fi
+
+  jobs=$(sed -n '/jobs:/,/^[^ ]/p' "$file")
+
+  while read -r line; do
+    if [[ "$line" =~ ^[[:space:]]*uses: ]]; then
+      uses=$(echo "$line" | awk -F: '{print $2}' | xargs)
+      run_assertions "$uses" || file_has_error=true
+    elif [[ "$line" =~ ^[[:space:]]*steps: ]]; then
+      steps_section=true
```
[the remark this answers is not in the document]
wmontwe: Yes, the `steps_section` is not necessary. I simplified the loop.

coreycb (approved):

PR #8694, opened 2024-12-19
Add accessible name for contact picture checkbox

williamspiegel:
_Fixes #8446_ for the most part, but announcing the checked state seems to mostly be already handled by the "X selected" announcement. Maybe need some clarification about what needs to be changed there if that is still considered an issue. This PR only handles the issue with the checkbox itself being unlabeled.

### DIFF
--- a/legacy/ui/legacy/src/main/res/layout/message_list_item.xml
+++ b/legacy/ui/legacy/src/main/res/layout/message_list_item.xml
@@ -226,6 +226,7 @@
         android:id="@+id/contact_picture_click_area"
         android:layout_width="@dimen/messageListStartKeyline"
         android:layout_height="0dp"
+        android:contentDescription="@string/checkbox_content_description"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent"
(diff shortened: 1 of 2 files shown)


dankeck:
Thanks for working on this and for adding the checkbox labeling. Hopefully I can clarify the other part of my issue.

It's true that, with some effort, a user could determine the current state of the checkbox by toggling it and listening to the "X selected" announcement, to see if X goes up or down. If X goes up, then the checkbox went from unchecked to checked. If X goes down, then the checkbox went from checked to unchecked.

But this constitutes extra steps for users with disabilities. It'd be great if landing on a checkbox could announce its current state right away.

In HTML, native checkboxes [have this behavior by default](https://a11ysupport.io/tests/tech__html__input__input-checkbox#assertion-html-input(type-checkbox)_element-convey_checked_state-). For [custom-built checkboxes in HTML](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox), an attribute named [aria-checked](https://w3c.github.io/aria/#aria-checked) is added to tell screen readers to announce the current state. But I don't know what an analog would be in Android code.

wmontwe (changes requested):
I think adding the content description is not enough in this case, as the selection status isn't announced by Talkback. It would be good to provide a way for screen readers to announce whether the checkbox is selected or not.

kewisch:
@williamspiegel have you had a chance to dive into this again? Would love to get this merged!

wmontwe:
We've decided not to move forward with this PR at this time. We appreciate the effort that went into creating it and would love to see improvements before revisiting it.

PR #8713, opened 2025-01-03
Show the "Use system default" theme option on pre API 28 devices

Ephemera42:
Android [Night Mode](https://developer.android.com/reference/kotlin/android/app/UiModeManager#mode_night_auto) was introduced in API 8, which enables automatic switching between day and night modes. Therefore, the follow system theme option is available on pre API 28 devices.

[output.webm](https://github.com/user-attachments/assets/d6f079e4-9335-43aa-9fc1-7e9b7f778f33)

### DIFF
--- a/legacy/core/src/main/res/values/arrays_general_settings_values.xml
+++ b/legacy/core/src/main/res/values/arrays_general_settings_values.xml
@@ -143,11 +143,6 @@
         <item>ja</item>
     </string-array>
 
-    <string-array name="theme_values_legacy" translatable="false">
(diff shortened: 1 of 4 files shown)


wmontwe (changes requested):
@Ephemera42 Thank you for simplifying the theme handling logic. I tested the behavior on API levels 21-27 and did not find any side effects.

Do you maybe know how to enforce night mode before API level 28?

Please also remove the remaining `theme_values_legacy` array.

wmontwe (approved):
Thank you.

PR #8719, opened 2025-01-06
Dark theme (& Material You theming) for widget

vayun-mathur:
Completes #3728

Rebuilt the message list widget using Jetpack Glance and incorporated dark mode and material you theming for the color scheme

This is what the widget looks like now when dark mode is enabled:

Screenshot_20250106-153429


vayun-mathur:
most recent commit should fix the style build issues @wmontwe

wmontwe (changes requested):
I apologize for the delay in getting back to you. This change is indeed very positive and represents a good starting point, though it still requires additional work. Instead of replacing the existing widget, I suggest that the changes be implemented in a separate module. Maybe guarded with a feature flag or integration in debug and daily only while beta and release disable that widget.

This approach will allow us to enhance the new version while keeping the original widget available.

Ideas/Improvements:

- The Glance Theme should align with our app theme and be compatible with K-9 and TfA. A new common module for widgets could serve this purpose.
- This common module could also house UI elements that can be reused in future widgets.
- The widget currently displays the unified folder but does not yet use account colors to differentiate messages.
- Naming should adhere to the existing scheme.
- It's important to test how the widget performs across both applications.

Please let me know which aspects you would like to address personally. Some of these improvements can be broken down into follow-up tasks to complete the widget development.

vayun-mathur:
Ok I think that makes sense. I'll start by moving this moving this into its own module and restoring the old widget

kewisch:
@vayun-mathur have you had a chance to get back to this? We'd love to see this merged and much appreciate the effort you are putting into it and would love to enable the new widget on beta once it's ready.

vayun-mathur:
old widget has been added back and new widget has been moved into own module and naming has been made more consistent

wmontwe (approved):
@vayun-mathur Hi, sorry for taking so long, I went over the widget and fixed a couple of issues, joined all the changes into one commit and enabled the widget for K-9 debug and TfA debug and daily.

I think it needs further work, but is now in a state that is a good foundation so we could merge this and do remaining work in follow-up pull-requests.

Open:
- UI needs a design review
- There are some glitches updating the widget, sometimes it just shows the loading screen.

PR #8730, opened 2025-01-09
Fix GH Action token permissions

wmontwe:
This resolves the token permission reports by setting the minimal permission `contents: read` for all workflows. This is sufficient as the repository is setup with restrictive GH Action permissions.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -12,6 +12,9 @@ on:
       - 'LICENSE'
       - 'NOTICE'
 
+permissions:
+  contents: read
+
 jobs:
   build:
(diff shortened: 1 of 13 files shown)


coreycb (commented):
It looks like we already have contents to read-only by default. Or is this a safe guard in case settings get changed?

https://github.com/thunderbird/thunderbird-android/settings/actions
https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

wmontwe:
This is just a safeguard.

coreycb (approved):

PR #8737, opened 2025-01-13
Switch Import Settings to Activity Result API

marcRDZ:
Change related to #7758

Migrate pick document result to new API on settings import


cketti (approved):
Thanks :+1:

PR #8756, opened 2025-01-17
Fix TalkBack Accessibility for Contribution Amounts

terofeev:
Fixes https://github.com/thunderbird/thunderbird-android/issues/8346

This update applies `Modifier.selectable` instead of `Modifier.clickable`.

As a result, TalkBack now pronounces these examples correctly:
- Not selected **Give once** Radio button, double tap to toggle
- Selected **Monthly** Radio button
- Selected **25 dollars** Radio button

To resolve this issue, the `enabled = !isSelected` argument was removed. This allows users to select a type of contribution even if it is already selected. When this happens, the selected amount is reset to the second-lowest option.


wmontwe (approved):
Thank you

PR #8758, opened 2025-01-18
Account name should be single line if email and display name are identical

### LINKED ISSUE #8739: Account name should be single line if email and display name are identical
### Checklist

### App

Thunderbird for Android

### App version

10.0

Other

### Android version

12

### Device model

_No response_

### Steps to reproduce

Open drawer

### Expected behavior

Account name should be single line if email and display name are identical

### Actual behavior

Drawer displays lines with repeated text for both email and display name

### Logs

_No response_

shamim-emon:
- Show only  account name among account name and account email in `AccountView `if account name and account email is identical
- Fixes #8739


wmontwe (approved):
Thank you!

PR #8760, opened 2025-01-20
Add gradle-remote to execute gradle builds on a build server

wmontwe:
This adds a `gradle-remote` script that executes Gradle commands on a remote build server. It uses `rsync` to transfer the project files to the server and then connects via SSH to run the Gradle command remotely. Once the execution is complete, the contents of the build folder are synced back to the local machine.

Please note that this script is quite basic and not recommended for use with public build servers. Since `rsync` transfers all project files from the developer's machine, including potentially sensitive information like secrets or configuration files, it should be used with care.

### DIFF
--- a/gradle-remote
+++ b/gradle-remote
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+# This script serves as a drop-in replacement for running Gradle tasks locally.
+# It automates the process of building a Gradle project on a remote server by synchronizing local files,
+# executing specified Gradle tasks remotely, and copying all build outputs back to your local machine.
+
+# Disclaimer
+echo "Please note that this script is quite basic and not recommended for use with public build servers."
+echo "Since rsync transfers all files from the developer's machine, including potentially sensitive information"
+echo "like secrets or configuration files, it should be used with care."
+
+if ! [ -f build.gradle.kts ]; then
+  echo "No build.gradle.kts found"
+  exit 1
+fi
+
+remote=$GRADLE_REMOTE_SSH
+if [ -z "$remote" ]; then
+  echo "\$GRADLE_REMOTE_SSH is not set, please set it to a valid SSH login on your build server"
+  exit 1
+fi
+
+remote_dir=$GRADLE_REMOTE_DIR
+if [ -z "$remote_dir" ]; then
+  echo "\$GRADLE_REMOTE_DIR is not set, please set it to the directory where the project should be copied on the remote server"
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


gradle-remote
```diff
@@ -0,0 +1,35 @@
...
+  echo "$GRADLE_REMOTE_DIR is not set, please set it to the directory where the project should be copied on the build server"
+  exit 1
+fi
+
+remote_project_dir="$GRADLE_REMOTE_DIR/$(basename "$(dirname "$PWD")")_$(basename "$PWD")"
+
+set -e
+
+# Copy the project to the remote
+rsync -irci --exclude build --exclude .gradle --include .gradle/gradle.properties --delete . "$remote:$remote_project_dir"
```
kewisch, line 29: I think we should go ahead and exclude `.signing` to avoid accidental footguns. We can't protect all cases, but this one would be pretty important.

You could also consider using `git ls-files` with options to include only tracked files, the developer would have to git add the files if they want to transfer them.

gradle-remote
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+# Please note that this script is quite basic and not recommended for use with public build servers.
```
kewisch, line 3: If we're adding this file top-level instead of hidden in the build scripts folder, I think we should `echo` this message and prompt before continuing, optionally with a `-y` argument or env var to skip the prompt.

gradle-remote
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
kewisch, line 1: It would be great to add an intro to this file that explains what this file does and why. It can be inferred from the very short script, but knowing what for would be useful. (which I should have done for some of the new scripts in the scripts subfolder as well)

kewisch (approved):
r+ with above comments considered

gradle-remote
wmontwe: Adding `.signing` makes total sense. I updated the script to include an array for exclusions and inclusions.

`git ls-files` would make sense based on the use case for the script. At the moment I use it to offload builds to another machine under my control. So risk is minimal, but when this script is used to offload builds to a remoter server it needs further mitigations.

gradle-remote
wmontwe: I think adding a prompt defeats the purpose of a drop-in replacement for Gradle calls.

kewisch: I'm thinking more of the first time use case. If you use an environment variable to subvert the prompt, then you can simply set `GRADLE_REMOTE_ACCEPT_PITFALLS=y` or something like that and you can use it as a drop-in.

PR #8769, opened 2025-01-23
Address of only one recipient not shown but "+1"

shamim-emon:
- See #7449

### DIFF
--- a/library/TokenAutoComplete/src/main/java/com/tokenautocomplete/TokenCompleteTextView.java
+++ b/library/TokenAutoComplete/src/main/java/com/tokenautocomplete/TokenCompleteTextView.java
@@ -612,7 +612,7 @@ public abstract class TokenCompleteTextView<T> extends AppCompatAutoCompleteText
      */
     public void performCollapse(boolean hasFocus) {
         internalEditInProgress = true;
-        if (!hasFocus) {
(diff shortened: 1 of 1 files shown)


cketti (changes requested):
Using "Fixes #issue-number" will close the referenced issue once the pull request has been merged. However, this change only fixes one very specific case and not the entire problem. Please change the pull request description to use "see #7449" instead of "fixes".

The check for more than one token is performed after a bunch of work has been done that could have been avoided if there's only one token. Please move the check further up, to the "if (!hasFocus)" line.

Please also make sure to follow the code style. It's not automatically enforced for Java code. But there should be spaces around the comparison operator.

shamim-emon:
> Using "Fixes #issue-number" will close the referenced issue once the pull request has been merged. However, this change only fixes one very specific case and not the entire problem. Please change the pull request description to use "see #7449" instead of "fixes".
> 
> The check for more than one token is performed after a bunch of work has been done that could have been avoided if there's only one token. Please move the check further up, to the "if (!hasFocus)" line.
> 
> Please also make sure to follow the code style. It's not automatically enforced for Java code. But there should be spaces around the comparison operator.

@cketti done

cketti (approved):
Thanks :+1:

PR #8786, opened 2025-01-31
Add Release documentation

coreycb:
This adds documentation that describes the release train model, feature flags, versioning system, merge process, and uplift process.


docs/RELEASES.md
```diff
@@ -0,0 +1,422 @@
...
+```sh
+[Approval Request Comment]
+Original Issue/Pull request:
+Regression caused by (issue #):
+User impact if declined:
+Testing completed (on daily, etc.):
+Risk to taking this patch (and alternatives if risky):
+```
+
+## Create K-9 Mail releases
```
kewisch, line 191: This section describes the manual release process, mostly. I think it might be useful to keep in some form, let's say the apocalypse happens and GitHub actions are no longer available or certain steps fail, we'd know how to do it manually.

Otherwise I'd rather suggest this section describes how we trigger releases, what manual intervention is necessary, etc.

If you'd rather do this incrementally I'm fine with adding this section as long as we add a NOTE at the top of the section that we do things automatically now.

docs/RELEASES.md
```diff
@@ -0,0 +1,422 @@
...
+9. Review the release
+10. Configure a full rollout for beta versions
+11. On the Publishing overview page, click _Send change for review_
+12. Wait for the review to complete
+13. In case of a rejection, fix the issues and repeat the process
+
+### Release a stable version
+
+When the team decides the `main` branch is stable enough and it's time to release a new stable version, create a new
+maintenance branch (off `main`) using the desired version number with the last two digits dropped followed by `-MAINT`.
```
kewisch, line 325: This might also be confusing as we don't use -MAINT in the new system any more. See former comment though.

kewisch (approved):
r+wc, see comments below

wmontwe (commented):
@coreycb Looks good, just added some comments for clarification and some tickets to add the missing parts.

coreycb:
@wmontwe Thanks for the review and apologies for the delay. I'm pushing updates now based on your comments.

wmontwe (approved):
👍

PR #8806, opened 2025-02-10
Set maxLines and ellipsis on drawer folder labels

### LINKED ISSUE #8738: 2 lines max limit for folder names
### Checklist

### App

Thunderbird for Android

### App version

10.0

Other

### Android version

12

### Device model

_No response_

### Steps to reproduce

Open drawer

### Expected behavior

Show folder names limited to 2 lines, shortening otherwise with ellipsis in the middle

### Actual behavior

Drawer displays unlimited lines

Image

### Logs

_No response_

marcRDZ:
Fixes #8738

 - Add overflow and maxLines arguments to `TextLabelLarge` composables.
 - Set 2 max lines and ellipsis overflow on `NavigationDrawerItem`.


wmontwe (approved):
Thanks, this adds the ellipsis at the end at the moment, but it's  easy to be changed to [MiddleEllipsis](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/style/TextOverflow?hl=en#MiddleEllipsis()) once Jetpack Compose 1.8 is available, we might need to limit the text field to single line then.

PR #8878, opened 2025-03-03
Handle new changelog_master.xml directories in shippable builds workflow

coreycb:


.github/workflows/shippable_builds.yml
```diff
@@ -382,7 +383,21 @@ jobs:
...
+            case "$RELEASE_TYPE" in
+              "daily")
+                git add ./app-${APP_NAME}/src/daily/res/raw/changelog_master.xml
+                ;;
+              "beta")
+                git add ./app-${APP_NAME}/src/beta/res/raw/changelog_master.xml
+                ;;
+              "release")
+                git add ./app-${APP_NAME}/src/release/res/raw/changelog_master.xml
+                ;;
```
kewisch, line 398: Any reason not to use $RELEASE_TYPE in the path name here? 

Good to go either way though.

kewisch (approved):

.github/workflows/shippable_builds.yml
coreycb: That would have been a lot simpler.

coreycb:
[Approval Request Comment]
Original Issue/Pull request: https://github.com/thunderbird/thunderbird-android/pull/8878
Regression caused by (issue #): https://github.com/thunderbird/thunderbird-android/issues/8802
User impact if declined: beta and release builds will fail to "git add ./app-thunderbird/src/main/res/raw/changelog_master.xml" in "Release Commits" step.
Testing completed (on daily, etc.): Personal "Shippable Builds" runs
Risk to taking this patch (and alternatives if risky): Very low. I've run the workflow with the patches on all 3 branches and verified the paths are correct.

coreycb:
[Approval Request Comment]
Approved for release

PR #8885, opened 2025-03-05
Change demo backend to provide nested folders

### LINKED ISSUE #8886: Change demo backend to provide nested folders
This should help to validate and test nested folders with the demo account.

wmontwe:
This is changing the demo backend to provide nested folders for the account drawer improvements. While doing this I changed the JSON serializer from Moshi to KotlinX Serialization and refactored the code for better readability.

The demo data was changed to use `example.com` as domain and replace mentions of names by a neutral version. Also changed from mentioning K-9 to Thunderbird.

Fixes #8886


kewisch (approved):
r+ on this. One question though, the Kotlin files are in src/main/java, would they need to be moved to a kotlin-specific folder? It appears all the files are Kotlin.

wmontwe:
> r+ on this. One question though, the Kotlin files are in src/main/java, would they need to be moved to a kotlin-specific folder? It appears all the files are Kotlin.

Good catch, changed it.

PR #8887, opened 2025-03-05
Fix K-9 Mail badging

wmontwe:
The last translation update changed the K-9 Mail name and this updates the badging.


kewisch (approved):

app-k9mail/badging/fossRelease-badging.txt
```diff
@@ -54,15 +54,15 @@ application-label-nb:'K-9 e-post'
 application-label-nl:'K-9 Mail'
 application-label-nn:'K-9 e-post'
 application-label-pl:'K-9 Mail'
-application-label-pt:'K-9 Mail'
+application-label-pt:'Email K-9'
```
kewisch, line 57: Actually, this seems a bit weird. I think we'd want to keep the K-9 Mail format generally, and if needed localize the mail part. Can we reach out to the pt localizers?

wmontwe: Done

PR #8888, opened 2025-03-05
Change CI to use build command

### LINKED ISSUE #8937: Weblate update breaks the build
Some translation updates are breaking the build due to unintended changes to app names in K-9 Mail or introduction of non-compliant strings that don't follow Android guidelines. Although Weblate filters out most of these issues, there are still cases where it fails to catch them. Recently updated Android behavior has also introduced new challenges, as Weblate is not yet aware of these changes or lacks full support for specific language plural forms. To mitigate this, we need to employ lint rules to prevent such issues from occurring in the future.

See: #8887

wmontwe:
Change CI to use build instead of assemble to fully cover all app variants and catch errors early.

Reason: Last translation update broke the build for K-9 Mail due to changes to the app name.

This requires #8887 to be merged first to run successfully.

Fixes #8937

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -46,8 +46,5 @@ jobs:
       - name: Quality - Dependency Guard
         run: ./gradlew dependencyGuard
 
-      - name: Build
-        run: ./gradlew assembleDebug
-
-      - name: Test
-        run: ./gradlew testsOnCi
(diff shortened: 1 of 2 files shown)


coreycb (approved):

coreycb:
Waiting for https://github.com/thunderbird/thunderbird-android/pull/8887 to land before merging.

PR #8891, opened 2025-03-05
Fix Spotless setup

### LINKED ISSUE #7720: Markdown formatting inconsistent
I noticed that the markdown formatting of Android Studio and Spotless Markdown differ. Which leads to issues when autoformatting is enabled to run before a commit in Android Studio.

Both should produce similar formatting results to avoid unnecessary build failures.

wmontwe:
Spotless got really slow lately and this changes the configuration to improve speed.

Also bump the plugin from 6.25.0 -> 7.0.2 which allows us to enable Gradle configuration cache again.

Fixes #7720


core/ui/compose/designsystem/README.md
```diff
@@ -36,3 +36,4 @@ This can save time and improve the overall quality, as well as facilitate collab
 
 - [Atomic Design Methodology | Atomic Design by Brad Frost](https://atomicdesign.bradfrost.com/chapter-2/)
 - [Atomic Design: Getting Started | Blog | We Are Mobile First](https://www.wearemobilefirst.com/blog/atomic-design)
+
```
kewisch, line 39: Is this just adding a newline at EOF, or a full empty line? The latter seems a bit weird to me, the former I support. I guess a flexmark rule?

kewisch (approved):

core/ui/compose/designsystem/README.md
wmontwe: It's a flexmark rule to have a newline after the last list item + newline on end of file, not ideal but way faster than prettier before.

coreycb (approved):

PR #8892, opened 2025-03-05
Fix settings gradle config

wmontwe:
This fix addresses an issue in `settings.gradle.kts`, restricting dependency retrieval to only allowed registries while ensuring compatibility with Java 17.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -1,7 +1,13 @@
 pluginManagement {
     repositories {
         includeBuild("build-plugin")
-        google()
+        google {
+            content {
+                includeGroupByRegex("com\\.android.*")
+                includeGroupByRegex("com\\.google.*")
+                includeGroupByRegex("androidx.*")
+            }
+        }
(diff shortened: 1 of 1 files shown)


settings.gradle.kts
```diff
@@ -180,3 +202,15 @@ include(
...
 )
+
+check(JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_17)) {
+    """
+        Java 17+ is required to build Thunderbird for Android.
+        But it found an incompatible Java version ${{JavaVersion.current()}}.
+
+        Java Home: [${System.getProperty("java.home")}]
+
+        Please install Java 17 and set JAVA_HOME to the directory containing the Java 17 installation.
```
kewisch, line 213: Is it Java 17 specifically, or 17+?

kewisch (approved):

settings.gradle.kts
wmontwe: Fixed

Issue #8784, opened 2025-01-31
'Find Folder' option incorrectly appears in the 'Three-dot' menu on the 'Manage Folders' page when 'Find Folder' is already enabled


SuYanqi:
### Checklist

- [x] I have used the search function to see if someone else has already submitted the same bug report.
- [x] I will describe the problem with as much detail as possible.

### App

Thunderbird for Android

### App version

9.0b3

### Where did you get the app from?

None

### Android version

14

### Steps to reproduce

1. Click 'Hamburger' icon on the top-left
2. Click 'Manage Folders'
3. Click 'Find folder' icon
4. Click 'Three-dot' icon on the right side of screen

### Expected behavior

The 'Find folder' option should not appear in the 'Three-dot' menu when the 'Find folder' function is already enabled.

### Actual behavior

The 'Find folder' option still appears in the 'Three-dot' menu even after the 'Find folder' function has been enabled.

https://github.com/user-attachments/assets/9b6d7166-998a-4b26-aa18-01d06fba1671

rohnsha0:
was able to replicate!

Salkinnoma:
I think this can be solved by adding an implementation of the onPrepareOptionsMenu functions. Im taking a look at that

PR #8910, opened 2025-03-08
Issue #8784 

Salkinnoma:
Fix for Issue #8784: 'Find Folder' option incorrectly appears in the 'Three-dot' menu on the 'Manage Folders' page when 'Find Folder' is already enabled.

This issue was resolved by adding onPrepareOptionsMenu() in **ManageFoldersFragment**. This method dynamically toggles the visibility of folderMenuItem based on whether the search view is expanded or collapsed.

Additionally, an expand listener was added inside configureFolderSearchView() to ensure that folderMenuItem becomes visible again when the search view collapses, preventing it from incorrectly appearing in the overflow menu.

These changes ensure a smoother and more intuitive user experience by maintaining the correct visibility behavior for the Find Folder option.

[tfa_issue_8784.webm](https://github.com/user-attachments/assets/0ec9073b-2c1d-4a38-88bc-35e5c45162f0)


wmontwe (approved):
Thank you!

Issue #7646, opened 2024-02-15
Let's encrypt certificates stop working for Android 7.0 and earlier beginning Feb 8th, 2024


wmontwe:
Let's encrypt is phasing out their certificate [cross-signing](https://letsencrypt.org/2020/12/21/extending-android-compatibility.html) that allowed Android Devices prior to 7.1.1 to establish TLS connections.

They've announced that it will [phase out](https://letsencrypt.org/2023/07/10/cross-sign-expiration) begining Feb 8th, 2024. 

This will result in a `java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.` exception for every domain that renewed their certificate after that date or creates a new one.

We need to change to a trust store that trusts ISRG Root X1.

GottemHams:
@Benno-K Waiting for 1 sync round should also inform you of the problem. I eventually get notifications for every account about being unable to fetch mail due to an error, which I can tap to bring up the server settings for that account. Note that this is on **Android 14**, so it's not just an issue for older versions (anymore). My phone does in fact know the `ISRG Root X1` and `X2` certificates so it should be able to trust my mail server just fine.

Little log excerpt:
```
10-23 20:47:57.317 25501 25527 E MessagingController: Caused by: com.fsck.k9.mail.CertificateValidationException: javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.
```

Not quite sure if this is actually related to the original issue, but seeing as the error is still the same I would presume it is. Does TB on Android use its own cert store like the desktop edition?

voronind-com:
This is so awful. Why my whole system and other devices accept those LE certs, but Thunderbird for android silently fetches nothing? Please make an error at least in the main app that sync failed.

wmontwe:
We're fixing the certificate issues, but the notification need to be done separately as they require substancial changes, see #8831.

itsTyrion:
perhaps the title should be updated if this affects just Android in general, not just below version 7? I'm running into this issue on Android 15 (and used to on Android 14 IIRC)

PR #8952, opened 2025-03-13
Prepare fix lets encrypt support part2

wmontwe:
This is part of #7646 and changes some of the involved classes from Java to Kotlin to allow easier testing.


mail/testing/src/main/java/com/fsck/k9/mail/testing/StringHelper.kt
```diff
@@ -1,4 +1,4 @@
-package com.fsck.k9.mail
+package com.fsck.k9.mail.testing
```
kewisch, line 1: Does this need to go in a non-java folder?

kewisch: Same for a few other files

kewisch (approved):

legacy/core/src/test/java/com/fsck/k9/mailstore/MimePartStreamParserTest.kt
```diff
@@ -24,19 +24,19 @@ class MimePartStreamParserTest {
             To: <y@example.org>
             Subject: Testmail 1
             Content-Type: multipart/mixed; boundary=1
-            
+
```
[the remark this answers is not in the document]
wmontwe: Good catch, I think we need to move these examples out of the source code and provide them by dedicated sample files to avoid these issues.

mail/testing/src/main/java/com/fsck/k9/mail/testing/StringHelper.kt
wmontwe: The java source folder could host Java and Kotlin files. So as long there are still some Java files present it's not strictly necessary to separate them. The Kotlin folder on the other side can't contain Java files.

Using the kotlin source set prevents from accidentally adding Java files and the compiler could work a little bit more efficient.

PR #8954, opened 2025-03-13
Update release documentation for play store and milestones

kewisch:


docs/RELEASES.md
```diff
@@ -197,9 +205,17 @@ These are the general steps for a release:
...
+   - Beta versions only require a basic smoke test to ensure it installs
 6. Promote TfA and K-9 releases in Play Store.
+   - Betas are only released for TfA. K-9 beta users are advised to use Thunderbird.
 7. Wait for Play Store review to complete.
-8. Update F-Droid to new TfA and K-9 releases.
-9. Send notification to Matrix channels, and beta or planning mailing lists.
+   - Release versions of TfA and K-9 have managed publishing enabled. Once the review has completed you need to publish the release
+   - Beta versions of TfA do not have managed publishing enabled. It will be available once Google has reviewed, even on a weekend.
+8. Update F-Droid to new TfA and K-9 releases by sending a pull request to [fdroiddata](https://gitlab.com/fdroid/fdroiddata)
+9. Send community updates to Matrix channels, and beta or planning mailing lists.
```
coreycb, line 220: Would you like an email sent for each release or would this be as-needed?

kewisch: I think as needed is fine for now. For the beta list we'd post if there are specific things we'd like to test on. For planning lists we could consider notifying about releases, but I want to be clear on the benefit.

kewisch: Added "as needed" for now

coreycb:
@kewisch I'm not seeing your updates. I wonder if they didn't get pushed.

kewisch:
Ah yes, pushed to the wrong remote. How is this?

wmontwe (approved):
LGTM

Issue #8955, opened 2025-03-13
Validate gradle wrapper and workflows run twice on pull requests


kewisch:
These targets are run twice, once for the pull_request trigger, once for the push trigger. We probably only have to do this once.

Example https://github.com/thunderbird/thunderbird-android/pull/8954

![Image](https://github.com/user-attachments/assets/7d46c5f0-ac8f-4d60-894e-8e351cb31269)

Salkinnoma:
Ok so once per PR or once per push?

PR #8957, opened 2025-03-14
Fix Duplicate execution of workflows

Salkinnoma:
Fixes Issue #8955 
Validate Gradle Wrapper and validate-workflows are now executed only on PR

### DIFF
--- a/.github/workflows/validate-gradle.yml
+++ b/.github/workflows/validate-gradle.yml
@@ -1,7 +1,6 @@
 name: "Validate Gradle Wrapper"
 
 on:
-  push:
   pull_request:
 
 permissions:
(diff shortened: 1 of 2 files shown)


wmontwe (approved):
LGTM

coreycb (approved):

Salkinnoma:
So there is no Branch protection to prevent pushes directly to main?

kewisch:
There is, I guess we'd be safe on main since everything needs a pull request, good point. We have beta and release branches where we do uplifts from main and could theoretically push directly, which might need to happen if we have to change a patch due to merge conflicts. Am I being overcautious here?

wmontwe:
There are no pushes allowed to main, all changes have to be done by a pull-request.

Exceptions are beta and release branches where normal pushes are allowed, but force pushes are disabled. This could be changed where the uplifts are required to have a branch and code review as well.

PR #8958, opened 2025-03-16
Remove font size customization

### LINKED ISSUE #8949: Remove font size customization
As per #8420 we're considering to remove the font size customization and replace it with some more specific options to choose what the emphasis is. The font size selection is also broken in many ways, so it isn't quite as useful.

The removing bit would be a good first bug, and then we can do a follow up with design if someone is eager to follow on.

Please make sure to use feature flags for this, we'd definitely want to test this on beta for a while before we ship this to the majority of users.

shamim-emon:
- Fixes #8949

### DIFF
--- a/legacy/core/src/main/java/com/fsck/k9/K9.kt
+++ b/legacy/core/src/main/java/com/fsck/k9/K9.kt
@@ -2,6 +2,9 @@ package com.fsck.k9
 
 import android.content.Context
 import android.content.SharedPreferences
+import app.k9mail.core.featureflag.FeatureFlagProvider
+import app.k9mail.core.featureflag.FeatureFlagResult
+import app.k9mail.core.featureflag.toFeatureFlagKey
 import app.k9mail.feature.telemetry.api.TelemetryManager
 import app.k9mail.legacy.account.Account
 import app.k9mail.legacy.account.Account.SortType
@@ -13,6 +16,7 @@ import com.fsck.k9.preferences.RealGeneralSettingsManager
 import com.fsck.k9.preferences.Storage
 import com.fsck.k9.preferences.StorageEditor
 import kotlinx.datetime.Clock
+import org.koin.android.ext.android.inject
 import org.koin.core.component.KoinComponent
 import org.koin.core.component.inject
 import timber.log.Timber
@@ -22,6 +26,7 @@ import timber.log.Timber.DebugTree
 object K9 : KoinComponent {
     private val generalSettingsManager: RealGeneralSettingsManager by inject()
     private val telemetryManager: TelemetryManager by inject()
+    private val featureFlagProvider: FeatureFlagProvider by inject()
 
     /**
      * If this is `true`, various development settings will be enabled.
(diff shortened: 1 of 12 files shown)


legacy/core/src/main/java/com/fsck/k9/K9.kt
```diff
@@ -383,7 +388,15 @@ object K9 : KoinComponent {
...
-        fontSizes.load(storage)
+
+        val disableFontSizeConfig = when (featureFlagProvider.provide("disable_font_size_config".toFeatureFlagKey())) {
+            FeatureFlagResult.Enabled -> true
+            FeatureFlagResult.Disabled -> false
+            FeatureFlagResult.Unavailable -> false
+        }
+        if (!disableFontSizeConfig) {
+            fontSizes.load(storage)
+        }
```
wmontwe, line 399: This could be

```suggestion
        featureFlagProvider.provide("disable_font_size_config".toFeatureFlagKey())
            .onDisabledOrUnavailable {
                fontSizes.load(storage)
            }
```

legacy/core/src/test/java/com/fsck/k9/TestApp.kt
```diff
@@ -44,4 +48,13 @@ val testModule = module {
     single { mock<NotificationStrategy>() }
     single(named("controllerExtensions")) { emptyList<ControllerExtension>() }
     single { mock<WorkManager>() }
+    single<FeatureFlagProvider> {
+        InMemoryFeatureFlagProvider(
+            featureFlagFactory = {
+                listOf(
+                    FeatureFlag(key = FeatureFlagKey("feature1"), enabled = true),
+                )
```
wmontwe, line 56: I think an emptyList should be enough.

```suggestion
                emptyList<FeatureFlag>()
```

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/general/GeneralSettingsFragment.kt
```diff
@@ -37,6 +43,21 @@ class GeneralSettingsFragment : PreferenceFragmentCompat() {
...
+        }
+
+        if (disableFontSizeConfig) {
+            val parentPreference = findPreference<PreferenceCategory>("global_preferences")
+            val fontSizePreferenceScreen = findPreference<PreferenceScreen>("font_size")
+
+            if (parentPreference != null && fontSizePreferenceScreen != null) {
+                parentPreference.removePreference(fontSizePreferenceScreen)
+            }
+        }
```
wmontwe, line 59: Could be replaced by:

```suggestion
        featureFlagProvider.provide("disable_font_size_config".toFeatureFlagKey())
            .onEnabled {
                val parentPreference = findPreference<PreferenceCategory>("global_preferences")
                val fontSizePreferenceScreen = findPreference<PreferenceScreen>("font_size")

                if (parentPreference != null && fontSizePreferenceScreen != null) {
                    parentPreference.removePreference(fontSizePreferenceScreen)
                }
            }
```

wmontwe (changes requested):
Thanks, the `FeatureFlagResult` has methods to simplify handling, like `onEnabled` and `onDisabledOrUnavailable`.

Also there is no need to do so many commits, the changes are all related to adding the feature flag and should be one.

wmontwe:
@shamim-emon I added a git commit guide, please have a look and comments are welcome.

shamim-emon:
> Thanks, the `FeatureFlagResult` has methods to simplify handling, like `onEnabled` and `onDisabledOrUnavailable`.
> 
> Also there is no need to do so many commits, the changes are all related to adding the feature flag and should be one.
> 
> We aim to group related changes into a commit and split them up when the change is to big or touches different aspects of the source code. Additionally, commits could be used to help the reviewer or whoever retroactively looks into the changes to understand how an existing functionality was changed to new behaviour.

@wmontwe Updated accordingly

wmontwe:
@shamim-emon Just one last thing, rebase your changes on main instead of merging them into your branch, please.

wmontwe:
Thanks, just the debug flags are missing.

shamim-emon:
> @shamim-emon Just some last things, rebase your changes on main instead of merging them into your branch, please. The second commit should be named after what it's changing. In this case: `remove unused properties from K9 class`.
> 
> The the flag should be set for K-9 debug, Thunderbird debug, daily and beta.

@wmontwe Done

wmontwe (approved):
Thank you!

PR #8960, opened 2025-03-18
Use botmobile for commenting

kewisch:
@coreycb draft pull request to switch over to botmobile for all actions once https://github.com/thunderbird/cloudops/issues/30 is resolved. I need to update the variable names once the following question is answered.

Where do we put the secrets for this? 
- For the release right now they are in the publish_release environment, which makes sense. 
- To be able to use that username for the earlier release bump commit, we'd also need to put it into the respective  `thunderbird_beta` and `thunderbird_release` environments.
- To be able to use the username for commenting and the deploy docs, we'd either need yet another environment.

Or, we put them as repository secrets, which would however make them available to all environments. Contributors in pull requests would still not get the secrets. 

If we're able to lock down the permissions so that all that is available is repository contents, issues, pull requests, then it might be a bit more safe.

Let me know what you recommend in terms of security.

### DIFF
--- a/.github/workflows/uplift-merges.yml
+++ b/.github/workflows/uplift-merges.yml
@@ -11,28 +11,41 @@ on:
 permissions:
   contents: read
 
+environment: botmobile
+
 jobs:
   uplift:
     name: Uplift
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
-      contents: write
     steps:
+      - name: App token generate
+        uses: actions/create-github-app-token@21cfef2b496dd8ef5b904c159339626a10ad380e # v1.11.6
+        if: ${{ !inputs.dryRun && vars.BOT_CLIENT_ID }}
+        id: app-token
+        with:
+          app-id: ${{ vars.BOT_CLIENT_ID }}
+          private-key: ${{ secrets.BOT_PRIVATE_KEY }}
+
       - name: Checkout repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
+          token: ${{ steps.app-token.outputs.token || github.token }}
 
       - name: Configure for push
         if: ${{ !inputs.dryRun }}
+        env:
+          APP_SLUG: ${{ steps.app-token.outputs.app-slug || 'github-actions'}}
+          APP_USER_ID: ${{ vars.BOT_USER_ID || '41898282' }}
         run: |
-          git config --global user.name "GitHub Actions Bot"
-          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "${APP_SLUG}"
+          git config --global user.email "${APP_USER_ID}+${APP_SLUG}[bot]@users.noreply.github.com"
 
       - name: Run uplift script
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ steps.app-token.outputs.token || github.token }}
           DRYRUN: ${{ !inputs.dryRun && '--no-dry-run' || '' }}
           BRANCH: ${{ github.ref_name }}
           PUSH: ${{ !inputs.dryRun && '--push' || '' }}
--- a/scripts/ci/setup_release_automation
+++ b/scripts/ci/setup_release_automation
@@ -168,7 +168,8 @@ def get_github_public_key(repo, environment_name):
     }
     response = requests.get(url, headers=headers)
     if response.status_code == 200:
-        return response.json()
+        data = response.json()
+        return [data["key_id"], data["key"]]
     else:
         raise Exception(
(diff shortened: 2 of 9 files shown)


kewisch:
This is now ready for review with the open question on where to put the secrets. Botmobile is ready and we've reduced permissions to contents and issues/prs.

coreycb:
For more context, botmobile has the following permissions:
- Read access to commit statuses and metadata
- Read and write access to code, issues, and pull requests 
only on github.com/thunderbird/thunderbird-android repository.

Write access to code is the major one that concerns me. We could consider not pushing at all in any of our workflows, and instead creating PRs that require manual review. That would slow us down a bit, but we should be reviewing any automated commits anyway, and this would force us to do that. That way we can remove the permission granting write access to code, and I'd be more comfortable with switching to repository secrets.

kewisch:
Our current `thunderbird-mobile-gh-releaser-bot` has write access to code, I think we could certainly separate that out into making pull requests, though I'd like to do that separately also because I'm not sure how much effort we should put into the GH actions process given the TC migration.

If you want to keep the "read and write access to issues and pull requests" separate from the "read and write access to code" pieces, we could continue using the `thunderbird-mobile-gh-releaser-bot` for releases (and maybe rename to `thunderbird-botmobile-release`), and simply use `thunderbird-botmobile` for actions related to commenting that are more user-visible in this PR.

Aesthetically this isn't quite what I was aiming for, I was hoping that we could use botmobile for all bot related actions so it would show in commits and releases as well. But I understand security is more important than optics.

Ultimately though, even if we don't use botmobile, the github-actions bot already has commit access in the respective jobs due to `permissions` having `contents: write` in some spots, so if something goes wrong the attacker doesn't need to make the effort to use thunderbird-botmobile, they could simply use the github-actions bot.

coreycb:
I'm good with landing what you have here for now and then a follow-up to switch pushes to PRs, and drop write access to code from botmobile privileges. I believe `thunderbird-mobile-gh-releaser-bot` will be obsoleted by botmobile so we can remove it once this is merged, and clean up any instances of `contents: write`?

kewisch:
@coreycb could you give this another look? If we are for now using botmobile for everything I needed to adjust the environments a bit. I'm also using the token for pulls/pushes now, which means we can remove the `contents: write`. We can afterwards sunset thunderbird-mobile-gh-releaser-bot.

.github/workflows/uplift-merges.yml
```diff
@@ -11,28 +11,41 @@ on:
 permissions:
   contents: read
 
+environment: botmobile
```
coreycb, line 14: Is shippable_builds missing the botmobile environment?

scripts/ci/setup_release_automation
```diff
@@ -448,14 +447,41 @@ def create_signing_environment(repo, environment, branch, props_file):
...
 
+def make_bot_environment(repo, environment):
+    key_id, public_key = get_github_public_key(repo, environment)
+
+    with open("botmobile.key.pem") as fp:
+        encrypted_bot_key = encrypt_secret(public_key, fp.read())
+    with open("botmobile.clientid.txt") as fp:
+        bot_client_id = fp.read().strip()
+    with open("botmobile.userid.txt") as fp:
+        bot_client_id = fp.read().strip()
```
coreycb, line 458: Should this be bot_user_id?

scripts/ci/setup_release_automation
```diff
@@ -448,14 +447,41 @@ def create_signing_environment(repo, environment, branch, props_file):
...
+
+
+    set_github_environment_secret(
+        repo, "BOT_PRIVATE_KEY", encrypted_bot_key, key_id, environment
+    )
+
+    set_github_environment_variable(repo, "BOT_CLIENT_ID", bot_client_id, environment)
+    set_github_environment_variable(repo, "BOT_USER_ID", bot_client_id, environment)
```
coreycb, line 466: Same here

coreycb (commented):
I added a couple of comments. And also created https://github.com/thunderbird/thunderbird-android/issues/8990.

.github/workflows/uplift-merges.yml
kewisch: It is not. shippable_builds only needs botmobile in certain environments, but it does need them in places where there is already an existing one, like `thunderbird_beta`. There can only be one environment. So I need to add the keys to each of these environments as well. This is what I was alluding to in my initial comments.

scripts/ci/setup_release_automation
kewisch: Yes thank you!

.github/workflows/uplift-merges.yml
coreycb: That's right. For now at least I think we should keep separate environments.

coreycb:
LGTM

coreycb:
This was included in today's merge of main->beta

PR #8961, opened 2025-03-18
Subfolder for CI docs and scripts

kewisch:
* Move docs into a subdirectory so we can separate CODEOWNERS
* Clean up some stray files for a better structure
* Add link to rendered documentation in the main README so people browsing source will see it

I hope I didn't miss any places where script execution paths need to be renamed.


docs/README.md
```diff
@@ -1,3 +1,3 @@
-# About Thunderbird for Android
+# Thunderbird for Android Documentation
 
-This document provides an overview of the Thunderbird for Android project.
+The latest available documentation is rendered at: [https://thunderbird.github.io/thunderbird-android/docs/latest/](https://thunderbird.github.io/thunderbird-android/docs/latest/)
```
wmontwe, line 3: This should be part of the main readme, as this file is the landing page for the docs and would render a link to itself.

docs/ci/AUTOMATION.md
wmontwe, line 1: I see that you've been changing names to uppercase, should we do this for all files and add a mention in the `how-to-document`?

docs/assets/activity_diagram.graphml
wmontwe, line 1: I see that you've been moving resources to an assets folder relative to the document, should we do this for all files and add a mention in the how-to-document?

wmontwe (changes requested):
I think we should be consisten when naming files and where to place resources needed for the documents.

docs/README.md
kewisch: I was wondering about this. I'm trying to avoid the situation where a user stumbles into the docs folder on github, and simply sees the empty README. I know the rendered version would link to itself, but I think this is a tradeoff that is ok to make. What do you think?

docs/ci/AUTOMATION.md
kewisch: I don't really have a preference, what do you think? Uppercase is always like shouting, but then again some files like README and SUMMARY simply are uppercase.

docs/assets/activity_diagram.graphml
kewisch: I separated assets for the main docs and docs/ci so that if releng adds an asset they can do so themselves. I don't think we need more of a split right now, but I'm certainly open to doing so if you prefer.

docs/README.md
wmontwe: The current README needs more content, at the moment it's just a placeholder and should transform into a real landing page, guiding where to look for which aspect of the documentation.

But we could remove the README from the SUMMARY and add the landing page as separate document.

docs/ci/AUTOMATION.md
wmontwe: The uppercase is a GitHub convention, but only used to emphasize certain documents like README, LICENSE, CONTRIBUTING, SECURITY. SUMMARY is uppercase by convention in mdbook.

Otherwise it's a free choice.

docs/assets/activity_diagram.graphml
wmontwe: Fine, I tend to keep assets close to the document they are used in. But that's not really important.

docs/README.md
kewisch: Let's take care of the landing page a bit later, I don't really have thoughts on content for it yet that wouldn't duplicate the main README. At this time we can also consider if there are aspects we should integrate into DTN.

For now I removed the main README, making the contributing page the landing page, which isn't too bad.

docs/ci/AUTOMATION.md
kewisch: Ok, so then we'd only really have consistency if we make it all uppercase everywhere. I'm going to go for all uppercase in the top level, and then normal in the subfolders aside from the README.md. Seems the least worst approach.

docs/assets/activity_diagram.graphml
kewisch: Then we are close enough I think. The png files are for DESIGN.md which is in the toplevel. The graphml file is sources, and the release files have their own assets folder. I added something to the howto

wmontwe (approved):
👍

coreycb (approved):

coreycb:
Uplifted to beta https://github.com/thunderbird/thunderbird-android/commit/52529411bef91ce480fcc4aec3275b321474d032

PR #8965, opened 2025-03-18
Add a Git commit guide

### LINKED ISSUE #8964: Write Git Commit Guide
As we've seen a broad range of contributions and different styles of creating git commits, we should provide a guide to help contributors crafting meaningfull commits.

wmontwe:
Resolves #8964

### DIFF
--- a/docs/contributing/git-commit-guide.md
+++ b/docs/contributing/git-commit-guide.md
@@ -1,0 +1,106 @@
+# Git Commit Guide
+
+Use [Conventional Commits](https://www.conventionalcommits.org/) to write consistent and meaningful commit messages.
+This makes your work easier to review, track, and maintain for everyone involved in the project.
+
+## ✍️ Commit Message Format
+
+```git
+<type>(<scope>): <description>
+
+<body>
+
+<footer(s)>
+```
+
+Components:
+
+- `<type>`: The [type of change](#-commit-types) being made (e.g., feat, fix, docs).
+- `<scope>` **(optional)**: The [scope](#optional-scope) indicates the area of the codebase affected by the change (e.g., auth, ui).
+- `<description>`: Short description of the change (50 characters or less)
+- `<body>` **(optional)**: Explain what changed and why, include context if helpful.
+- `<footer(s)>` **(optional)**: Include issue references, breaking changes, etc.
+
+### Examples
+
+Basic:
+
+```git
+feat: add QR code scanner
+```
+
+With scope:
+
+```git
+feat(auth): add login functionality
+```
+
+With body and issue reference:
+
+```git
+fix(api): handle null response from login endpoint
+
+Checks for missing tokens to prevent app crash during login.
+
+Fixes #123
+```
+
+### 🏷️ Commit Types
+
+|    Type    |            Use for...            |                  Example                  |
+|------------|----------------------------------|-------------------------------------------|
+| `feat`     | New features                     | `feat(camera): add zoom support`          |
+| `fix`      | Bug fixes                        | `fix(auth): handle empty username crash`  |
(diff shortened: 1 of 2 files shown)


docs/contributing/git-commit-guide.md
```diff
@@ -0,0 +1,90 @@
...
+### 4. Write Descriptive Commit Messages
+
+A good commit message follows this structure:
+
+- **Title (short summary, 50 characters max)**
+- **Body (optional, but recommended for context)**
+  - Explain *what* the change does and *why* it was needed.
+  - Describe any important details, such as edge cases addressed.
+  - Reference relevant issue numbers or discussions.
+
```
shamim-emon, line 68: Adding an example for commit  message that follows above structure would be helpful.

docs/contributing/git-commit-guide.md
```diff
@@ -0,0 +1,90 @@
...
+Avoid generic descriptions like "Fix bug" or "Update code"
+
+### 5. Ensure Commits Are Self-Contained
+
+Each commit should:
+
+- Be complete and independently buildable.
+- Not introduce temporary hacks or broken functionality.
+- Ideally, pass tests to maintain project integrity.
```
kewisch, line 77: You mention most of these before, potentially we could consolidate here to not make the guide too long. In section 2 you mention each commit should be functional and not leave it in a broken state. This covers the first item, and potentially the third, and with a light stretch also the second.

docs/contributing/git-commit-guide.md
```diff
@@ -0,0 +1,90 @@
...
+Consider the commit history as a documentation tool. Someone reviewing past changes should be able to:
+
+- Follow the logical evolution of a feature or fix.
+- Understand *how* and *why* a functionality changed over time.
+- Easily revert a commit if needed without breaking unrelated functionality.
+
+### 7. Use Atomic Commits
+
+Atomic commits mean that each commit represents a single unit of change that can be reverted independently. This
+practice ensures that if an issue arises, it is easier to pinpoint and undo the specific problematic change.
```
kewisch, line 90: Based on how we've worked before I had the understanding that we're working atomic at the level of a merge, but that individual commits within the merge might not be completely functional.

I think this is worth clarifying. If we really require atomic commits to each individual level, then we might as well do rebases instead of merges because each commit should work as it is and we can just back out the offending commit.

If we're working at the merges level, then while we shouldn't intentionally break between commits, we'd be backing out the merge and bisecting based on merges.

There is also some overlap between (6) and (2) where we're again saying it needs to work inbetween commits.

kewisch (changes requested):

docs/contributing/git-commit-guide.md
wmontwe: This doesn't mean to have something completely functional, but it should compile, tests should succeed and all quality tools issues should be resolved. This is meant to prevent commits like:

1. Add funtionality
2. Add test for functionality
3. Fix broken tests
4. Fix linter issues
5. Fix code formatting
6. ...

These changes should be part of one commit.

docs/contributing/git-commit-guide.md
wmontwe: Yes, I think having examples would help.

docs/contributing/git-commit-guide.md
wmontwe: I'll rework the sections.

docs/contributing/git-commit-guide.md
wmontwe: This is a special case. If it is detectable using our build, quality checks and tests then yes, otherwise I think it's fine.

shamim-emon (commented):
LGTM!

PR #8967, opened 2025-03-18
Simplify Feature Flag Result Handling on Achive Operations

shamim-emon:
- Added a `get` function in `FeatureFlagResult` to return values based on its sealed objects (`Enabled`, `Disabled`, `Unavailable`).  
- Integrated the `get` function in `ArchiveOperations` for handling `FeatureFlagResult`.


core/featureflags/src/main/kotlin/app/k9mail/core/featureflag/FeatureFlagResult.kt
```diff
@@ -5,6 +5,14 @@ sealed interface FeatureFlagResult {
     data object Disabled : FeatureFlagResult
     data object Unavailable : FeatureFlagResult
 
+    fun <T> get(
```
wmontwe, line 8: It think `get` is not the right function name. What about `fold` from functional programming, which could be used when reducing or combining data. Or `map`, but it's more related to lists. A long version `whenEnabledOrNot`. What do you think?

shamim-emon:
> @shamim-emon Looks good. Could you just rebase your commits. The `Refactor FeatureFlagResult's function ...` is not necessary as the code wasn't part of `main` yet. The commits for new code should be limited to adding new functionality and describe the necessary parts. A separate commit for adding tests is ideally also not necessary.

@wmontwe Done!

wmontwe (approved):
Thank you!

PR #8982, opened 2025-03-24
Update Java Version to 21

wmontwe:
Change Java to 21 to fix some compatibility issues on macOS.

Remove unused AndroidX DrawerLayout.


kewisch:
Does this mean we're increasing the minimum java version for this project? We have a few places like https://github.com/thunderbird/thunderbird-android/blob/dd2c716d0540e330bd004750ec690b7abc6df091/settings.gradle.kts#L210 where we note Java 17.

wmontwe:
> Does this mean we're increasing the minimum java version for this project? We have a few places like
> 
> https://github.com/thunderbird/thunderbird-android/blob/dd2c716d0540e330bd004750ec690b7abc6df091/settings.gradle.kts#L210
> where we note Java 17.

It's complicated, the source code needs to be build with Java 11 compatibility, defined in `ThunderbirdProjectConfig`. Then there is the Android Gradle Plugin, which requires at least Java 17 or any higher Java version that is still compatible with Java 17, hence the check.

As I have issues with my Java 17 version, I increased the [jenv](https://github.com/jenv/jenv) project local version to 21.

kewisch (approved):

kewisch:
We have a few github actions that also run java 17, should we be upgrading those to 21? Maybe we can use the `java-version-file` argument for the `setup-java` action instead.

wmontwe:
> We have a few github actions that also run java 17, should we be upgrading those to 21? Maybe we can use the `java-version-file` argument for the `setup-java` action instead.

Not yet, there is a different mechanism to manage the used JDK version. I would like to explore that instead and create a ticket if it makes sense for the project.

PR #8994, opened 2025-03-26
Message compose: support completion of list of recipients

### LINKED ISSUE #5199: Comma Delimited Copy/Paste Addresses
Currently, and since at least 2017, K9 has been unable to parse multiple emails out of a comma-delimited list.  This was added to a feature request list in 2018, but no action appears to have been taken to facilitate.

Simply put, I need to be able to *either* Reply-to-All when it's a BCC list, OR Copy/Paste/Auto-Parse a batch of comma-delimited email addresses.

The only alternative is to copy the group into a notepad, then copy/paste them one-by-one back into K9.  This isn't a solution, and unfortunately will result in necessitating the move to another email client after several years of being a big fan of this one.

ruuzia:
Thank you for creating a wonderful piece of software!
This fixes #5199, which I chose to implement because it was tagged "good first issue".

This PR adds support to the message compose screen for completing entire lists of recipient addresses, rather than just a single address. This way, the user can paste in comma/semicolon delimited address lists. It uses the existing functionality of parsing lists of addresses.

The other way of implementing the feature of pasting in lists of addresses is overriding the paste behavior for the input, and then virtually inserting the pasted characters one by one. However, I considered this method pretty hacky, and I also wanted to use the existing support for parsing lists of addresses.

Unfortunately, this could not be implemented in Kotlin as I was not in the mood to refactor ~2000 lines of Java.

[android_thunderbird_compose_paste_list_demo.webm](https://github.com/user-attachments/assets/514d0110-85c4-4d82-b108-450aac96a1ad)

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/view/RecipientSelectView.java
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/view/RecipientSelectView.java
@@ -5,6 +5,7 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.List;
 
 import android.annotation.SuppressLint;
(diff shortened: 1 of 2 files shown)


wmontwe:
> Unfortunately, this could not be implemented in Kotlin as I was not in the mood to refactor ~2000 lines of Java.

Unfortunately, this is on us I guess

kewisch:
@ruuzia I know it might be a little bit of extra work, but we would much appreciate if you could take a leap and refactor that file. We're trying to get rid of our remaining java, and it would make a difference!

ruuzia:
Sure thing @kewisch, I can jump on that!

PR #9000, opened 2025-03-28
Remove un-used com.fsck.k9.glide.K9AppGlideModule

shamim-emon:
- `com.fsck.k9.glide.K9AppGlideModule` is no longer being used. So, this PR removes it.


wmontwe:
This is still needed, I'll revert the change.

PR #9010, opened 2025-03-30
Refactor: Move legacy:search to feature:search

shamim-emon:
- Move` :legacy:search` to `:feature:search`
- Move `app.k9mail.legacy.account.BaseAccount` to `core:account`


legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListConfig.kt
```diff
@@ -2,10 +2,10 @@ package com.fsck.k9.ui.messagelist
 
 import app.k9mail.legacy.account.SortType
 import app.k9mail.legacy.message.controller.MessageReference
-import app.k9mail.legacy.search.LocalSearch
+import net.thunderbird.feature.search.LocalSearch
 
 data class MessageListConfig(
-    val search: LocalSearch,
+    val search: net.thunderbird.feature.search.LocalSearch,
```
asoucar, line 8: This syntax usage is not familiar to me, is there some reason it wouldn't be `LocalSearch` instead of `net.thunderbird...`? If its a duplicate name issue the `as` keyword could be used

asoucar (commented):
I had the one actual code question but otherwise the code looks fine and like it would work. I don't feel comfortable enough with my understanding of the expected behavior of the search functionality at this moment to actually approve

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListConfig.kt
shamim-emon: @asoucar fixed it.

kewisch (approved):

PR #8956, opened 2025-03-14
Add basic support for subfolders in Thunderbird Android

ClemRiviere:
**Description**

This PR introduces basic support for email subfolders in Thunderbird Android (#630). Currently, subfolders can be managed up to a depth of 2 levels, which is hardcoded in the implementation. Feedback on a more flexible approach to handling deeper folder structures would be greatly appreciated.

Since this is my first contribution to the project, I would love to get some feedback on the approach taken. This PR is likely not a final version, and I'm open to suggestions for improvements.

**Changes**

- Added support for email subfolders.
- Implemented a basic hierarchy with a maximum depth of 2 levels (hardcoded).
- UI adjustments to display subfolders appropriately.

**Visual Example**

image

**Feedback Requested**

- Is there a preferred way to handle unlimited or more flexible folder depths in this project?
- Any recommendations on code structure or performance improvements?
- General feedback on the implementation is welcome!

Looking forward to your feedback!

PR #9045, opened 2025-04-04
Reworked: Add basic support for subfolders in Thunderbird Android 

wmontwe:
This is a rework of #8956 with tests and added to the `dropdown` version of the drawer. This version of the drawer is behind a feature flag and enabled for `debug` and `daily` at the moment.

Depends on #9044


kewisch (approved):

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/domain/usecase/GetDisplayTreeFolder.kt
```diff
@@ -0,0 +1,103 @@
...
+            displayFolder = null,
+            displayName = null,
+            totalUnreadCount = accountFolderTreeList.sumOf { it.totalUnreadCount },
+            totalStarredCount = accountFolderTreeList.sumOf { it.totalStarredCount },
+            children = (unifiedFolderTreeList + accountFolderTreeList).toImmutableList(),
+        )
+    }
+
+    private fun flattenPath(folderName: String, maxDepth: Int): List<String> {
+        val parts = folderName.split("/").map { it.takeIf { it.isNotBlank() } ?: "(Unnamed)" }
```
da-anda, line 43: I know I am late because this is already merged (thank you so much btw), but could the separator be made configurable? I am not familiar with the IMAP spec, but at least my mail provider seems to use `.` as separator for foldernames instead of `/`. A workaround/fix for this could be to "normalize" the folder names by simply replacing `.` with `/` before the tree is generated and keep using `/` throughout the code.

side note: using a const for the separator sign might be a good idea.

kewisch: Please file a new issue on this, we'd have to create a new PR anyway :)

Issue #8231, opened 2024-10-03
External keyboard gets stuck on the address fields on Compose screen


dankeck:
### Checklist

- [X] I have used the search function to see if someone else has already submitted the same bug report.
- [X] I will describe the problem with as much detail as possible.

### App version

8.0b1

### Where did you get the app from?

Google Play

### Android version

14

### Device model

Pixel 5

### Steps to reproduce

1. Connect a wireless keyboard to Android device.
2. Open Thunderbird Beta.
3. Compose an email.
4. Focus on the To field, Cc field, or Bcc field.
5. Hit the tab key on the keyboard.

### Expected behavior

Keyboard focus proceeds to the next field.

### Actual behavior

Keyboard focus remains in the field.

asoucar:
Taking this on

wmontwe:
This is not taken at the moment.

snowykte0426:
I think I found why this happens, and it is not in the layout.

`RecipientSelectView` extends `TokenCompleteTextView`, the vendored copy under `library/token-auto-complete`, and its `onKeyDown` catches `KEYCODE_TAB` along with ENTER and DPAD_CENTER:

    case KeyEvent.KEYCODE_TAB:
    case KeyEvent.KEYCODE_ENTER:
    case KeyEvent.KEYCODE_DPAD_CENTER:
        if (event.hasNoModifiers()) {
            shouldFocusNext = true;
            handled = true;
        }

Returning `handled` means the key event never reaches the framework's focus navigation, so focus cannot move. On the following `onKeyUp` that flag leads to `handleDone()`, which completes the pending token and hides the soft keyboard, but despite the name it never actually moves focus anywhere. That also explains the earlier comment about Shift+Tab working: with a modifier held `hasNoModifiers()` is false, the event is not swallowed, and the default backwards focus search runs.

It lines up with what happened in #9058 too. The `focusable` attribute added there only ended up on `reply_to`, and this issue was dropped from that PR's Fixes list because it appeared to work locally but not on some devices. Adding focusable attributes would not have fixed it either way, since the event is consumed before focus is ever considered.

[...]

PR #9058, opened 2025-04-10
Fix known keyboard navigation bugs

### LINKED ISSUE #8230: Compose button not focusable by keyboard
### Checklist

### App version

8.0b1

Google Play

### Android version

14

### Device model

Pixel 5

### Steps to reproduce

1. Connect a wireless keyboard to Android device.
2. Open Thunderbird Beta.
3. While viewing the Inbox:
- Hit the tab key on the keyboard multiple times.
- Hit ctrl + tab multiple times.

### Expected behavior

Eventually, keyboard focus will reach the floating compose button using either tab or ctrl + tab.

### Actual behavior

Keyboard focus never reaches the floating compose button.

### Logs

_No response_

asoucar:
Explicitly adding the `focusable` field to the RecipientSelectView allows it to be navigated through via arrow keys. 
Explicitly adding the `focusable` field to the Floating Action Button for Compose adds it to the tab navigation. Moving it to the first child of the parent Layout Coordinator allows for it to be focused before the messages in the list while retaining its position. I will say the focus on the compose button is very faint and hard to see. It may be worth investigating during a future refactor
Using Ctrl-Tab shifts tab focus to the top navigation bar and its assorted buttons.

Fixes #8230, #5852 
Edit: Removed `#8321` from fixes list as it appears to work locally but did not work on some Samsung devices, which I don't have access to for testing

### DIFF
--- a/legacy/ui/legacy/src/main/res/layout/message_compose_recipients.xml
+++ b/legacy/ui/legacy/src/main/res/layout/message_compose_recipients.xml
@@ -252,6 +252,7 @@
             android:background="@android:color/transparent"
             android:paddingTop="10dp"
             android:paddingBottom="10dp"
+            android:focusable="true"
             android:dropDownWidth="wrap_content"
             style="@style/RecipientEditText"
             />
--- a/legacy/ui/legacy/src/main/res/layout/message_list_fragment.xml
+++ b/legacy/ui/legacy/src/main/res/layout/message_list_fragment.xml
@@ -8,6 +8,17 @@
     android:layout_height="match_parent"
     tools:context=".messagelist.MessageListFragment"
     >
+    <com.google.android.material.floatingactionbutton.FloatingActionButton
+        android:id="@+id/floating_action_button"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="bottom|end"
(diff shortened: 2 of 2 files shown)


kewisch:
I'm able to confirm #8230 is solved, it is possible to focus the compose button with Shift+Tab as you mentioned. I'd say it is still not ideal, if you shift tab once too often you end up in the middle of the list and there is no way to get back to the compose button aside from going all the way up again. I'm not very familar with tab navigation on Android, but would it be reasonable behavior that Ctrl+Tab switches between toolbar, message list and compose button?

#8231 is still an issue for me. I am in the compose screen and can tab forward from the "From" line to the "To" line, but from there I cannot get to the subject or message text. I tried both tabbing forward and the arrow keys. My setup is an Apple Magic Keyboard connected via USB to a Samsung device. Did I miss anything here?

Confirming also #5852 is fixed with Ctrl+Tab.

asoucar:
@kewisch I was not able to simply restructure the views to get the floating compose button to be always quickly reachable via tab unfortunately. I think a larger restructure will be necessary. I think it is worth merging this fix for what it does and make a note of the rest for future work.
I am also removing #8231 from the fixes list as I cannot reproduce with my current emulator setup.

kewisch:
Sounds good, thanks for checking!

kewisch (approved):

PR #9061, opened 2025-04-14
Move :legacy:testing.testing to core module

shamim-emon:
- Moves ` :legacy:testing.testing` to core module

### DIFF
--- a/core/android/testing/build.gradle.kts
+++ b/core/android/testing/build.gradle.kts
@@ -3,10 +3,16 @@ plugins {
 }
 
 android {
-    namespace = "app.k9mail.core.android.testing"
+    namespace = "net.thunderbird.core.android.testing"
 }
 
 dependencies {
     api(libs.junit)
     api(libs.robolectric)
+
+    implementation(projects.legacy.core)
+
+    api(libs.koin.core)
+    api(libs.mockito.core)
+    api(libs.mockito.kotlin)
 }
--- a/core/android/testing/src/main/kotlin/net/thunderbird/core/android/testing/MockHelper.kt
+++ b/core/android/testing/src/main/kotlin/net/thunderbird/core/android/testing/MockHelper.kt
@@ -1,4 +1,4 @@
-package com.fsck.k9.testing
+package net.thunderbird.core.android.testing
 
 import org.mockito.Mockito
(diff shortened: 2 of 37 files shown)


core/android/testing/build.gradle.kts
```diff
@@ -9,4 +9,10 @@ android {
 dependencies {
     api(libs.junit)
     api(libs.robolectric)
+
+    implementation(projects.legacy.core)
```
kewisch, line 13: Help me understand this better - I heard we were trying to avoid dependencies or code in legacy and moving it over. If the implementation is in projects.legacy.core, does that mean we're crossing the boundary?

core/android/testing/src/main/kotlin/app/k9mail/core/android/testing/MockHelper.kt
```diff
@@ -1,4 +1,4 @@
-package com.fsck.k9.testing
+package app.k9mail.core.android.testing
```
kewisch, line 1: Would we be renaming this to net.thunderbird in the process?

core/android/testing/build.gradle.kts
shamim-emon: @kewisch yes, our ultimate goal is to get rid of dependencies on legacy code. But sometimes, we still need to depend on legacy module as an  interim solution. Like here  the moving parts(from legacy to corresponding module) have lot of dependencies that would be more than usual to handle in one PR. In such case we add such dependency with the aim to remove it once all corresponding dependencies are substracted(ideally in multiple PRs).
Please have a look at  such existing legacy dependency on main branch [here](https://github.com/thunderbird/thunderbird-android/blob/f881fb8f408d648d7f06ea509924f0bf3fc960b6/app-common/build.gradle.kts#L20-L21)

core/android/testing/src/main/kotlin/app/k9mail/core/android/testing/MockHelper.kt
shamim-emon: @kewisch Sure, will update it.

shamim-emon: @kewisch done!

PR #9067, opened 2025-04-15
Moves :legacy:ui:account to core:ui:account 

### LINKED ISSUE #9066: Moves :legacy:ui:account to core:ui:account
Moves `:legacy:ui:account` to `core:ui:account`

shamim-emon:
- Fixes #9066


kewisch:
@shamim-emon for my own sake, can you explain a bit more on this one? I recall that our account model is very complex and we want to reduce dependencies, so every time I hear account I am a bit wary. 

Did Wolf share any guidelines on moving these modules to core?

shamim-emon:
> @shamim-emon for my own sake, can you explain a bit more on this one? I recall that our account model is very complex and we want to reduce dependencies, so every time I hear account I am a bit wary.
> 
> Did Wolf share any guidelines on moving these modules to core?

@kewisch I haven't had a discussion particularly on ` :legacy:account` but  the idea is to move stuff from legacy module to `core`/`feature `module and some cases `app-common` module. That's the general guideline from discussion with @wmontwe .

kewisch:
Are there any steps we need to take care of when moving things over or simply move code over? Such as renaming/refactoring some of the code contained, removing depdendencies into legacy, etc?  How would I know if app-common or core/feature?

shamim-emon:
> Are there any steps we need to take care of when moving things over or simply move code over? Such as renaming/refactoring some of the code contained, removing depdendencies into legacy, etc? How would I know if app-common or core/feature?

@kewisch yes, there are steps, such as:
1. You may need to create new sub-module if existing sub-modules are not specific enough to contain your changes.
2. We may need to dependencies both in legacy & destination module.

And it varies based  on what you're moving.

Regarding figuring out if changes are to be moved in app-common or core or feature is based on observation. While working on previous  legacy refactoring tasks, some cases I discussed with @wmontwe and other cases I was able to figure-out, based on my code-base investigation. In-case you find anything that seems like a wrong move(eg: whether to move in core or in feature ), please let me know.

kewisch (approved):

PR #9079, opened 2025-04-25
Narrow Search intent usage

asoucar:
Only run search if coming from the search component

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
@@ -407,10 +407,15 @@ open class MessageList :
             }
         }
     }
-
     private fun decodeExtras(intent: Intent): Boolean {
-        val launchData = decodeExtrasToLaunchData(intent)
+        if (intent.action === Intent.ACTION_SEARCH && !intent.component?.className.equals(
+                Search::class.java.name,
+            )
+        ) {
+            finish()
+        }
 
+        val launchData = decodeExtrasToLaunchData(intent)
         // If Unified Inbox was disabled show default account instead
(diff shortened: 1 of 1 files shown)


legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
```diff
@@ -407,10 +407,14 @@ open class MessageList :
...
         }
     }
-
     private fun decodeExtras(intent: Intent): Boolean {
         val launchData = decodeExtrasToLaunchData(intent)
-
+        if (intent.action === Intent.ACTION_SEARCH && !intent.component?.className.equals(
+                "com.fsck.k9.activity.Search",
+            )
+        ) {
```
rafaeltonholo, line 415: I would like to suggest the following change:
```suggestion
        if (intent.action == Intent.ACTION_SEARCH && intent.component?.className != Search::class.java.name) {
```

It will be easier to track whether the name or package of the `Search` Activity changes in the future. Additionally, as we are finishing the Activity if the intent does not originate from the `Search` Activity, I recommend moving the conditional check before the `decodeExtrasToLaunchData(intent)`, as we won’t need that data at that point.

rafaeltonholo (approved):
LGTM, added just a minor suggestion.

asoucar:
[Approval Request Comment]
Original Issue/Pull request: https://github.com/thunderbird/thunderbird-android/security/advisories/GHSA-3x28-g388-xvrf
Regression caused by (issue #): 
User impact if declined: Potential external access to Search
Testing completed (on daily, etc.): manual
Risk to taking this patch (and alternatives if risky): Third party integrations could encounter issues

coreycb:
Uplifted to beta https://github.com/thunderbird/thunderbird-android/commit/391bdd95c1ea1de4fcdea1dafe4fb2d56daaff1d

Issue #8024, opened 2024-07-26
outlook.com outgoing server autoconfig fails


Hanxyz13:
### Checklist

- [X] I have used the search function to see if someone else has already submitted the same bug report.
- [X] I will describe the problem with as much detail as possible.

### App version

6.904

### Where did you get the app from?

F-Droid

### Android version

13

### Device model

POCO X3 GT

### Steps to reproduce

1.Open mailbox
2.Add email address
3.Use OAuth2.0 to login
4.Display authentication error

### Expected behavior

When I use OAuth2.0 and authorise successfully, the mailbox should login successfully

### Actual behavior

After I have successfully authorised using OAuth2.0，it prompts authentication error
<img width="50%" src="https://github.com/user-attachments/assets/23802244-d733-4d4d-9fb1-56518ab61d38">

yephny:
See #8016

Hanxyz13:
> See #8016

Oh，when I try to change the smtp server from smtp.office365.com to smtp-mail.outlook.com according to the [microsoft docs](https://support.microsoft.com/en-us/office/pop-imap-and-smtp-settings-for-outlook-com-d088b986-291d-42b8-9564-9c414e2aa040),it also display Authentication error.![Screenshot_2024-08-01-06-53-11-379_com.fsck.k9.jpg](https://github.com/user-attachments/assets/50c687be-1c80-4f0d-826a-9481f3e0465c)

admbeck:
I am experiencing the same issue on Poco X3 NFC with both 6.804 and 6.904

kewisch:
Can you try again on Thunderbird or K-9 Mail 8.2 or later? We've made some updates to the authentication especially in Thunderbird, so I am hoping this is resolved by now.

Hanxyz13:
> Can you try again on Thunderbird or K-9 Mail 8.2 or later? We've made some updates to the authentication especially in Thunderbird, so I am hoping this is resolved by now.

When I saw this message, I tried to log in to outlook using Thunderbird 9.0, but still got authentication error.
Finally, I logged into outlook using Thunderbird on my computer and backed up my settings and imported them to my phone, then successfully logged into outlook on my phone using Thunderbird.![Screenshot_2025-03-13-07-59-47-291_net.thunderbird.android.jpg](https://github.com/user-attachments/assets/835a07f3-2cb3-40cc-94e5-e754dbdc2d4f)

kewisch:
[...]

PR #9089, opened 2025-04-28
Fixes Outlook outgoing server autoconfig failure

rafaeltonholo:
Fixes: #8024 

The issue mainly occurred when the primary email associated with an account did not match the username used for authenticating to the Outlook SMTP. To resolve this problem, we are now retrieving the primary email included in the OAuth2 tokens to ensure proper authentication with the outgoing server.

Changeset:
- Migrate `OAuth2TokenProvider` to Kotlin
- When available, use the primary email; otherwise, use the given username for authenticating in outgoing servers.

### DIFF
--- a/legacy/common/src/main/java/com/fsck/k9/backends/RealOAuth2TokenProvider.kt
+++ b/legacy/common/src/main/java/com/fsck/k9/backends/RealOAuth2TokenProvider.kt
@@ -17,19 +17,25 @@ import timber.log.Timber
 class RealOAuth2TokenProvider(
     context: Context,
     private val authStateStorage: AuthStateStorage,
-
 ) : OAuth2TokenProvider {
     private val authService = AuthorizationService(context)
     private var requestFreshToken = false
+    private val authState
+        get() = authStateStorage.getAuthorizationState()
+            ?.let { AuthState.jsonDeserialize(it) }
+            ?: throw AuthenticationFailedException("Login required")
+
+    override val primaryEmail: String?
+        get() = authState.parsedIdToken
+            ?.additionalClaims
+            ?.get("email")
+            ?.toString()
 
     @Suppress("TooGenericExceptionCaught")
     override fun getToken(timeoutMillis: Long): String {
         val latch = CountDownLatch(1)
         var token: String? = null
         var exception: AuthorizationException? = null
-
-        val authState = authStateStorage.getAuthorizationState()?.let { AuthState.jsonDeserialize(it) }
-            ?: throw AuthenticationFailedException("Login required")
 
         if (requestFreshToken) {
             authState.needsTokenRefresh = true
--- a/mail/protocols/smtp/src/test/java/com/fsck/k9/mail/transport/smtp/SmtpServerSettingsValidatorTest.kt
+++ b/mail/protocols/smtp/src/test/java/com/fsck/k9/mail/transport/smtp/SmtpServerSettingsValidatorTest.kt
@@ -18,6 +18,7 @@ import com.fsck.k9.mail.testing.security.SimpleTrustedSocketFactory
 import com.fsck.k9.mail.transport.mockServer.MockSmtpServer
 import java.net.UnknownHostException
 import kotlin.test.Test
+import okio.ByteString.Companion.encodeUtf8
 
 private const val USERNAME = "user"
 private const val PASSWORD = "password"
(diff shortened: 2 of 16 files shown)


legacy/common/src/main/java/com/fsck/k9/backends/RealOAuth2TokenProvider.kt
```diff
@@ -17,20 +17,26 @@ import timber.log.Timber
...
+    private val authState
+        get() = authStateStorage.getAuthorizationState()
+            ?.let { AuthState.jsonDeserialize(it) }
+            ?: throw AuthenticationFailedException("Login required")
+
+    override val primaryEmail: String?
+        get() = authState.parsedIdToken
+            ?.additionalClaims
+            ?.get("email")
+            ?.toString()
```
kewisch, line 32: I think it would be useful to have a comment here explaining this is mainly what Microsoft responds with, which I believe is true. It appears to be an optional claim of the JWT token and is documented here: https://learn.microsoft.com/en-us/entra/identity-platform/optional-claims-reference

mail/protocols/smtp/src/test/java/com/fsck/k9/mail/transport/smtp/SmtpServerSettingsValidatorTest.kt
```diff
@@ -334,7 +334,7 @@ class SmtpServerSettingsValidatorTest {
     }
 }
 
-class FakeOAuth2TokenProvider : OAuth2TokenProvider {
+class FakeOAuth2TokenProvider(override val primaryEmail: String? = null) : OAuth2TokenProvider {
```
kewisch, line 337: Would it make sense to have a test that verifies if the optional claim is included that it uses the primary email? Or am I misreading this test and it already does that?

kewisch (approved):
This looks good to me with the few comments considered. We'll want QA to test this with all the remaining OAuth providers. Once merged and in either a daily or beta release, let's get Corey to pass this on to QA.

legacy/common/src/main/java/com/fsck/k9/backends/RealOAuth2TokenProvider.kt
rafaeltonholo: The interface has the documentation explaining it, but it doesn't mention Microsoft directly. I can expand the explanation if you think it worth it

mail/protocols/smtp/src/test/java/com/fsck/k9/mail/transport/smtp/SmtpServerSettingsValidatorTest.kt
rafaeltonholo: I guess you are right. Let me add the test case

legacy/common/src/main/java/com/fsck/k9/backends/RealOAuth2TokenProvider.kt
rafaeltonholo: Done!

mail/protocols/smtp/src/test/java/com/fsck/k9/mail/transport/smtp/SmtpServerSettingsValidatorTest.kt
rafaeltonholo: Added a test case where the primary email is used instead of the given username

kewisch:
> Given your addressing @kewisch 's comments I think the only thing left to do is squash then this is good

I think in this case we could leave them separate. The converting works as is. Wolf is preparing some guidelines in #8964 for this.

kewisch:
[Approval Request Comment]
Original Issue/Pull request: #8024
Regression caused by (issue #):
User impact if declined: Users are not able to log in to Microsoft properties (Hotmail/Outlook) if their account email differs from their mailbox email
Testing completed (on daily, etc.): QA will commence testing as requested on Apr 29th
Risk to taking this patch (and alternatives if risky): If something goes substantially wrong, users will not be able to log in to one of the services in https://github.com/thunderbird/thunderbird-android/blob/main/app-thunderbird/src/beta/kotlin/net/thunderbird/android/auth/TbOAuthConfigurationFactory.kt. Mitigated by having QA test these scenarios.

coreycb:
This was included in today's merge of main->beta

PR #9091, opened 2025-04-30
Import DDD client-adapter from github

breed:
Set up build to download the client-adapter dependency
Configure a DDDClientAdapter in the DDD backend


breed:
dang sorry. we have an extremely experimental fork and this change was not supposed to go outside the fork.

kewisch:
I would love to hear more about what you are up to some time, sounds super interesting!

PR #9102, opened 2025-05-03
Enable actual cherry-pick during --dry-run

coreycb:
When running with --dry-run, create a temporary branch to
perform the cherry-picks. Prior to this, cherry-picks
weren't performed with --dry-run.

Additional fixes:
* Add usage() to specify required and optional flags
* Add --limit 99 to 'gh pr list' as it defaults to 30
* Fix 'gh pr edit' command by adding quotation marks

### DIFF
--- a/scripts/ci/uplift-merges.sh
+++ b/scripts/ci/uplift-merges.sh
@@ -4,12 +4,23 @@ function fail() {
   echo "Error: $*"
   exit 1
 }
+
 function drydo() {
   if [ "$dry_run" = true ]; then
     echo "$@"
   else
     eval "$@"
   fi
+}
+
+function usage() {
+  echo "Usage: $0 [--release | --beta] [--no-dry-run] [--push]"
+  echo
+  echo "  --release      Required: merge into release branch (mutually exclusive with --beta)"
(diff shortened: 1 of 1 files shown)


kewisch:
Would we ever be in a situation where partial cherry picks make sense? I would imagine that the script should be in a way that we can run it once with all the cherry picks that succeed and remove the labels from only those. Then either run it again for a second pass, or manually take care of the remaining ones.

coreycb:
> Would we ever be in a situation where partial cherry picks make sense? I would imagine that the script should be in a way that we can run it once with all the cherry picks that succeed and remove the labels from only those. Then either run it again for a second pass, or manually take care of the remaining ones.

Maybe what we could do is keep the original single loop code and if a cherry-pick fails we could 'git cherry-pick --abort', then continue on with the push.
1) Run script to handle successful cherry-picks, remove corresponding labels, up to the failing cherry-pick, and push
2) Manually deal with the failing cherry-pick, remove label, and push
3) Repeat step 1 if there are more to uplift

kewisch (approved):

dandarnell (approved):

PR #9107, opened 2025-05-05
Select correct child folder in account drawer

### LINKED ISSUE #9078: New Drawer: Incorrect folder is loaded
### Checklist

### App

Thunderbird for Android

### App version

11.0 debug

Other

### Android version

-

### Device model

_No response_

### Steps to reproduce

1. Make sure you are using debug/daily with the dropdown drawwer
2. Collapse some folders
3. Select a folder under the collapsed folder

### Expected behavior

* The correct folder is selected

### Actual behavior

* A random other folder will be selected
* Sometimes there is an error message that the folder could not be found, I haven't been able to re-reproduce this
* When a placeholder folder is selected, the inbox is selected instead

### Logs

_No response_

kewisch:
Fixes #9078 

It looks to me like the wrong `displayFolder` is passed on. If I take the folder from the click handler directly it seems to work for me. I'd appreciate if someone could pick this up and write a test for it. It is only visible on folders a few levels down, for me the structure was

Archives -> 2008 -> 2008-11

Clicking on the 2008 folder loaded the 2008 folder. Clicking on the 2008-11 folder also loaded the 2008 folder. With this patch it selects the right folder.

### DIFF
--- a/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/folder/FolderListItem.kt
+++ b/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/folder/FolderListItem.kt
@@ -90,7 +90,7 @@ internal fun FolderListItem(
                 displayFolder = displayChild,
                 selected = false,
                 showStarredCount = showStarredCount,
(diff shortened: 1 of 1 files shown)


wmontwe (approved):
This resolves the issue.

rafaeltonholo (approved):

PR #9123, opened 2025-05-07
Add blank issue type and use consistent casing

kewisch:
Unfortunately there is no way to enable blank issues just for maintainers. It is cumbersome to do it with one of the hacks that still works after github restricted changing the URL. Here is an attempt to allow us to create targeted issues while encouraging others to use the existing templates.

If it doesn't work out we can also add some automation to auto-close issues that don't use a template for people who are not maintainers.

I've also made the sentences consistent with GitHub's auto-generated ones, the security report one does not have a period at the end, and sentence case for the headlines.


wmontwe (approved):
LGTM

PR #9137, opened 2025-05-08
Fixed Bug - Cannot disable New mail notifcations

### LINKED ISSUE #9135: Cannot disable "New mail notifcations" in 11.0b1
### Checklist

### App

Thunderbird for Android

### App version

11.0b1

Play store

### Android version

15

### Device model

Pixel 9 Pro

### Steps to reproduce

* Go to General settings->Select an account->Notifications
* Remove check mark for "New mail notifications"
* Hit back arrow
* Go to Notifications again
* "New mail notifications" contains check mark

### Expected behavior

"New mail notifications" check mark removal should be persistent

### Actual behavior

"New mail notifications" contains check mark after I removed it

### Logs

_No response_

shamim-emon:
- Fixes #9135

### DIFF
--- a/legacy/core/src/main/java/com/fsck/k9/AccountPreferenceSerializer.kt
+++ b/legacy/core/src/main/java/com/fsck/k9/AccountPreferenceSerializer.kt
@@ -24,9 +24,9 @@ import app.k9mail.legacy.account.QuoteStyle
 import app.k9mail.legacy.account.ShowPictures
 import app.k9mail.legacy.account.SortType
 import com.fsck.k9.helper.Utility
-import com.fsck.k9.preferences.Storage
 import com.fsck.k9.preferences.StorageEditor
 import net.thunderbird.core.mail.folder.api.SpecialFolderSelection
+import net.thunderbird.core.preferences.Storage
 import net.thunderbird.feature.notification.NotificationLight
 import net.thunderbird.feature.notification.NotificationSettings
 import net.thunderbird.feature.notification.NotificationVibration
--- a/app-common/src/main/kotlin/net/thunderbird/app/common/account/CommonAccountDefaultsProvider.kt
+++ b/app-common/src/main/kotlin/net/thunderbird/app/common/account/CommonAccountDefaultsProvider.kt
@@ -27,6 +27,7 @@ import app.k9mail.legacy.account.ShowPictures
 import com.fsck.k9.CoreResourceProvider
 import com.fsck.k9.K9
 import net.thunderbird.core.mail.folder.api.SpecialFolderSelection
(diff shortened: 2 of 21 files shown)


legacy/core/src/main/java/com/fsck/k9/preferences/Storage.java
```diff
@@ -66,4 +66,8 @@ public String getString(String key, String defValue) {
         }
         return val;
     }
+
+    public boolean valueAlreadyExist(String key) {
+        return values.containsKey(key);
+    }
```
rafaeltonholo, line 72: We already have a method called `contains(key)` that serves the same purpose as the newly introduced method. Please use the existing method instead.

legacy/core/src/main/java/com/fsck/k9/AccountPreferenceSerializer.kt
```diff
@@ -61,13 +63,18 @@ class AccountPreferenceSerializer(
...
                 "$accountUuid.folderNotifyNewMailMode",
                 FolderMode.ALL,
             )
-            isNotifySelfNewMail = storage.getBoolean("$accountUuid.notifySelfNewMail", true)
+            if (storage.valueAlreadyExist("$accountUuid.notifyNewMail")) {
+                isNotifyNewMail = storage.getBoolean("$accountUuid.notifyNewMail", false)
+                isNotifySelfNewMail = storage.getBoolean("$accountUuid.notifySelfNewMail", true)
+            } else {
+                accountDefaultsProvider.applyOverwrites(account = account)
+            }
```
rafaeltonholo, line 76: While the `applyOverwrites` method currently overrides both `isNotifyNewMail` and `isNotifySelfNewMail`, it is not intended to be limited to these functions. We need to explore alternative solutions for this issue.

rafaeltonholo (changes requested):
Although this solution addresses the issue, I believe it is not the ideal way to handle it. The `AccountDefaultsProvider.applyOverwrites` method is generic and could potentially include other fields in the future, beyond just `isNotifyNewMail` and `isNotifySelfNewMail`.

I suggest exploring a solution where the `applyOverwrites` method checks whether the user has already modified the field it is attempting to override.

legacy/core/src/main/java/com/fsck/k9/preferences/Storage.java
shamim-emon: Done

legacy/core/src/main/java/com/fsck/k9/AccountPreferenceSerializer.kt
shamim-emon: Added alternate solution

PR #9152, opened 2025-05-09
Change text preference dialog to auto focus

### LINKED ISSUE #9151: Account Settings - Edit Text dialogs should automatically request focus
The dialog for text preferences should focus the editable text and place the cursor at the end of the text. The keyboard should be opened automatically, too.

wmontwe:
Resolves #9151

### DIFF
--- a/core/ui/compose/preference/src/main/kotlin/net/thunderbird/core/ui/compose/preference/ui/components/dialog/PreferenceDialogTextView.kt
+++ b/core/ui/compose/preference/src/main/kotlin/net/thunderbird/core/ui/compose/preference/ui/components/dialog/PreferenceDialogTextView.kt
@@ -4,13 +4,26 @@ import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.height
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
 import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.setValue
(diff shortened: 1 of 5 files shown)


core/ui/compose/preference/src/main/kotlin/net/thunderbird/core/ui/compose/preference/ui/components/dialog/PreferenceDialogTextView.kt
```diff
@@ -20,13 +28,25 @@ internal fun PreferenceDialogTextView(
...
+        mutableStateOf(
+            TextFieldValue(
+                text = preference.value,
+                selection = TextRange(preference.value.length),
+            ),
+        )
+    }
+
+    LaunchedEffect(Unit) {
+        focusRequester.requestFocus()
```
rafaeltonholo, line 42: We may need to introduce a brief delay before requesting focus; otherwise, the keyboard may not open.

rafaeltonholo (commented):
The Account name dialog gains focus, and the cursor is at the end of the text. However, on some devices I tested, the keyboard does not open.

There may be an issue where Compose is setting the focus before the dialog animation finishes. We might need to introduce a slight delay before requesting focus to ensure the dialog animation completes.

core/ui/compose/preference/src/main/kotlin/net/thunderbird/core/ui/compose/preference/ui/components/dialog/PreferenceDialogTextView.kt
wmontwe: Yeah, seems this is still not working 😞 Despite https://issuetracker.google.com/issues/204502668 beeing marked as fixed. 

I tested a bit and 200ms delay seems to work, 300ms breaks the keyboard on my devices. Maybe you could give this another test on your devices.

rafaeltonholo (approved):

PR #9158, opened 2025-05-12
Enable creating Archive folder when swipe action is archive and account doesn't have one.

### LINKED ISSUE #7062: Not swiping when archive action is set

rafaeltonholo:
Resolves #7062.

### DIFF
--- a/mail/protocols/imap/src/test/java/com/fsck/k9/mail/store/imap/RealImapFolderTest.kt
+++ b/mail/protocols/imap/src/test/java/com/fsck/k9/mail/store/imap/RealImapFolderTest.kt
@@ -19,6 +19,7 @@ import com.fsck.k9.mail.Body
 import com.fsck.k9.mail.DefaultBodyFactory
 import com.fsck.k9.mail.FetchProfile
 import com.fsck.k9.mail.Flag
+import com.fsck.k9.mail.FolderType
 import com.fsck.k9.mail.MessageRetrievalListener
 import com.fsck.k9.mail.MessagingException
 import com.fsck.k9.mail.Part
@@ -31,6 +32,7 @@ import java.io.IOException
 import java.nio.file.Files
 import java.util.Date
 import java.util.TimeZone
+import net.thunderbird.protocols.imap.folder.attributeName
 import okio.Buffer
 import org.apache.james.mime4j.util.MimeUtil
 import org.junit.After
@@ -53,6 +55,7 @@ import org.mockito.kotlin.never
 import org.mockito.kotlin.stub
 import org.mockito.kotlin.whenever
 
+@Suppress("MaxLineLength")
 class RealImapFolderTest {
     private val imapStoreConfig = FakeImapStoreConfig()
     private val internalImapStore = object : InternalImapStore {
@@ -1517,6 +1520,31 @@ class RealImapFolderTest {
         val message = folder.getMessage("uid")
 
         assertThat(message.uid).isEqualTo("uid")
+    }
+
+    @Test
+    fun `create() when folderType not in (REGULAR INBOX OUTBOX) and connection has CREATE_SPECIAL_USE capability, should call CREATE command with USE command and return true`() {
+        // Arrange
+        val types = FolderType.entries.filterNot { folderType ->
+            folderType == FolderType.REGULAR ||
+                folderType == FolderType.INBOX ||
+                folderType == FolderType.OUTBOX
+        }
+        imapConnection.stub {
+            on { hasCapability(Capabilities.CREATE_SPECIAL_USE) } doReturn true
+        }
+        val folderName = "New Folder"
+        val remoteFolder = createFolder(folderName)
+
+        for (folderType in types) {
+            // Act
+            remoteFolder.create(folderType = folderType)
+        }
+
+        // Assert
+        for (folderType in types) {
+            assertCommandIssued("CREATE \"$folderName\" (USE (${folderType.attributeName}))")
+        }
     }
 
     @Suppress("SameParameterValue")


feature/messages/build.gradle.kts
rafaeltonholo, line 1: I'm still unsure if this should be `feature/messages` or `feature/mail/messages`

wmontwe: It should be `feature/mail/messages` as it's a subfeature of mail.

rafaeltonholo: Thanks, I have moved it to `feature/mail`

legacy/ui/legacy/src/main/res/values/attrs.xml
```diff
@@ -8,6 +8,7 @@
         <attr name="messageListSwipeToggleReadColor" format="reference|color" />
         <attr name="messageListSwipeToggleStarColor" format="reference|color" />
         <attr name="messageListSwipeArchiveColor" format="reference|color" />
+        <attr name="messageListSwipeArchiveFolderNotSetColor" format="reference|color" />
```
wmontwe: I was working on removing the attributes, as they have been a huge issue when trying to establish the whitelabel theming. Instead we should use the material reference directly.

legacy/core/src/main/java/com/fsck/k9/preferences/KoinModule.kt
```diff
@@ -23,7 +24,8 @@ val preferencesModule = module {
         )
     }
     factory { FolderSettingsProvider(folderRepository = get()) }
-    factory<AccountManager> { get<Preferences>() }
+    factory<LegacyAccountManager> { get<Preferences>() }
+    factory<AccountManager<*>> { get<LegacyAccountManager>() }
```
wmontwe, line 28: Why do you use `*` here? Shouldn't it be `LegacyAccount`?

feature/mail/messages/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
+plugins {
+    id(ThunderbirdPlugins.Library.androidCompose)
+}
+
+android {
+    namespace = "net.thunderbird.feature.mail.messages"
+}
+
+dependencies {
+    implementation(libs.androidx.compose.material3)
```
wmontwe: We want to avoid using the Material 3 dependency. The composables should be taken from the design system. In case they are not present, they need to be added there.

gradle/libs.versions.toml
```diff
@@ -92,6 +92,7 @@ mime4j = "0.8.12"
 minidns = "1.0.5"
 mockito = "5.17.0"
 mockitoKotlin = "5.4.0"
+mockk = "1.14.2"
```
wmontwe: While `mockk` might offer a more refined API compared to `mockito`, I believe we should avoid using mocks altogether due to their cross-platform compatibility issues and performance drawbacks. Mocks can introduce significant delays, particularly during test initialization, which can quickly become overwhelming when an excessive number of tests includes mock implementations. In most cases a `fake` is the better option.

gradle/libs.versions.toml
```diff
@@ -297,6 +299,7 @@ shared-kmp-common-test = [
   "kotlin-test-junit",
   "kotlinx-coroutines-test",
   "turbine",
+  "mockk",
```
wmontwe: Mocking frameworks have been deliberately removed from shared testing dependencies as part of an effort to discourage their use. Only the shared jvm remained due to heavy use.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/domain/usecase/BuildSwipeActionsTest.kt
wmontwe, line 1: This should be in `feature:mail:messages` and not in`feature:messages`.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/domain/usecase/GetAccountFoldersTest.kt
wmontwe, line 1: This should be in `feature:mail:messages` and not in `feature:messages`.

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListItem.kt
```diff
@@ -25,6 +26,8 @@ data class MessageListItem(
     val databaseId: Long,
     val threadRoot: Long,
 ) {
+    val accountWrapper: LegacyAccountWrapper
+        get() = LegacyAccountWrapper.from(account)
```
wmontwe: I changed the `LegacyAccountWrapper` and removed the `from` method in favor of a dedicated mapper, see #9271.

settings.gradle.kts
```diff
@@ -94,6 +94,7 @@ include(
 include(
     ":feature:mail:account:api",
     ":feature:mail:folder:api",
+    ":feature:mail:messages",
```
wmontwe: I prefer a singular form for the module name, it fits better to a `:feature:mail:message:detail`, `:feature:mail:message:list` naming scheme. And sorry that I proposed the wrong name before.

mail/protocols/imap/src/main/java/com/fsck/k9/mail/store/imap/RealImapFolder.kt
```diff
@@ -253,11 +255,20 @@ internal class RealImapFolder(
             this.connection ?: connectionManager.getConnection()
         }
 
+        val hasSpecialUseCapability = connection.hasCapability(Capabilities.CREATE_SPECIAL_USE)
+
         return try {
             val escapedFolderName = ImapUtility.encodeString(encodedName)
 
+            // https://www.rfc-editor.org/rfc/rfc6154.html#section-5.3
```
wmontwe, line 263: It would be good to add a test for this new behavior.

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListSwipeCallback.kt
```diff
@@ -337,7 +388,7 @@ interface MessageListSwipeListener {
 
 private data class SwipeActionConfig(
     private val colorRoles: ColorRoles,
-    val icon: Drawable,
+    val icon: Drawable?,
```
wmontwe, line 399: Why is the drawable now nullable?

legacy/core/src/main/java/com/fsck/k9/preferences/RealGeneralSettingsManager.kt
```diff
@@ -21,6 +21,8 @@ import net.thunderbird.core.preferences.Storage
 import net.thunderbird.core.preferences.SubTheme
 import net.thunderbird.core.preferences.getEnumOrDefault
 
+private const val KEY_SHOULD_SHOW_SETUP_ARCHIVE_FOLDER_DIALOG = "shouldShowSetupArchiveFolderDialog"
```
wmontwe: 👍

wmontwe (commented):
My first bunch of comments, I continue the review later today.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/domain/usecase/BuildSwipeActionsTest.kt
wmontwe: Sorry, I should have written that they belong into `feature:mail:message`

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/domain/usecase/GetAccountFoldersTest.kt
wmontwe: Sorry, I should have written that they belong into `feature:mail:message`

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListSwipeCallback.kt
rafaeltonholo: When a user selects the Swipe Action as "Archive," but their account is set up as POP3 or they choose not to specify an Archive folder (when it cannot be automatically detected), we will only display the message "Change Swipe Gestures" without an accompanying icon. Therefore, the icon can now be set to null.

Screenshot for reference: 
image

settings.gradle.kts
rafaeltonholo: I would prefer `:feature:mail:message:list`, as `detail` resembles more when I'm reading the message, other than the list itself. Thanks for the suggestion

gradle/libs.versions.toml
rafaeltonholo: I'm not entirely sure I got the cross-platform part, but MockK is compatible with KMP for JVM and the common module. This is another reason I've opted to use MockK instead of Mockito, as it will make the future migration of the module to KMP easier.

Regarding the Fakes vs. Mocks, I agree with your perspective. However, libraries like Mockito and MockK serve not only as mock objects but also as verifiers, which is how I'm using them at the moment. Ideally, I'd like to add a dependency only for the `spyk` API, but unfortunately, MockK doesn't split its APIs into different modules.

Implementing the `verify` behavior with Fakes would require significant effort in the Fake implementations, which I don't think would be advantageous. Therefore, I believe using `spyk` is a suitable option.

feature/mail/messages/build.gradle.kts
rafaeltonholo: I guess that is a leftover I forgot to remove. Thanks for pointing this out, going to remove

legacy/core/src/main/java/com/fsck/k9/preferences/KoinModule.kt
rafaeltonholo: I have used the wildcard `*` because there are instances where I need the `AccountManager` to be injected as `AccountManager<BaseAccount>` instead of `AccountManager<LegacyAccount>`. However, since Koin currently does not differentiate between generics, I can use `AccountManager<LegacyAccount>` without issues. The only drawback of this approach is that if Koin changes its API to start differentiating between generic types, it could break the injection if I continue to use `LegacyAccount` instead of `*`.

Let me know your preference.

gradle/libs.versions.toml
wmontwe: I think in case `mockito` vs `mockk` we should migrate to `mockk`. Mockito has some issues with Kotlin and the mockito Kotlin version is infrequently updated.

Maybe I'm wrong but I thought that `mockk` doesn't support native targets, so a `commonTest` doesn't work.

legacy/core/src/main/java/com/fsck/k9/preferences/KoinModule.kt
wmontwe: I was just wondering if koin would inject the right instance then, but yeah I get it now. Fine for me.

gradle/libs.versions.toml
rafaeltonholo: You are right, MockK doesn't support native targets and might not be available when we add the iOS target. Maybe we could use another library instead. 

I see that [Mokkery](https://klibs.io/project/lupuuss/Mokkery) supports all KMP targets, and it has the verify API. [mockative](https://klibs.io/project/mockative/mockative) is another option too.
I could experiment with it if you feel it's worth it

wmontwe: From a first glance I dislike the @Mockable annotations from Mockative.  I liked the mockk api very much and Mokkery seems to look very familiar. I think from both I would favor Mokkery and would give it a try.

core/preferences/src/commonMain/kotlin/net/thunderbird/core/preferences/GeneralSettings.kt
```diff
@@ -19,6 +19,7 @@ data class GeneralSettings(
     val fixedMessageViewTheme: Boolean,
     val isShowUnifiedInbox: Boolean,
     val isShowStarredCount: Boolean,
+    val shouldShowSetupArchiveFolderDialog: Boolean,
```
wmontwe: I think you need to register this new setting also within `GeneralSettingsDescriptions`

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/PreviewLightDarkLandscape.kt
```diff
@@ -0,0 +1,13 @@
...
+import android.content.res.Configuration.UI_MODE_TYPE_NORMAL
+import androidx.compose.ui.tooling.preview.Preview
+
+@Preview(name = "Light", device = "spec:width=673dp,height=841dp,orientation=landscape")
+@Preview(
+    name = "Dark",
+    uiMode = UI_MODE_NIGHT_YES or UI_MODE_TYPE_NORMAL,
+    device = "spec:width=673dp,height=841dp,orientation=landscape",
+)
+annotation class PreviewLightDarkLandscape
```
wmontwe, line 13: I like the new previews

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/PreviewWithThemeLightDark.kt
```diff
@@ -0,0 +1,105 @@
...
+    modifier: Modifier = Modifier,
+    useRow: Boolean = false,
+    useScrim: Boolean = false,
+    scrimAlpha: Float = 0.8f,
+    scrimPadding: PaddingValues = PaddingValues(24.dp),
+    arrangement: Arrangement.HorizontalOrVertical = Arrangement.spacedBy(24.dp),
+    content: @Composable () -> Unit,
+) {
+    val movableContent = remember {
+        movableContentOf {
```
wmontwe, line 43: Neat, didn't know that existed.

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/PreviewWithThemes.kt
```diff
@@ -99,7 +99,7 @@ private fun PreviewWithThunderbirdTheme(
 }
 
 @Composable
-private fun PreviewHeader(
+fun PreviewHeader(
```
wmontwe: I think this could be internal

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/PreviewWithThemes.kt
```diff
@@ -117,7 +117,7 @@ private fun PreviewHeader(
 }
 
 @Composable
-private fun PreviewSurface(
+fun PreviewSurface(
```
wmontwe: I think this could be internal

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/PreviewWithThemeLightDark.kt
```diff
@@ -0,0 +1,105 @@
...
+ * @param arrangement The arrangement for the previews. Defaults to `Arrangement.spacedBy(24.dp)`.
+ * @param content The content to be displayed in the previews.
+ */
+@Composable
+fun PreviewWithThemeLightDark(
+    modifier: Modifier = Modifier,
+    useRow: Boolean = false,
+    useScrim: Boolean = false,
+    scrimAlpha: Float = 0.8f,
+    scrimPadding: PaddingValues = PaddingValues(24.dp),
```
wmontwe: The theme defines spacings already, this should be replaced by `MainTheme.spacings.triple`

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/PreviewWithThemeLightDark.kt
```diff
@@ -0,0 +1,105 @@
...
+    arrangement: Arrangement.HorizontalOrVertical = Arrangement.spacedBy(24.dp),
```
wmontwe: `MainTheme.spacings.triple`

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/organism/BasicDialogPreview.kt
```diff
@@ -0,0 +1,147 @@
...
+import app.k9mail.core.ui.compose.designsystem.atom.text.TextHeadlineSmall
+import app.k9mail.core.ui.compose.theme2.MainTheme
+
+@PreviewLightDarkLandscape
+@Composable
+private fun BasicDialogPreview() {
+    PreviewWithThemeLightDark(
+        useRow = true,
+        useScrim = true,
+        scrimPadding = PaddingValues(32.dp),
```
wmontwe: The MainTheme should define the spacings, here `MainTheme.spacings.quadruple`

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/organism/BasicDialogPreview.kt
```diff
@@ -0,0 +1,147 @@
...
+        arrangement = Arrangement.spacedBy(24.dp),
```
wmontwe: `MainTheme.spacings.triple`

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/organism/BasicDialogPreview.kt
```diff
@@ -0,0 +1,147 @@
...
+                            end = MainTheme.spacings.triple,
+                        ),
+                ) {
+                    Row(
+                        verticalAlignment = Alignment.CenterVertically,
+                        horizontalArrangement = Arrangement.spacedBy(MainTheme.spacings.double),
+                    ) {
+                        Box(
+                            modifier = Modifier
+                                .size(48.dp)
```
wmontwe: The MainTheme defines sizes, this could be `MainTheme.sizes.iconAvatar`. The theme and design system will be updated soon to apply better naming.

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/organism/BasicDialogPreview.kt
```diff
@@ -0,0 +1,147 @@
...
+                        )
+                        Text(text = "Account 1")
+                    }
+                    Row(
+                        verticalAlignment = Alignment.CenterVertically,
+                        horizontalArrangement = Arrangement.spacedBy(MainTheme.spacings.double),
+                    ) {
+                        Box(
+                            modifier = Modifier
+                                .size(48.dp)
```
wmontwe: This could be `MainTheme.sizes.iconAvatar`

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/organism/BasicDialogPreview.kt
```diff
@@ -0,0 +1,147 @@
...
+                        )
+                        Text(text = "Account 2")
+                    }
+                    Row(
+                        verticalAlignment = Alignment.CenterVertically,
+                        horizontalArrangement = Arrangement.spacedBy(MainTheme.spacings.double),
+                    ) {
+                        Box(
+                            modifier = Modifier
+                                .size(48.dp)
```
wmontwe: This could be `MainTheme.sizes.iconAvatar`

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/organism/BasicDialogPreview.kt
```diff
@@ -0,0 +1,147 @@
...
+}
+
+@PreviewLightDarkLandscape
+@Composable
+private fun PreviewOnlySupportingText() {
+    PreviewWithThemeLightDark(
+        useRow = true,
+        useScrim = true,
+        scrimPadding = PaddingValues(32.dp),
+        arrangement = Arrangement.spacedBy(24.dp),
```
wmontwe: Should be MainTheme.spacings

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/organism/BasicDialogPreview.kt
```diff
@@ -0,0 +1,147 @@
...
+            buttons = {
+                TextButton(onClick = {}) {
+                    Text(text = "Skip for now")
+                }
+                TextButton(onClick = {}) {
+                    Text(text = "Set archive folder")
+                }
+            },
+            showDividers = false,
+            modifier = Modifier.width(300.dp),
```
wmontwe, line 144: This could be `MainTheme.sizes.huge` (256dp) or `MainTheme.sizes.huger` (384dp), but there is no fitting size yet.

core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/organism/BasicDialog.kt
wmontwe, line 1: 👍

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/fakes/FakeSpecialFolderUpdaterFactory.kt
wmontwe, line 1: I think this needs to be moved to the new module.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/fakes/FakeSpecialFolderUpdaterFactory.kt
```diff
@@ -0,0 +1,23 @@
+package net.thunderbird.feature.messages.fakes
+
+import io.mockk.spyk
+import net.thunderbird.feature.mail.account.api.BaseAccount
+import net.thunderbird.feature.mail.folder.api.FolderType
+import net.thunderbird.feature.mail.folder.api.SpecialFolderSelection
+import net.thunderbird.feature.mail.folder.api.SpecialFolderUpdater
+
+class FakeSpecialFolderUpdaterFactory : SpecialFolderUpdater.Factory<BaseAccount> {
+    val specialFolderUpdater = spyk(
```
wmontwe: I now understand how you like to use `spyk`, need to try that too 👍

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/fakes/FakeBackendStorageFactory.kt
wmontwe, line 1: I think this needs to be moved to the new module.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/fakes/FakeBackendFolderUpdater.kt
wmontwe, line 1: I think this needs to be moved to the new module.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/fakes/FakeAccountManager.kt
wmontwe, line 1: I think this needs to be moved to the new module.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/fakes/FakeAccount.kt
wmontwe, line 1: I think this needs to be moved to the new module.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/domain/usecase/SetArchiveFolderTest.kt
wmontwe, line 1: I think this needs to be moved to the new module.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/domain/usecase/GetAccountFoldersTest.kt
wmontwe, line 1: I think this needs to be moved to the new module.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/domain/usecase/CreateArchiveFolderTest.kt
wmontwe, line 1: I think this needs to be moved to the new module.

feature/messages/src/test/kotlin/net/thunderbird/feature/messages/domain/usecase/BuildSwipeActionsTest.kt
wmontwe, line 1: I think this needs to be moved to the new module.

feature/mail/messages/src/main/kotlin/net/thunderbird/feature/mail/messages/ui/dialog/SetupArchiveFolderDialogContract.kt
```diff
@@ -0,0 +1,45 @@
+package net.thunderbird.feature.mail.messages.ui.dialog
+
+import androidx.compose.runtime.Stable
+import app.k9mail.core.ui.compose.common.mvi.UnidirectionalViewModel
+import net.thunderbird.feature.mail.folder.api.RemoteFolder
+
+internal interface SetupArchiveFolderDialogContract {
+    interface ViewModel : UnidirectionalViewModel<State, Event, Effect>
+
+    sealed interface State {
```
wmontwe: I didn't use a sealed interface for state, as I tried to model the whole state as one data class. But I see the appeal of this approach.

feature/mail/messages/src/main/kotlin/net/thunderbird/feature/mail/messages/ui/dialog/SetupArchiveFolderDialog.kt
```diff
@@ -0,0 +1,209 @@
...
+import net.thunderbird.feature.mail.messages.ui.dialog.SetupArchiveFolderDialogContract.ViewModel
+import org.koin.androidx.compose.koinViewModel
+import org.koin.core.parameter.parametersOf
+
+@Composable
+internal fun SetupArchiveFolderDialog(
+    accountUuid: String,
+    onDismissDialog: () -> Unit,
+    modifier: Modifier = Modifier,
+    viewModel: ViewModel = koinViewModel<SetupArchiveFolderDialogViewModel> { parametersOf(accountUuid) },
```
wmontwe: Do you maybe know how to use the interface here with koin and avoid a direct reference?

feature/mail/messages/src/main/kotlin/net/thunderbird/feature/mail/messages/domain/usecase/GetAccountFolders.kt
```diff
@@ -0,0 +1,30 @@
...
+import kotlinx.coroutines.withContext
+import net.thunderbird.core.outcome.Outcome
+import net.thunderbird.feature.mail.folder.api.FolderType
+import net.thunderbird.feature.mail.folder.api.RemoteFolder
+import net.thunderbird.feature.mail.messages.domain.AccountFolderError
+import net.thunderbird.feature.mail.messages.domain.DomainContract
+
+class GetAccountFolders(
+    private val folderRepository: FolderRepository,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
```
wmontwe: I should use a dedicated dispatcher too. 👍

feature/mail/messages/src/main/kotlin/net/thunderbird/feature/mail/messages/FeatureMessageModule.kt
```diff
@@ -0,0 +1,57 @@
...
+import net.thunderbird.feature.mail.messages.domain.usecase.GetAccountFolders
+import net.thunderbird.feature.mail.messages.domain.usecase.SetArchiveFolder
+import net.thunderbird.feature.mail.messages.ui.dialog.SetupArchiveFolderDialogFragment
+import net.thunderbird.feature.mail.messages.ui.dialog.SetupArchiveFolderDialogFragmentFactory
+import net.thunderbird.feature.mail.messages.ui.dialog.SetupArchiveFolderDialogViewModel
+import org.koin.core.module.dsl.viewModel
+import org.koin.core.qualifier.named
+import org.koin.dsl.module
+
+val featureMessageModule = module {
```
wmontwe: You should add a test to verify that the module is setup properly.

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/organism/BasicDialogPreview.kt
wmontwe, line 1: This should be added to the catalog app too as a real world preview of the dialog.

wmontwe (commented):
Looks very good, I left a couple of comments for your consideration.

I like the separation of the UI components and the overall structure and refactoring efforts. Nice trick with the spyk fakes.

core/ui/compose/designsystem/src/debug/kotlin/app/k9mail/core/ui/compose/designsystem/organism/BasicDialogPreview.kt
rafaeltonholo: Since this is just a preview, I would rather not create a new `size` property for this case.

PR #9161, opened 2025-05-13
fix(drawer): openFolder must be called with accountId and folderId

wmontwe:
The MessageList maintains it's own state, but in case an account is deleted it still references the deleted account and when trying to open a folder without accountId, it will use the old account that doesn't exist anymore.


wmontwe:
[Approval Request Comment]
Original Issue/Pull request: #8548
Regression caused by (issue #): N/A
User impact if declined: Crashes when deleting accounts and continuing to use the app.
Testing completed (on daily, etc.): debug
Risk to taking this patch (and alternatives if risky): It's not fixing all possible sources of the original crash, but lowers the likelihood of crashes related to account deletions.

kewisch (approved):

coreycb:
[Triage Comment]
Approved for beta
Approved for release

PR #9169, opened 2025-05-15
chore(ci): Make dependabot use semantic commits and correct label

### LINKED ISSUE #9168: Enforcing semantic commit messages
We recently agreed in the git commit guide we'd use semantic commits. I'd like us to enforce this automatically as much as possible. Proposed actions:

* [ ] Adjust dependabot config to use `chore(deps):` as a prefix
* [x] Change merge setting "Default commit message" to either "Pull request title" or "Pull request title and description", and add a GitHub action to enforce the pull request title needs to be semantic as well.
* [ ] Add a GitHub action and local checks that enforces that each commit uses semantic commits (e.g. commitlint)

@wmontwe @toddheasley does this make sense? Should we add the same checks for iOS as well?

kewisch:
Fixes #9168


coreycb (commented):
Looks good except for the label spelling

kewisch:
🤦 Thanks for catching that.

kewisch:
Thanks for checking. I was intending to add some more content to this, e.g. using a linter to enforce it also for contributors. But I think we can start with this and add the rest later.

coreycb (approved):

PR #9171, opened 2025-05-15
Add (or rather declare) support for RFC 6531, the SMTPUTF8 extension.

arnt:
This gives Thunderbird the ability to send messages to/from unicode addresses such as grå@grå.org. The user interface is not changed by this; it will be changed in a subsequent PR.

Thunderbird already supported the addresses substantively, since it uses the String type for addresses. That class has no support for ASCII-only strings, only for unicode strings. In effect, the programming language prevented Thunderbird from being incompatible with unicode addresses.

RFC 6531 requires more than substantive compatibility though, it also requires correct declaration of the addresses in the SMTP protocol, and that's what this commit adds.


arnt:
Pushed again; the cyclomatic complexity of sentMessage() increased from 15 to 16. I feel that I should perhaps get it down to 10 or so, but I actually left it at 15.

wmontwe (approved):
Thank you!

wmontwe:
@arnt I found the issue with the test and fixed the pr. This is ready to merge.

PR #9172, opened 2025-05-15
Bug Fix: Sharing multiple images from galary, only selects last image

### LINKED ISSUE #9170: Sharing Image Attachments via email via Android Gallery now broken
### Checklist

### App

K-9 Mail

### App version

10.0

Other

### Android version

Android 13

### Device model

Samsung A03

### Steps to reproduce

1. Go to Gallery on Android
2. Select multiple images
3. Press share icon
4. Select K-9 email
5. New mail window now opens with only one image attached, instead of all the images selected.

It used to work in v9.*. Stopped after upgrading to v10. Got apk from github.

### Expected behavior

Step 5 above should be:  New mail window now opens with all images enclosed as attachments.

### Actual behavior

New mail window now opens with only one image attached, instead of all the images selected.

### Logs

_No response_

shamim-emon:
- Fixes #9170

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
@@ -4,6 +4,7 @@ package com.fsck.k9.activity;
 import java.io.File;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


wmontwe (approved):
LGTM

kewisch:
I'm getting reports this happens on 10.0 as well, once tested on beta we should uplift to release.

Issue #9122, opened 2025-05-06
Couldn't load the changelog after 9.0 to 10.0 update


AJCxZ0:
### Checklist

- [x] I have used the search function to see if someone else has already submitted the same bug report.
- [x] I will describe the problem with as much detail as possible.

### App

Thunderbird for Android

### App version

10.0

### Where did you get the app from?

Other

### Android version

15

### Device model

Pixel 7

### Steps to reproduce

1. Start Obtanium or similar package manager sourcing APKs from Github
2. Start the process to update Thunderbird app from 9.0 to 10.0
3. Start Thunderbird app
4. See notice in the footer about changes
5. Click the link in the footer to view the changes

### Expected behavior

The change log is presented.

### Actual behavior

An error is presented: `Couldn't load the changelog.`

### Logs

![Image](https://github.com/user-attachments/assets/8946c264-1416-44ec-82b6-76d2f3af0959)

jammekod:
I had the same issue

kewisch:
Did you install via Play Store, f-droid, or github releases?

AJCxZ0:
> Did you install via Play Store, f-droid, or github releases?

Github release.

I would have specified both this and that I used Obtanium, but the template offered no such option, so I added the details to **Steps to reproduce**.

> Also can you confirm this doesn't happen for 8.0 to 9.0 ?

I cannot. Version 9.0 may have been the initial install.

jammekod:
Yeah, I had the same issue as @AJCxZ0 I couldn't update through Obtanium because it only shows K9. So I manually installed the apk through Github.

PR #9175, opened 2025-05-16
fix: Correctly escape recent changelog entries on release branch

kewisch:
Related to #9122


wmontwe (approved):
LGTM

coreycb (approved):

wmontwe:
Fixed the badging information

wmontwe:
The build is failing because of an old issue with Spotless, that is fixed on main and beta already. 

I've run the build locally and it succeeds. Either we merge this as is or we need to backport the Spotless fix first.

kewisch:
Is the spotless failure uncovered through these changes in some way? I think the only think we need to avoid is that this spotless failure is going to make the release builds fail. If that isn't the case (e.g. it worked for the last release build which should have the same spotless errors), then we could merge.

kewisch:
If there is a chance it might succeed, I guess we can always backport the spotless fix if the release fails.

PR #9176, opened 2025-05-16
fix: Correctly escape recent changelog entries on beta

kewisch:
Related to #9122, this fixes the changelog on beta


wmontwe (approved):
👍

PR #9179, opened 2025-05-16
Add kotlin multiplatform support

### LINKED ISSUE #9178: Add Kotlin Multiplatform support
We'd like to discover which code can be shared across both iOS and Android platforms, so setting up Kotlin Multiplatform is a prerequisite for our project.

wmontwe:
Resolves #9178 

This provides a first Kotlin Multiplatform project setup and converts the modules under `core` to Kotlin Multiplatform. Modules under `core:android` and `core:ui` have been skipped due to complexity of the change.

As a side effect the namespace has been changed to `net.thunderbird`.


kewisch:
Would you like me to take a look at this, or wait until Rafael is back?

core/android/contact/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ plugins {
 }
 
 android {
-    namespace = "net.thunderbird.core.contact"
+    namespace = "net.thunderbird.core.android.contact"
```
kewisch, line 6: So this means this is an android-specific module example? I think we need to have some documentation on when to use `android` in the package name, and at what level. Otherwise I guess we'll have a mix of some namespaces being with `android` and some being without.

kewisch (approved):
This looks good to me from a high level perspective. As noted I think we should add some documentation on our KMP approach so people getting started with namespacing know what to expect.

Final review should be up to @rafaeltonholo

core/android/contact/build.gradle.kts
wmontwe: Yes, it contains only android specific code, hence the package name. I'm preparing a documentation for the overall structure.

core/common/build.gradle.kts
rafaeltonholo, line 1: Would moving the files of `:core:android:common` to `:core:common:androidMain` be part of another task?

core/testing/src/commonMain/kotlin/assertk/assertions/ListExtensions.kt
rafaeltonholo, line 1: Shouldn't this be moved to `commonTest` rather than `commonMain`?

core/testing/src/commonMain/kotlin/assertk/assertions/TurbineExtensions.kt
rafaeltonholo, line 1: Shouldn't this be moved to `commonTest` rather than `commonMain`?

feature/mail/folder/api/build.gradle.kts
rafaeltonholo, line 1: nice!

gradle/libs.versions.toml
```diff
@@ -64,6 +64,7 @@ icu4j = "72.1"
 javaDiffUtils = "4.12"
 jcipAnnotations = "1.0"
 jetbrainsAnnotations = "26.0.2"
+jetbrainsCompose = "1.8.0"
```
rafaeltonholo, line 67: The `1.8.1` was released last week, I would consider upgrading it too

rafaeltonholo:
We also need to create a root `detektAll` task. The `./gradlew detekt` command runs it without type resolution, keeping KMP projects out of the execution and running only the `android` ones. See [Enabling on a KMP project](https://detekt.dev/docs/gettingstarted/type-resolution/#enabling-on-a-kmp-project) and https://github.com/detekt/detekt/issues/3663

core/common/build.gradle.kts
wmontwe: Yes, that's another task. We need to identify common api to move some of the Android specific things into the `:core:common` KMP module.

gradle/libs.versions.toml
wmontwe: ^^ this slipped into here, I have another branch where I introduce Compose Multiplatform

Will update.

wmontwe:
> We also need to create a root `detektAll` task. The `./gradlew detekt` command runs it without type resolution, keeping KMP projects out of the execution and running only the `android` ones. See [Enabling on a KMP project](https://detekt.dev/docs/gettingstarted/type-resolution/#enabling-on-a-kmp-project) and [detekt/detekt#3663](https://github.com/detekt/detekt/issues/3663)

Thanks, I'll think this could be a separate task. I'll check Spotless, too.

core/testing/src/commonMain/kotlin/assertk/assertions/ListExtensions.kt
rafaeltonholo: Oh, I got it. Makes sense

feature/mail/folder/api/src/commonMain/kotlin/net/thunderbird/feature/mail/folder/api/SpecialFolderUpdater.kt
```diff
@@ -1,7 +1,8 @@
-package net.thunderbird.core.mail.folder.api
+package net.thunderbird.feature.mail.folder.api
 
-import net.thunderbird.core.account.BaseAccount
+import net.thunderbird.feature.mail.account.api.BaseAccount
 
+// TODO move to ???
```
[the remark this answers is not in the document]
rafaeltonholo: I believe it would be more appropriate to leave it in `feature:mail:folder:api` rather than moving to `feature:mail:storage:api`, as it is more closely related to folder functionality. While we do need to perform some storage operations, via `Preferences` and `FolderRepository`, the primary focus of the operation is the folder itself. 

In my view, code that belongs to a `:storage` module should deal directly with data operations, which does not seem to be the case for this particular interface, unless I'm missing something.

I initially suggested `:core:mail:mailserver` because, for some odd reason, I misread it as `com.fsck.k9.mailserver` instead of `com.fsck.k9.mailstore`, so you can disconsider that, lol

PR #9187, opened 2025-05-18
Fix: Correct attachment naming for duplicate files

### LINKED ISSUE #9141: Incorrect numbering when saving duplicate attachments for application/octet-stream attachments
### Checklist

### App

K-9 Mail

### App version

9.0

F-Droid

### Android version

Android 15, CalyxOS 6.6.23

### Device model

Fairphone 4 (FP4)

### Steps to reproduce

1. Receive a mail with an attachment.
2. Save that and do it again with another mail with the same attachment (I guess it also works for the same mail).

### Expected behavior

Let's say the attachment is named `file.pdf`. Given a file named `file.pdf` already exists in the Downloads folder, the file should be named  `file (2).pdf` or similar.

### Actual behavior

The file is named  `file.pdf (1)`. This is not (actually) (very) funny, but causes problems, because many applications including the built-in file picker of Android do not detect that as a PDF file anymore in that example.
So this breaks a lot of functionality in a subtle way and you have to look twice and have at least basic knowledge about files (notably many people may not have this anymore) to fix it.
Thing is it e.g. may also break images and people could wonder why they don't see saved images in their gallery etc.

Workaround:  renaming, of course

Another less erious issue: Notice how it is suffixed with `(1)`? While this may make sense for technical people, it makes less for everyone else.
I checked how Nautilus (GNOME file manager) does it e.g. when I copy a file in the same dir and it names them `file (copy).pdf` and even `file (second copy).pdf` for the second.

And if I copy a file that already exists in the target location (which is likely more similar to the issue at hand), I can rename it and the default suggestion is: `file (2).pdf`:
Image

IMHO both versions are a better UX.

### Logs

[k9mail-logs_duplicatefile.txt](https://github.com/user-attachments/files/20110658/k9mail-logs_duplicatefile.txt)

karthikb1612:
Fixes #9141

- Follows our existing codestyle (`gradlew spotlessCheck` to check and `gradlew spotlessApply` to format your source code; will be checked by CI).

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/AttachmentController.java
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/AttachmentController.java
@@ -1,23 +1,18 @@
 package com.fsck.k9.ui.messageview;
 
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-
 import android.content.ActivityNotFoundException;
 import android.content.ContentResolver;
 import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
-import android.os.AsyncTask;
+import android.os.Handler;
+import android.os.Looper;
 import android.widget.Toast;
 
 import androidx.annotation.WorkerThread;
-import app.k9mail.legacy.account.LegacyAccount;
+
 import com.fsck.k9.Preferences;
 import com.fsck.k9.controller.MessagingController;
-import app.k9mail.legacy.message.controller.SimpleMessagingListener;
 import com.fsck.k9.mail.Message;
 import com.fsck.k9.mail.Part;
 import com.fsck.k9.mailstore.AttachmentViewInfo;
@@ -25,9 +20,18 @@ import com.fsck.k9.mailstore.LocalMessage;
 import com.fsck.k9.mailstore.LocalPart;
 import com.fsck.k9.provider.AttachmentTempFileProvider;
 import com.fsck.k9.ui.R;
-import org.apache.commons.io.IOUtils;
-import timber.log.Timber;
 
+import org.apache.commons.io.FilenameUtils;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+
+import app.k9mail.legacy.account.LegacyAccount;
+import app.k9mail.legacy.message.controller.SimpleMessagingListener;
+import timber.log.Timber;
 
 public class AttachmentController {
     private final Context context;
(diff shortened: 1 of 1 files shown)


kewisch:
Thanks for taking a stab at this, a reviewer will take care shortly! You are welcome to wait until they comment before you make changes, but would you be willing to go the extra mile and convert that file from java to Kotlin? We'd much appreciate, as this allows us to eventually get rid of our Java dependencies. Ideally if you could split the convert to Kotlin commit from this fix it will make future commit archaeology a bit easier to parse.

karthikb1612:
Thanks for the feedback! I’d be happy to convert the file to Kotlin and split it into a separate commit. However, I currently have semester exams, so I’ll need a bit of time to make these changes. I’ll get back to it as soon as possible—thanks for your patience!

kewisch:
That works for me, good luck with your exams! 🤞 I've tested this PR with the email I can reproduce this on but still it appears to be adding to the end:
image

karthikb1612:
Thank you for testing and for your kind words! I really appreciate it. I'm glad the fix is working for you, though I see that in your test the filename is still getting the number added at the end.

Just to clarify: this PR specifically addresses the issue of how duplicate filenames like file.pdf (1) are handled — converting them to file (1).pdf instead. If you're still seeing the old behavior, could you let me know which steps or email you're using to reproduce it? I'll double-check and test again after my exams to make sure it's fully resolved.

Thanks again for your support!

kewisch:
Apologies I might have worded that unclearly. I did mean to say it isn't fixed for me and that I have an email I can reproduce this on. I'd rather not share the email I received since it contains some personal info, but I can share the email structure. @rugk might be willing to share his REWE receipt if you can share where to send it.

Here is the body with the PDF attachment:

```
MIME-Version: 1.0
Content-Type: multipart/mixed;
 boundary="b1=_4aSSw3poeqeh2SIeU0gIL1yanzjX0Oi5dTlFrwbYiY"

--b1=_4aSSw3poeqeh2SIeU0gIL1yanzjX0Oi5dTlFrwbYiY
Content-Type: text/html; charset=utf-8

<there is html content here that is being displayed>

--b1=_4aSSw3poeqeh2SIeU0gIL1yanzjX0Oi5dTlFrwbYiY
Content-Type: application/octet-stream; name=REWE-eBon.pdf
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename=REWE-eBon.pdf 

<there is a PDF with a receipt here>

--b1=_4aSSw3poeqeh2SIeU0gIL1yanzjX0Oi5dTlFrwbYiY--
```

As you can see it says `Content-Type: application/octet-stream`. If I manually re-send that email using SMTP via command line and change it to `application/pdf`, then it works for me.

karthikb1612:
Thanks for the clarification! That makes sense now. I understand the issue is still present when the Content-Type is set to application/octet-stream instead of application/pdf, even though the file name ends with .pdf.

I’ll look into updating the logic to better handle this case — possibly by checking the filename extension if the content type is too generic. I’ll follow up with a revised commit once I’ve tested it.

Thanks again for sharing the structure and helping me reproduce it!

kewisch:
Thanks! Using the file extension might be risky, though I am not sure. Maybe an attacker could disguise a binary file by using the wrong file extension, while the system might interpret the file based on the magic number or other indicator. There might be a few comments around this hidden in our issue tracker, we've generally been trying to push for adhering to standards rather than trying to work around them and asking people to fix their providers.

kewisch:
@karthikb1612 I had marked this as draft since you wanted to make some further changes. Did I miss anything and is this ready for review?

karthikb1612:
Thanks for checking in! I had marked it as draft because I needed some more time due to my semester exams. I’ll get back to it and mark it as ready for review once I complete the pending changes. Appreciate your patience!

kewisch:
Hey @karthikb1612, hope you are doing well. I know you've been busy and you are welcome to take as much time as you like before you get back to this. For the sake of reducing the list of pull requests I'm going to close this one now. When you are ready to update the PR please just go ahead and open a new one. Looking forward to your contribution!

PR #9207, opened 2025-05-22
feat: add support for single select from dropdown for preference system

### LINKED ISSUE #9195: Add Support For  Single Select From a Dropdown
- Sub-task of #9164 
 -  Add Support For Single Select From a Dropdown for he new Preference System.

shamim-emon:
- Fixes #9195


core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/atom/RadioGroup.kt
```diff
@@ -0,0 +1,40 @@
...
+import androidx.compose.foundation.layout.Row
+import androidx.compose.material3.RadioButton
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import app.k9mail.core.ui.compose.designsystem.atom.text.TextLabelLarge
+import kotlinx.collections.immutable.ImmutableList
+
+@Composable
+fun <T> RadioGroup(
```
wmontwe, line 12: There should be a preview to investigate the RadioGroup in the design system. 

Also the component should be added to the Catalog app in the atom section.

core/ui/compose/preference/src/debug/kotlin/net/thunderbird/core/ui/compose/preference/ui/components/list/PreferenceItemButtonSegmentedSingleChoiceViewPreview.kt
```diff
@@ -7,9 +7,9 @@ import net.thunderbird.core.ui.compose.preference.ui.fake.FakePreferenceData
 
 @Composable
 @Preview(showBackground = true)
-internal fun PreferenceItemSingleChoiceViewPreview() {
+internal fun PreferenceItemButtonSegmentedSingleChoiceViewPreview() {
```
wmontwe, line 10: Why the rename?

wmontwe (changes requested):
The `Preference` should contain all types of preferences that could be used to create a preference screen. They are meant to be a 1to1 match to ui elements.

In this context, the user could select from a list of options and is only allowed to select one of all. The ui for this preference is a list item that shows the title, an optional icon and the current selected value. As rendering the full list of options is not desired for the resulting settings page, a dialog should be openend when the user clicks on the item. The dialog shows the title, optional icon and description along with the list of options.

This sets it appart from the SingleChoice preference which is only used for in place selection of 2-3 options or categories, so only an item view is needed.

So we need a new PreferenceSetting, used for preferences that are editable vs static elements as part of PreferenceDisplay.

Then there needs to be a corresponding item and dialog view to render the new PreferenceSetting.

core/ui/compose/preference/src/debug/kotlin/net/thunderbird/core/ui/compose/preference/ui/components/list/PreferenceItemButtonSegmentedSingleChoiceViewPreview.kt
shamim-emon: Reverted!

core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/atom/RadioGroup.kt
```diff
@@ -0,0 +1,40 @@
...
+                    onClick(option)
+                },
+                selected = option == selectedOption,
+            )
+
+            TextLabelLarge(
+                text = optionTitle(option),
+            )
+        }
+    }
```
wmontwe: I think this should be wrapped by a `Column` and the modifier should be applied to it. At the moment the modifier applies padding to each `Row` which results in a lot of spacing.

core/ui/compose/preference/src/main/kotlin/net/thunderbird/core/ui/compose/preference/api/Preference.kt
```diff
@@ -67,6 +68,25 @@ sealed interface PreferenceSetting<T> : Preference {
...
+    data class SingleChoiceCompact(
+        override val id: String,
+        val title: () -> String,
+        val description: () -> String? = { null },
+        val icon: () -> ImageVector? = { null },
+        override val value: CompactChoice,
+        val options: ImmutableList<CompactChoice>,
+    ) : PreferenceSetting<CompactChoice> {
+        @IgnoredOnParcel
+        override val requiresEditView: Boolean = false
```
wmontwe: This should be true to signal that there is an edit view. Not used at the moment, but I'll update the implementation to check the flag.

core/ui/compose/preference/src/main/kotlin/net/thunderbird/core/ui/compose/preference/ui/components/list/PreferenceItem.kt
```diff
@@ -47,6 +47,8 @@ internal fun PreferenceItem(
             )
         }
 
+        is PreferenceSetting.SingleChoiceCompact -> Unit
```
wmontwe: This should still render the list item similarly to the `PreferenceItemColorView` just rendering the current selected value.

wmontwe (approved):
Thank you!

PR #9213, opened 2025-05-24
feat: remove gmail prefix in folder structure

### LINKED ISSUE #9134: Gmail folder naming contains [] and / in prefix
### Checklist

### App

Thunderbird for Android

### App version

10

Other

### Android version

15

### Device model

Pixel 9

### Steps to reproduce

1. Go to the app (must have a Gmail account)
2. Click on the folder

Image

### Expected behavior

The folder name shouldn't have [Gmail[\

### Actual behavior

I attached a screenshot

### Logs

_No response_

shamim-emon:
- Fixes #9134

### DIFF
--- a/legacy/core/src/test/java/com/fsck/k9/mailstore/K9BackendFolderTest.kt
+++ b/legacy/core/src/test/java/com/fsck/k9/mailstore/K9BackendFolderTest.kt
@@ -16,10 +16,13 @@ import com.fsck.k9.backend.api.BackendFolder
 import com.fsck.k9.backend.api.FolderInfo
 import com.fsck.k9.backend.api.updateFolders
 import com.fsck.k9.mail.Address
+import com.fsck.k9.mail.AuthType
+import com.fsck.k9.mail.ConnectionSecurity
 import com.fsck.k9.mail.Flag
 import com.fsck.k9.mail.FolderType
 import com.fsck.k9.mail.Message
 import com.fsck.k9.mail.MessageDownloadState
+import com.fsck.k9.mail.ServerSettings
 import com.fsck.k9.mail.internet.MimeMessage
 import com.fsck.k9.mail.internet.MimeMessageHelper
 import com.fsck.k9.mail.internet.TextBody
@@ -97,8 +100,10 @@ class K9BackendFolderTest : K9RobolectricTest() {
     fun createAccount(): LegacyAccount {
         // FIXME: This is a hack to get Preferences into a state where it's safe to call newAccount()
         preferences.clearAccounts()
-
-        return preferences.newAccount()
+        return preferences.newAccount().apply {
+            incomingServerSettings = SERVER_SETTINGS
+            outgoingServerSettings = SERVER_SETTINGS
+        }
     }
 
     fun createBackendFolder(): BackendFolder {
@@ -158,5 +163,16 @@ class K9BackendFolderTest : K9RobolectricTest() {
         const val FOLDER_NAME = "Test Folder"
         val FOLDER_TYPE = FolderType.INBOX
         const val MESSAGE_SERVER_ID = "msg001"
+
+        private val SERVER_SETTINGS = ServerSettings(
+            type = "irrelevant",
+            host = "irrelevant",
+            port = 993,
+            connectionSecurity = ConnectionSecurity.SSL_TLS_REQUIRED,
+            authenticationType = AuthType.PLAIN,
+            username = "username",
+            password = null,
+            clientCertificateAlias = null,
+        )
     }
 }
--- a/legacy/storage/src/main/java/com/fsck/k9/storage/messages/K9MessageStoreFactory.kt
+++ b/legacy/storage/src/main/java/com/fsck/k9/storage/messages/K9MessageStoreFactory.kt
@@ -13,8 +13,18 @@ class K9MessageStoreFactory(
     private val storageFilesProviderFactory: StorageFilesProviderFactory,
     private val basicPartInfoExtractor: BasicPartInfoExtractor,
 ) : MessageStoreFactory {
+    private lateinit var folderNameSanitizer: FolderNameSanitizer
+
     override fun create(account: LegacyAccount): ListenableMessageStore {
         val localStore = localStoreProvider.getInstance(account)
+        if (account.incomingServerSettings.host.isGoogle() ||
+            account.outgoingServerSettings.host.isGoogle()
+        ) {
+            if (!this::folderNameSanitizer.isInitialized) {
+                folderNameSanitizer = FolderNameSanitizer(lockableDatabase = localStore.database)
+            }
+            folderNameSanitizer.removeGmailPrefixFromFolders()
+        }
         val storageFilesProvider = storageFilesProviderFactory.createStorageFilesProvider(account.uuid)
         val messageStore = K9MessageStore(
             localStore.database,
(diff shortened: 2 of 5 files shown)


asoucar (approved):
LGTM thank you!

kewisch:
Is this something we could rather change in the folder modeling code to just skip the [Gmail] folder completely, rather than just formatting away the name?

Also note that sometimes the folder is called [Google Mail], and we should only be doing this on gmail based accounts.

asoucar:
> Is this something we could rather change in the folder modeling code to just skip the [Gmail] folder completely, rather than just formatting away the name?
> 
> Also note that sometimes the folder is called [Google Mail], and we should only be doing this on gmail based accounts.

Its my (admittedly imperfect) understanding that the first section of code is doing the filtering out? Clearly it would need some additional robustness if the folder has alternate names

asoucar:
Thank you for the update! While reviewing this again and thinking further about the problem, I think this is heading in the right direction but is starting with a fundamental gap. Right now this (and other current filtering ideas) is happening on the UI layer. Your fix would have the app attempting to update the database every time the `MessageList` screen is created. While the display name for a folder is a UI component, the issue is at its core an issue with the data model. That makes addressing it at the UI layer a band-aid at best.
I think your actual filter code is actually pretty comprehensive. What if we could apply it as the database was being populated? That way it only has to happen once (with perhaps a migration step for previously loaded folder structures), decoupling the data model from the UI for better separation of concerns. Potentially there would then have to be some handling for when new emails are fetched to ensure they end up in the appropriate folder too.

shamim-emon:
> Thank you for the update! While reviewing this again and thinking further about the problem, I think this is heading in the right direction but is starting with a fundamental gap. Right now this (and other current filtering ideas) is happening on the UI layer. Your fix would have the app attempting to update the database every time the `MessageList` screen is created. While the display name for a folder is a UI component, the issue is at its core an issue with the data model. That makes addressing it at the UI layer a band-aid at best. I think your actual filter code is actually pretty comprehensive. What if we could apply it as the database was being populated? That way it only has to happen once (with perhaps a migration step for previously loaded folder structures), decoupling the data model from the UI for better separation of concerns. Potentially there would then have to be some handling for when new emails are fetched to ensure they end up in the appropriate folder too.

@asoucar with the latest approach I'm doing following:
1.  I'm removing Gmail/Google Mail prefix while folders are being created [here](https://github.com/shamim-emon/thunderbird-android/blob/9861c176fd5cabbb8af01f0d27860df8afc96938/legacy/storage/src/main/java/com/fsck/k9/storage/messages/CreateFolderOperations.kt#L18)
2. Also to deal with previously loaded folders containing Gmail/Google Mail prefix I am calling `removeGmailPrefixFromFolders()` when instance of `CreateFolderOperations` is being created like [this](https://github.com/shamim-emon/thunderbird-android/blob/9861c176fd5cabbb8af01f0d27860df8afc96938/legacy/storage/src/main/java/com/fsck/k9/storage/messages/CreateFolderOperations.kt#L9-L11). Calling `removeGmailPrefixFromFolders()` over there ensures it's called once when the instance of `CreateFolderOperations` is being injected.

Please let me know what do you think of this approach? Is it alright and/or anything else needed?

asoucar:
Thanks @shamim-emon ! I think this is 90% of the way there. As it stands right now I believe this will query all db folders, whether or not it is a Gmail account. For efficiency we should only do the folder filtering for a gmail account. But otherwise I think this is a solid solution

legacy/storage/src/main/java/com/fsck/k9/storage/messages/K9MessageStoreFactory.kt
```diff
@@ -13,8 +13,18 @@ class K9MessageStoreFactory(
...
     override fun create(account: LegacyAccount): ListenableMessageStore {
         val localStore = localStoreProvider.getInstance(account)
+        if (account.incomingServerSettings.host.contains("gmail") ||
+            account.incomingServerSettings.host.contains("gmail")
+        ) {
+            if (!this::folderNameSanitizer.isInitialized) {
+                folderNameSanitizer = FolderNameSanitizer(lockableDatabase = localStore.database)
+            }
+            folderNameSanitizer.removeGmailPrefixFromFolders()
+        }
```
asoucar, line 27: There is an old deprecated `isGoogle` function in the app that shows another hostname case for gmail. Obviously we shouldn't use that one but it does seem like `gmail` alone can't be used as the filter

asoucar (changes requested):

legacy/storage/src/main/java/com/fsck/k9/storage/messages/K9MessageStoreFactory.kt
shamim-emon: @asoucar Also added check for  googlemail

legacy/storage/src/main/java/com/fsck/k9/storage/messages/K9MessageStoreFactory.kt
```diff
@@ -13,8 +13,18 @@ class K9MessageStoreFactory(
     private val storageFilesProviderFactory: StorageFilesProviderFactory,
     private val basicPartInfoExtractor: BasicPartInfoExtractor,
 ) : MessageStoreFactory {
+    private lateinit var folderNameSanitizer: FolderNameSanitizer
+
     override fun create(account: LegacyAccount): ListenableMessageStore {
         val localStore = localStoreProvider.getInstance(account)
+        if (account.incomingServerSettings.host.isGoogle() ||
+            account.incomingServerSettings.host.isGoogle()
+        ) {
```
asoucar, line 22: These are both the same

legacy/core/src/test/java/com/fsck/k9/mailstore/K9BackendFolderTest.kt
```diff
@@ -158,5 +163,16 @@ class K9BackendFolderTest : K9RobolectricTest() {
...
+
+        private val SERVER_SETTINGS = ServerSettings(
+            type = "irrelevant",
+            host = "irrelevant",
+            port = 993,
+            connectionSecurity = ConnectionSecurity.SSL_TLS_REQUIRED,
+            authenticationType = AuthType.PLAIN,
+            username = "username",
+            password = null,
+            clientCertificateAlias = null,
```
asoucar, line 175: Is there a reason to add these settings to the tests in this PR? Just a question/nit, not necessarily a change request

asoucar (changes requested):

legacy/core/src/test/java/com/fsck/k9/mailstore/K9BackendFolderTest.kt
shamim-emon: @asoucar These tests involve  calling K9MessageStoreFactory.fun create(account: LegacyAccount): ListenableMessageStore
Since I  have added the check `account.incomingServerSettings.host.isGoogle()` over there, I need to add them otherwise there is null pointer exception due to null settings in tests.

legacy/storage/src/main/java/com/fsck/k9/storage/messages/K9MessageStoreFactory.kt
shamim-emon: @asoucar sorry, typo, the other one meant to be outgoingServerSettings. Fixed it,

asoucar (approved):
Thank your work on this!

PR #9234, opened 2025-05-28
Fix crash on Android 7 when fetching emails.

terofeev:
Fixes https://github.com/thunderbird/thunderbird-android/issues/9184

It appears that `desugar_jdk_libs` has been split into three different configurations. The default one no longer supports desugaring for `java.nio` packages.

See the following for more details:
- [CHANGELOG – Version 2.0.0 (2022-09-16)](https://github.com/google/desugar_jdk_libs/blob/master/CHANGELOG.md#version-200-2022-09-16)
- [Java 11+ NIO support in Android desugaring](https://developer.android.com/studio/write/java11-nio-support-table)

### DIFF
--- a/build-plugin/src/main/kotlin/thunderbird.app.android.gradle.kts
+++ b/build-plugin/src/main/kotlin/thunderbird.app.android.gradle.kts
@@ -31,7 +31,7 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


wmontwe (approved):
Thank you!

PR #9250, opened 2025-06-03
feat(core): Add logging KMP edition

wmontwe:
Fixes [#9265](https://github.com/thunderbird/thunderbird-android/issues/9265)

This pull request introduces a comprehensive logging framework for the application, including a new logging API, implementations for composite and console logging, and integration into the existing app infrastructure.

This should support our logging efforts by offering a flexible logging system that allows multiple `LogSink` for various needs.

Main motivation was the dependency on `mail:common` for the logging definition and implementation, with various different implementations accross the application. These APIs are also a direct copy of the Timber API which is not a good base for cross-platform use.

For compatibility a static Log implementation was added to be able to migrate in class logging. But I think this approach should only be used for legacy code that can't be easily changed otherwise.

This helps to continue my work on the profile indicator changes and account settings. A replacement of the exisitng logging infrastructure should be done as a separate task.

If this API is reviewed, it could be the first candiate to be a shared KMP component.

The use of all existing logging infrastructure was replaced by the new one.

## Logging Framework

  - Added a `Logger` interface and its default implementation (`DefaultLogger`) to support logging at various levels (e.g., VERBOSE, DEBUG, INFO, WARN, ERROR).
  - Introduced `LogEvent`, `LogLevel`, and `LogSink` to define the structure and behavior of log events and sinks.
  - Implemented `CompositeLogSink` to aggregate multiple log sinks and forward log events to them.
  - Added a console-based logging sink to output log events to the console.
  - Added `appCommonCoreModule` to provide the logging components (e.g., `Logger`, `LogSink`, `LogLevel`) via dependency injection.
  - Integrated the `Logger` into the `BaseApplication` class and assigned it to the legacy `Log` object for backward compatibility.
  - Added comprehensive tests for `DefaultLogger`, `LogSink`, and related components to ensure correct behavior and coverage.


core/logging/impl-legacy/src/commonMain/kotlin/net/thunderbird/core/logging/legacy/Log.kt
```diff
@@ -0,0 +1,46 @@
...
+ *     val sink: LogSink = // Your LogSink implementation
+ *     val logger: Logger = DefaultLogger(sink)
+ *
+ *     Log.logger = logger
+ *     Log.i("Application started")
+ *     // Your application code here
+ *  }
+ * ```
+ */
+object Log : Logger {
```
rafaeltonholo, line 30: As this is only to keep our current approach compatible, I would consider annotating it with `@Discouraged` or even with  `@Deprecated` as this is inside a "legacy" module.

Example:
```kotlin
@Discouraged(
    message = 
        "Use only as a last resort. You must prefer retrieving a " +
        "net.thunderbird.core.logging.Logger instance via dependency  " +
        "injection.",
)
object Log : Logger {
    // ... implementation
}
```

core/logging/impl-console/src/jvmMain/kotlin/net/thunderbird/core/logging/console/JvmConsoleLogSink.kt
```diff
@@ -0,0 +1,23 @@
...
+import net.thunderbird.core.logging.LogEvent
+import net.thunderbird.core.logging.LogLevel
+import net.thunderbird.core.logging.LogSink
+
+internal class JvmConsoleLogSink(
+    override val level: LogLevel,
+) : LogSink {
+    override fun log(event: LogEvent) {
+        val composedMessage = if (event.tag != null) "[${event.tag}] ${event.message}" else event.message
+        println("[$level] $composedMessage")
```
rafaeltonholo: My only concern about using `println()` in the JVM source set is that when we call logs within a JVM library consumed by Android, logcat will print the logs using `System.out` as the tag. This makes it difficult to find relevant information during debugging. 

I believe we can also use Timber in JVM environments. If that’s the case, I would consider using it as well.

core/logging/api/src/commonMain/kotlin/net/thunderbird/core/logging/Logger.kt
```diff
@@ -0,0 +1,71 @@
+package net.thunderbird.core.logging
+
+/**
+ * A logging interface that provides methods for logging messages at specific log levels.
+ */
+interface Logger {
```
rafaeltonholo, line 6: I suggest considering the addition of lambda versions for all methods in this interface. Logs can often negatively impact performance due to the transformations needed to format the message.

For this reason, many logging libraries ([Kermit](https://kermit.touchlab.co/docs/#2-log) is one good example) provide arguments as parameters for the log call, ensuring that the message transformation occurs only when the log level is appropriate.

Since KMP does not yet offer a `String.format` implementation, we need to rely on string interpolation to achieve similar functionality. However, this means that the message transformation takes place in the caller.

By implementing an API that utilizes lambda expressions, we can ensure that the transformation (string interpolation) happens only when the log is at the appropriate level. This approach allows for detailed debug and verbose logging without compromising performance.

core/logging/api/src/commonMain/kotlin/net/thunderbird/core/logging/Logger.kt
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @param tag An optional [LogTag] to categorize the log message.
+     * @param message The [LogMessage] to log.
+     * @param throwable An optional throwable to log.
+     */
+    fun verbose(
+        tag: LogTag? = null,
+        message: LogMessage,
+        throwable: Throwable? = null,
+    )
```
rafaeltonholo: This would be an example of the `verbose` method using the lambda approach:

```kotlin
fun verbose(
    tag: LogTag? = null,
    throwable: Throwable? = null,
    message: () -> LogMessage,
)

@Discouraged(
    message = 
        "Log messages can perform badly when heavy computation is done in " +
        "its interpolation or when introduced within loops. Use wisely or consider " +
        "using verbose(tag: LogTag?, throwable: Throwable?, message: () -> LogMessage) " +
        "instead.",
)
fun verbose(
    tag: LogTag? = null,
    message: LogMessage,
    throwable: Throwable? = null,
) {
    verbose(
        tag = tag,
        throwable = throwable,
        message = { message() },
    ) 
}
```

rafaeltonholo (commented):
I loved to see the introduction of a more robust Logging API in our project!! I have left a few comments for your appreciation.

Please let me know if you think they make sense :)

core/logging/impl-legacy/src/commonMain/kotlin/net/thunderbird/core/logging/legacy/Log.kt
wmontwe: That makes total sense, I was very hesitant to add this at all, but there where too many cases where Timber got hard wired.

core/logging/impl-console/src/jvmMain/kotlin/net/thunderbird/core/logging/console/JvmConsoleLogSink.kt
wmontwe: Ideally the logger should be injected and it should default to the calling platform version. So I guess this is only used for desktop or server apps.

core/logging/api/src/commonMain/kotlin/net/thunderbird/core/logging/Logger.kt
wmontwe: I didn't use Kermit yet, but that makes sense. I could change the message to a lambda only. Just need to check the Java interop.

core/logging/impl-console/src/jvmMain/kotlin/net/thunderbird/core/logging/console/JvmConsoleLogSink.kt
rafaeltonholo: I see! If we only call the `platformLogSink()` is only called once when the application starts, then it is fine! Thanks for explaining that

asoucar:
Just to confirm my understanding here: If someone were to want to either migrate some old logging or implement a new logging feature (such as log to file or analytics), they would need to add a new type of `LogSink` similar to the `AndroidConsoleLogSink`, correct?

wmontwe:
> Just to confirm my understanding here: If someone were to want to either migrate some old logging or implement a new logging feature (such as log to file or analytics), they would need to add a new type of `LogSink` similar to the `AndroidConsoleLogSink`, correct?

Yes, the `ConsoleLogSink` is a good example to start on.

core/logging/impl-legacy/src/commonMain/kotlin/net/thunderbird/core/logging/legacy/Log.kt
wmontwe: @rafaeltonholo seems like that the `Discouraged` annotation is not available.

core/logging/impl-console/src/commonMain/kotlin/net/thunderbird/core/logging/console/BaseConsoleLogSink.kt
```diff
@@ -0,0 +1,103 @@
...
+    protected open fun processTag(tag: String): String {
+        return tag
+    }
+
+    /**
+     * Gets the pattern used to identify anonymous classes in class names.
+     *
+     * @return The pattern for anonymous classes.
+     */
+    protected abstract fun getAnonymousClassPattern(): java.util.regex.Pattern
```
rafaeltonholo: This will fail to compile once we add `ios` as a target on KMP. We should avoid using `java.*` classes inside the `commonMain` module even if they are available for now.

Use `kotlin.text.Regex` instead

core/logging/impl-console/src/commonMain/kotlin/net/thunderbird/core/logging/console/BaseConsoleLogSink.kt
```diff
@@ -0,0 +1,103 @@
...
+        return rawTag?.let { processTag(it) }
+    }
+
+    /**
+     * Extracts a tag from the stack trace.
+     *
+     * @return The extracted tag, or null if no suitable tag could be found.
+     */
+    protected fun extractTagFromStackTrace(): String? {
+        return Throwable().stackTrace
```
rafaeltonholo: `Throwable.stackTrace` is a JVM only property.

core/logging/impl-console/src/commonMain/kotlin/net/thunderbird/core/logging/console/BaseConsoleLogSink.kt
```diff
@@ -0,0 +1,103 @@
...
+import net.thunderbird.core.logging.LogSink
+
+/**
+ * An abstract base class for console log sinks that provides common functionality.
+ *
+ * This class handles tag extraction from stack traces and other common operations.
+ *
+ * @param level The minimum [LogLevel] for messages to be logged.
+ */
+abstract class BaseConsoleLogSink(
```
rafaeltonholo: The implementation of this class requires specific classes from `java.lang`. We might need to consider an alternative approach, such as using `expect class` instead, and moving the current implementation to `androidMain` or `jvmMain`.

If we choose the `expect class` approach, our current project setup would require us to duplicate the implementation for both Android and JVM. However, there is a way to [create a common implementation](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-hierarchy.html#manual-configuration) for both platforms. Please let me know if you need assistance with that.

core/logging/impl-console/src/commonMain/kotlin/net/thunderbird/core/logging/console/BaseConsoleLogSink.kt
```diff
@@ -0,0 +1,103 @@
...
+            ?.let(::createStackElementTag)
+    }
+
+    /**
+     * Creates a tag from a stack trace element.
+     *
+     * @param element The stack trace element to create a tag from.
+     * @return The created tag.
+     */
+    protected fun createStackElementTag(element: StackTraceElement): String {
```
rafaeltonholo: `StackTraceElement` is a JVM only class

core/logging/impl-legacy/src/commonMain/kotlin/net/thunderbird/core/logging/legacy/Log.kt
rafaeltonholo: `@Discouraged` is available via `androidx.annotation:annotation` and would require adding the dependency in the `commonMain` dependencies

core/logging/api/src/commonMain/kotlin/net/thunderbird/core/logging/Logger.kt
rafaeltonholo, line 6: I would like to suggest moving the `message` parameter of all methods in this interface to the last position. This change will allow us to call these methods using a trailing lambda when the other parameters are not needed.

core/logging/impl-legacy/src/commonMain/kotlin/net/thunderbird/core/logging/legacy/Log.kt
```diff
@@ -0,0 +1,139 @@
...
+    private fun formatMessage(message: String?, args: Array<out Any?>): String {
+        return if (message == null) {
+            ""
+        } else if (args.isEmpty()) {
+            message
+        } else {
+            try {
+                String.format(message, *args)
+            } catch (e: Exception) {
+                "$message (Error formatting message: $e)"
```
rafaeltonholo: I would also include the list of arguments in the message, so it could help to diagnose which parameter is breaking the formatting

mail/common/build.gradle.kts
```diff
@@ -11,6 +11,8 @@ if (testCoverageEnabled) {
 dependencies {
     api(libs.jetbrains.annotations)
 
+    api(projects.core.logging.implLegacy)
```
rafaeltonholo, line 14: Why would we need to use `api` instead of `implementation` in this case?

rafaeltonholo (changes requested):
Great work! Thank you for removing the references to `Timber` throughout the project; that will greatly aid in consolidating our logging system.

However, I have a significant concern about the implementation of the `BaseConsoleLoggingSync` class. It could lead to incompatibilities when we add other targets to the project, so this issue needs to be addressed.

core/logging/impl-legacy/src/commonMain/kotlin/net/thunderbird/core/logging/legacy/Log.kt
wmontwe: Thanks

mail/common/build.gradle.kts
wmontwe: It was the home of the old logger definition, so `api` reduces the amount of imports that need to be declared.

wmontwe:
> `BaseConsoleLoggingSync`

The `BaseConsoleLoggingSync` is basically the old `TimberLogger` implementation. The tags might be handy, but it doesn't scale to all platforms.

Before continuing, do you think this behavior is worth keeping?

rafaeltonholo:
> > `BaseConsoleLoggingSync`
> 
> The `BaseConsoleLoggingSync` is basically the old `TimberLogger` implementation. The tags might be handy, but it doesn't scale to all platforms.
> 
> Before continuing, do you think this behavior is worth keeping?

I have mixed feelings about this, tbh. I strongly believe that having a tag for logs is crucial, even across different platforms, as it helps us filter logs more effectively. I can't speak to iOS logging since I'm not familiar with it, so I'll defer to @toddheasley regarding the necessity of tags on that platform.

That said, an automatic tagging system can be beneficial to reduce the overhead of creating a tag for each class or function you want to track and always passing it as a parameter. However, Android Studio already provides a built-in code snippet that simplifies the creation of log tags, so I haven't found this to be a significant issue.

If you believe this is important, I would be willing to create a platform implementation to derive the tags.

We could have something like:
<details>
<summary>Implementation brainstorm</summary>

```kotlin
// commonMain
interface LogTagComposer {
    fun composeTag(event: LogEvent): String?
}

expect fun LogTagComposer(processTag: (String) -> String): LogTagComposer

// androidMain/jvmMain (jvmCommonMain)
actual fun LogTagComposer(processTag: (String) -> String) = object : LogTagComposer {
    fun composeTag(event: LogEvent): String? {
        // If a tag is provided, use it; otherwise, extract it from the stack trace
        val rawTag = event.tag ?: extractTagFromStackTrace()
        // Process the tag before returning it
        return rawTag?.let { processTag(it) }
    }

    private fun extractTagFromStackTrace(): String? {
        return Throwable().stackTrace
            .firstOrNull { it.className !in getIgnoreClasses() }
            ?.let(::createStackElementTag)
    }

    private fun createStackElementTag(element: StackTraceElement): String {
        var tag = element.className.substringAfterLast('.')
        val matcher = getAnonymousClassPattern().matcher(tag)
        if (matcher.find()) {
            tag = matcher.replaceAll("")
        }
        return processTag(tag)
    }
}

// and in the future, appleMain
actual fun LogTagComposer(processTag: (String) -> String) = object : LogTagComposer {
    fun composeTag(event: LogEvent): String? {
        // If a tag is provided, use it; otherwise, extract it from the stack trace
        val rawTag = event.tag ?: extractTagFromStackTrace()
        // Process the tag before returning it
        return processTag(rawTag)
    }

    private fun extractTagFromStackTrace(): String {
        return Throwable()
            .stackTraceToString()
            .lines()
            .first()
            .somethingElse()
    }
}
```
</details>

TLDR: I'm okay with either explicitly passing a tag for each log or having an auto-tagging system; however, I would avoid having no log tags.

toddheasley:
> I strongly believe that having a tag for logs is crucial, even across different platforms, as it helps us filter logs more effectively. I can't speak to iOS logging since I'm not familiar with it, so I'll defer to @toddheasley regarding the necessity of tags on that platform.

A tag, or "category," is required by the Apple-provided [unified system log](https://developer.apple.com/documentation/oslog) that we'll be logging to. So, yes, at least one category is required. In practice, I don't think I've ever used the category for filtering logs.

wmontwe:
> > > `BaseConsoleLoggingSync`
> > 
> > 
> > The `BaseConsoleLoggingSync` is basically the old `TimberLogger` implementation. The tags might be handy, but it doesn't scale to all platforms.
> > Before continuing, do you think this behavior is worth keeping?
> 
> I have mixed feelings about this, tbh. I strongly believe that having a tag for logs is crucial, even across different platforms, as it helps us filter logs more effectively. I can't speak to iOS logging since I'm not familiar with it, so I'll defer to @toddheasley regarding the necessity of tags on that platform.
> 
> That said, an automatic tagging system can be beneficial to reduce the overhead of creating a tag for each class or function you want to track and always passing it as a parameter. However, Android Studio already provides a built-in code snippet that simplifies the creation of log tags, so I haven't found this to be a significant issue.
> 
> If you believe this is important, I would be willing to create a platform implementation to derive the tags.
> 
> We could have something like:
> Implementation brainstorm
> 
> ```kotlin
> // commonMain
> interface LogTagComposer {
>     fun composeTag(event: LogEvent): String?
> }
> 
> expect fun LogTagComposer(processTag: (String) -> String): LogTagComposer
> 
> // androidMain/jvmMain (jvmCommonMain)
> actual fun LogTagComposer(processTag: (String) -> String) = object : LogTagComposer {
>     fun composeTag(event: LogEvent): String? {
>         // If a tag is provided, use it; otherwise, extract it from the stack trace
>         val rawTag = event.tag ?: extractTagFromStackTrace()
>         // Process the tag before returning it
>         return rawTag?.let { processTag(it) }
>     }
> 
>     private fun extractTagFromStackTrace(): String? {
>         return Throwable().stackTrace
>             .firstOrNull { it.className !in getIgnoreClasses() }
>             ?.let(::createStackElementTag)
>     }
> 
>     private fun createStackElementTag(element: StackTraceElement): String {
>         var tag = element.className.substringAfterLast('.')
>         val matcher = getAnonymousClassPattern().matcher(tag)
>         if (matcher.find()) {
>             tag = matcher.replaceAll("")
>         }
>         return processTag(tag)
>     }
> }
> 
> // and in the future, appleMain
> actual fun LogTagComposer(processTag: (String) -> String) = object : LogTagComposer {
>     fun composeTag(event: LogEvent): String? {
>         // If a tag is provided, use it; otherwise, extract it from the stack trace
>         val rawTag = event.tag ?: extractTagFromStackTrace()
>         // Process the tag before returning it
>         return processTag(rawTag)
>     }
> 
>     private fun extractTagFromStackTrace(): String {
>         return Throwable()
>             .stackTraceToString()
>             .lines()
>             .first()
>             .somethingElse()
>     }
> }
> ```
> 
> TLDR: I'm okay with either explicitly passing a tag for each log or having an auto-tagging system; however, I would avoid having no log tags.

I think it could be handy, but should not be part of the `LogSink`. Instead it could be part of the `DefaultLogger` to enrich the log tags when not present, taking your idea of a `LogTagComposer`.

I'll remove the implementation from the `ConsoleLogSink` for now and we could add the tag handling in another task.

wmontwe:
I defered the iOS implementation, as I would like to move this dependency to a separate repository once we settled on the approach. This also avoids the build time impact we face with the current Android project setup.

rafaeltonholo (approved):
Great work! Thanks for this contribution!

PR #9260, opened 2025-06-04
Convert MessageActions class from Java to Kotlin

### LINKED ISSUE #9259: Convert MessageActions class from Java to Kotlin
We should migrate the `com.fsck.k9.activity.compose.MessageActions` class to Kotlin to align with the ongoing modernization efforts and Kotlin-first development approach across the codebase.

shamim-emon:
- Fixes #9259


legacy/ui/legacy/src/main/java/com/fsck/k9/activity/compose/MessageActions.kt
```diff
@@ -0,0 +1,119 @@
...
+     * will be used. If there is no default account set, user will be sent to AccountSetup
+     * activity.
+     */
+    @JvmStatic
+    fun actionCompose(context: Context, account: LegacyAccount?) {
+        val defaultAccount = Preferences.getPreferences().defaultAccount
+        if (account == null && defaultAccount == null) {
+            FeatureLauncherActivity.launch(context, AccountSetup)
+        } else {
+            val accountUuid = account?.uuid ?: defaultAccount!!.uuid
```
rafaeltonholo: I recommend avoiding the use of the double-bang operator (`!!`). While we expect that either `account` or `defaultAccount` will not be null in this situation, using `!!` could result in a `NullPointerException` if we encounter an unexpected state. This exception does not provide much context for debugging.

Instead, let's use `kotlin.requireNotNull`. This approach allows us to include additional context, making it easier to identify and resolve issues if something unexpected occurs.

```suggestion
            val accountUuid = account?.uuid ?: requireNotNull(defaultAccount?.uuid) {
                "Unexpected state. At this point, either account ($account) or defaultAccount " +
                    "($defaultAccount) must have a value."
            }
```

rafaeltonholo (approved):
Thank you for converting `MessageActions` to Kotlin! I left a comment that I believe is worth considering, but it does not block this PR.

wmontwe:
@shamim-emon Could you please split the Java to Kotlin conversion into two commits. The first for the `.java` to `.kt` rename and the second for the actual changes.

shamim-emon:
> @shamim-emon Could you please split the Java to Kotlin conversion into two commits. The first for the `.java` to `.kt` rename and the second for the actual changes.

@wmontwe  Done!

PR #9271, opened 2025-06-05
Add account settings profile indicator - part 2

wmontwe:
Sorry this got huge. I'll break up changes for the next iteration.

This PR implements the second part of the account profile indicator feature, adding support for avatars in account settings. The UI implementation will follow in a subsequent PR.

This also introduces significant refactoring and modernization to introduce similar Account Id 
handling as on iOS and improve testability of the code base.

## Avatar Support

- Added support for three types of avatars:
  - **Monogram**: Uses the first two letters of the user's name or email address
  - **Image**: Uses a custom image specified by a URI
  - **Icon**: Uses a named icon

- Added new account settings:
  - `avatarType`: Enum setting to specify the type of avatar (MONOGRAM, IMAGE, or ICON)
  - `avatarMonogram`: String setting for the monogram text
  - `avatarImageUri`: String setting for custom image URIs
  - `avatarIconName`: String setting for icon names

- Implemented `AccountSettingsUpgraderTo104` to automatically generate monogram avatars for existing accounts:
  - Uses the first two letters of the user's name (or email if name is not available)
  - Converts the letters to uppercase
  - Uses "XX" as the default monogram if neither name nor email is available

## Architecture Changes

- Introduced `AccountId` as a type alias for `Id<Account>` to:
  - Uniquely identify accounts in the system
  - Link avatar settings to specific accounts
  - Enable type-safe account references
  - Align with iOS implementation using the `Identifiable` interface

- Implemented the `DataMapper` pattern to separate domain and storage layers:
  - Added `AccountAvatarDataMapper` interface extending `DataMapper<AccountAvatar, AvatarDto>`
  - Created `DefaultAccountAvatarDataMapper` implementation for bidirectional mapping
  - Defined clear methods for domain-DTO conversions:
    - `toDomain(dto: AvatarDto): AccountAvatar`
    - `toDto(domain: AccountAvatar): AvatarDto`

## Refactoring Changes

- Renamed classes to follow consistent naming conventions:
  - `CommonAccountDefaultsProvider` → `DefaultAccountDefaultsProvider`
  - `CommonAccountProfileLocalDataSource` → `DefaultAccountProfileLocalDataSource`
  - `CommonLegacyAccountWrapperManager` → `DefaultLegacyAccountWrapperManager`

- Improved dependency management:
  - Replaced `K9.DEFAULT_VISIBLE_LIMIT` with `AccountDefaultsProvider.DEFAULT_VISIBLE_LIMIT`
  - Removed direct dependencies on the `K9` class

- Added new data mappers to streamline domain-to-DTO conversions:
  - `AccountProfileDataMapper`
  - `DefaultLegacyAccountWrapperDataMapper`

### DIFF
--- a/core/architecture/api/src/commonMain/kotlin/net/thunderbird/core/architecture/model/BaseIdFactory.kt
+++ b/core/architecture/api/src/commonMain/kotlin/net/thunderbird/core/architecture/model/BaseIdFactory.kt
@@ -19,7 +19,7 @@ import kotlin.uuid.Uuid
  */
 @OptIn(ExperimentalUuidApi::class)
 abstract class BaseIdFactory<T> : IdFactory<T> {
-    override fun create(raw: String): Id<T> = Id(Uuid.parse(raw))
+    override fun of(raw: String): Id<T> = Id(Uuid.parse(raw))
 
-    override fun new(): Id<T> = Id(Uuid.random())
+    override fun create(): Id<T> = Id(Uuid.random())
 }
--- a/core/architecture/api/src/commonMain/kotlin/net/thunderbird/core/architecture/model/IdFactory.kt
+++ b/core/architecture/api/src/commonMain/kotlin/net/thunderbird/core/architecture/model/IdFactory.kt
@@ -6,17 +6,17 @@ package net.thunderbird.core.architecture.model
 interface IdFactory<T> {
 
     /**
-     * Creates an ID from a raw string representation.
+     * Creates an [Id] from a raw string representation.
      *
      * @param raw The raw string representation of the ID.
      * @return An instance of [Id] representing the ID.
      */
-    fun create(raw: String): Id<T>
+    fun of(raw: String): Id<T>
 
     /**
-     * Generates a new ID.
+     * Creates a new [Id].
      *
-     * @return A new instance of [Id] representing the generated ID.
+     * @return A new instance of [Id] representing the created ID.
      */
-    fun new(): Id<T>
+    fun create(): Id<T>
 }
--- a/app-common/src/test/kotlin/net/thunderbird/app/common/account/data/DefaultAccountProfileLocalDataSourceTest.kt
+++ b/app-common/src/test/kotlin/net/thunderbird/app/common/account/data/DefaultAccountProfileLocalDataSourceTest.kt
@@ -27,7 +27,7 @@ class DefaultAccountProfileLocalDataSourceTest {
     @Test
     fun `getById should return account profile`() = runTest {
         // arrange
-        val accountId = AccountIdFactory.new()
+        val accountId = AccountIdFactory.create()
         val legacyAccount = createLegacyAccount(accountId)
         val accountProfile = createAccountProfile(accountId)
         val testSubject = createTestSubject(legacyAccount)
@@ -41,7 +41,7 @@ class DefaultAccountProfileLocalDataSourceTest {
     @Test
     fun `getById should return null when account is not found`() = runTest {
         // arrange
-        val accountId = AccountIdFactory.new()
+        val accountId = AccountIdFactory.create()
         val testSubject = createTestSubject(null)
 
         // act & assert
@@ -53,7 +53,7 @@ class DefaultAccountProfileLocalDataSourceTest {
     @Test
     fun `update should save account profile`() = runTest {
         // arrange
-        val accountId = AccountIdFactory.new()
+        val accountId = AccountIdFactory.create()
         val legacyAccount = createLegacyAccount(accountId)
         val accountProfile = createAccountProfile(accountId)
 
--- a/core/android/account/src/main/kotlin/net/thunderbird/core/android/account/LegacyAccount.kt
+++ b/core/android/account/src/main/kotlin/net/thunderbird/core/android/account/LegacyAccount.kt
@@ -28,7 +28,7 @@ open class LegacyAccount(
 ) : Account, BaseAccount {
 
     // [Account]
-    override val id: AccountId = AccountIdFactory.create(uuid)
+    override val id: AccountId = AccountIdFactory.of(uuid)
 
     // [BaseAccount]
     @get:Synchronized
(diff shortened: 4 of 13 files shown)


app-common/src/main/kotlin/net/thunderbird/app/common/account/data/DefaultAccountProfileLocalDataSource.kt
```diff
@@ -4,34 +4,33 @@ import kotlinx.coroutines.flow.Flow
...
 
-internal class CommonAccountProfileLocalDataSource(
+internal class DefaultAccountProfileLocalDataSource(
     private val accountManager: LegacyAccountWrapperManager,
+    private val dataMapper: AccountProfileDataMapper,
 ) : AccountProfileLocalDataSource {
 
     override fun getById(accountId: AccountId): Flow<AccountProfile?> {
-        return accountManager.getById(accountId.value)
+        return accountManager.getById(accountId.asRaw())
```
rafaeltonholo: Since we are refactoring classes that use `AccountId`, wouldn’t it be beneficial to add a method overload for `accountManager.getById(id)` to also accept `AccountId`?

Then we could drop the `id.asRaw()` that is repeating along the class

core/architecture/api/src/commonMain/kotlin/net/thunderbird/core/architecture/data/DataMapper.kt
```diff
@@ -0,0 +1,12 @@
+package net.thunderbird.core.architecture.data
+
+/**
+ * Mapper definition for converting between domain models and data transfer objects (DTOs).
+ *
+ * @param Domain The domain model type.
+ * @param Dto The data transfer object type.
+ */
+interface DataMapper<Domain, Dto> {
```
rafaeltonholo: nit: Since both `Domain` and `Dto` are generic types rather than actual types, I would prefix them with `T`, `TDomain` and `TDto`, for example. This makes it clear that they are not concrete types, but rather type parameters, in my opinion.

core/architecture/api/src/commonMain/kotlin/net/thunderbird/core/architecture/model/BaseIdFactory.kt
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * ```kotlin
+ * class AccountIdFactory : BaseIdFactory<AccountId>()
+ * ```
+ *
+ * @param T The type of the ID.
+ */
+@OptIn(ExperimentalUuidApi::class)
+abstract class BaseIdFactory<T> : IdFactory<T> {
+    override fun create(raw: String): Id<T> = Id(Uuid.Companion.parse(raw))
```
rafaeltonholo: nit: `.Companion.` is reduntant

core/architecture/api/src/commonMain/kotlin/net/thunderbird/core/architecture/model/IdFactory.kt
```diff
@@ -0,0 +1,22 @@
...
+     */
+    fun create(raw: String): Id<T>
+
+    /**
+     * Generates a new ID.
+     *
+     * @return A new instance of [Id] representing the generated ID.
+     */
+    fun new(): Id<T>
+}
```
rafaeltonholo: I believe we are using the Abstract Factory pattern here, is that correct? To me, a more fitting definition of a factory would be something like this:

```kotlin
interface IdFactory<T> {
    fun of(raw: String): Id<T>
    fun create(): Id<T>
}
```

Feel free to consider, or disregard, this suggestion. There's no need to change the declaration to match my familiar conventions; I just wanted to share my thoughts.

feature/account/storage/legacy/src/main/kotlin/net/thunderbird/feature/account/storage/legacy/AccountStorageLegacyModule.kt
```diff
@@ -0,0 +1,45 @@
+package net.thunderbird.feature.account.storage.legacy
```
wmontwe, line 1: I forgot to write a DI test.

feature/account/storage/legacy/src/main/kotlin/net/thunderbird/feature/account/storage/legacy/mapper/DefaultLegacyAccountWrapperDataMapper.kt
```diff
@@ -0,0 +1,221 @@
...
+            senderName = domain.senderName
+            signatureUse = domain.signatureUse
+            signature = domain.signature
+            shouldMigrateToOAuth = domain.shouldMigrateToOAuth
+        }
+    }
+
+    private fun fromProfileDto(dto: ProfileDto, account: LegacyAccount) {
+        account.name = dto.name
+        account.chipColor = dto.color
```
wmontwe, line 219: The avatar is missing

rafaeltonholo (commented):
Great work! I've just left a few comments for your consideration

core/architecture/api/src/commonMain/kotlin/net/thunderbird/core/architecture/model/BaseIdFactory.kt
wmontwe: Hmm, Android Studio started to add `Companion` when I move classes around. Didn't catch that one.

app-common/src/main/kotlin/net/thunderbird/app/common/account/data/DefaultAccountProfileLocalDataSource.kt
wmontwe: That's a good idea, I'll change the signature as the raw id is actually not needed

core/architecture/api/src/commonMain/kotlin/net/thunderbird/core/architecture/data/DataMapper.kt
wmontwe: I have no strong opinion about this, I tried to avoid capitalized names as they may conflict with constants. So using T as a type indicator is fine.

core/architecture/api/src/commonMain/kotlin/net/thunderbird/core/architecture/model/IdFactory.kt
wmontwe: Yeah, I think that's better, will update.

PR #9272, opened 2025-06-05
In-App Error Notifications - Create AppNotification class for notification categorization

### LINKED ISSUE #9244: In-App Error Notifications - Create AppNotification class for notification categorization
To enhance our in-app notification system, we need to begin categorizing our notifications. This task involves creating an `AppNotification` type, which will be utilized by the `NotificationManager` to appropriately select the method of notifying users.

rafaeltonholo:
Resolves #9244.

## Description
This pull request establishes the foundation for revamping the notification system, which will allow in-app notification errors to be handled effectively.

The goal is to create a straightforward way to manage and define notifications within the application, enabling the app to decide where to display them, either as system notifications or in-app notifications.

Additionally, it provides a centralized API for triggering notifications, simplifying the process of delivering messages to the user.

## Changes
- Migrated `com.fsck.k9.mail.MessagingException` to Kotlin.
    - Created `net.thunderbird.core.common.exception.MessagingException` as its replacement and annotated the old one with `@Deprecated`
- Migrated `ExceptionHelper` to Kotlin and created an extension function for its replacement.
- Created `Notification`, `SystemNotification`, `InAppNotification` base types
    - Created notification types that the app is currently showing to the user.
    - Categorized each notification type using `NotificationSeverity` enum.

## Defining a notification
Getting the "Authentication Error" notification we already have in the application, this should be how we define a notification using the new API:
```kotlin
@ConsistentCopyVisibility
data class AuthenticationErrorNotification private constructor( // 5
    override val id: NotificationId,
    override val title: String,
    override val contentText: String?,
    override val channel: NotificationChannel,
) : AppNotification(), SystemNotification, InAppNotification { // 1
    // 2
    override val severity: NotificationSeverity = NotificationSeverity.Fatal
    // 3
    override val actions: Set<NotificationAction> = setOf(
        NotificationAction.Retry,
        NotificationAction.UpdateServerSettings,
    )
    // 4
    override val lockscreenNotification: SystemNotification = copy(contentText = null)

    companion object {
        // 5
        suspend operator fun invoke(
            id: NotificationId,
            accountUuid: String,
            accountDisplayName: String,
        ): AuthenticationErrorNotification = AuthenticationErrorNotification(
            id = id,
            title = getString(
                resource = Res.string.notification_authentication_error_title,
                accountDisplayName,
            ),
            contentText = getString(resource = Res.string.notification_authentication_error_text),
            channel = NotificationChannel.Miscellaneous(accountUuid = accountUuid),
        )
    }
}
```
Breaking it down:

1. We create a `data class` that extends either `SystemNotification`, `InAppNotification`, or both. This allows us to identify the type of notification we are dealing with, which will later help the app determine which notification to trigger.
2. Next, we define the severity of the notification, which will be useful for ranking the impact of the error on the user. We also specify the user actions that will be available.
3. Since this is a `SystemNotification`, we can provide a different notification for the lock screen, ensuring that any sensitive data is hidden if necessary. 
4. Given that this notification contains a composition in its message, it requires specific formatting. KMP resources necessitate a suspend function to retrieve the resource; therefore, we cannot use the `data class` constructor directly. To address this, we create a factory function using the invoke operator in the class's Companion object.

The main idea of this class is to provide an easy way to create the notification object. As a result, its constructor or factory function should only include parameters that are essential for building the notification.

## Triggering a notification (Not included in this PR - Future work)
After having the notification class definition created, we just need to trigger it.

Here is an example of how it will be done once the new notification system is implemented:
```kotlin
class SomeViewModel(
    private notificationSender: NotificationSender,
    // ... other dependencies
) {
    fun auth() {
        try { /* auth logic */ }
        catch (e : Exception) {
            notificationSender
                .send(
                    AuthenticationErrorNotification(
                        id = NotificationId.next(),
                        accountUuid = account.uuid,
                        accountDisplayName = account.name,
                    ),
                )
                .onEach { result -> logger.i { "notification result = $result" } }
                .launchIn(viewModelScope)
        }
    }
}
```

The `NotificationSender` will take care of identifying where to display the notification to the use.


feature/notification/src/commonMain/kotlin/net/thunderbird/feature/notification/NotificationGroup.kt
```diff
@@ -0,0 +1,7 @@
+package net.thunderbird.feature.notification
+
+// TODO: Properly handle notification groups, adding summary, etc.
```
rafaeltonholo: This will be better handled when implementing the `SystemNotificationCommand`/`InAppNotificationCommand` classes.

feature/notification/src/commonMain/kotlin/net/thunderbird/feature/notification/command/NotificationCommand.kt
```diff
@@ -0,0 +1,52 @@
...
+         *
+         * @param TNotification The type of notification associated with the command.
+         * @property command The command that failed.
+         * @property throwable The exception that caused the failure.
+         */
+        data class Failure<in TNotification : Notification>(
+            val command: NotificationCommand<in TNotification>,
+            val throwable: Throwable,
+        ) : CommandResult
+    }
```
rafaeltonholo: Would it be better if I used the `Outcome` class instead?

wmontwe: To keep it consistent, yes.

kewisch:
@rafaeltonholo I see one non-conventional commit, can you amend that before merging?

PR #9278, opened 2025-06-06
Convert ImapResponse class from java to kotlin

### LINKED ISSUE #9275: Convert ImapResponse class from Java To Kotlin
We should migrate the `com.fsck.k9.mail.store.imap.ImapResponse` class to Kotlin to align with the ongoing modernization efforts and Kotlin-first development approach across the codebase.

shamim-emon:
- Fixes #9275


mail/protocols/imap/src/main/java/com/fsck/k9/mail/store/imap/ImapResponse.kt
```diff
@@ -0,0 +1,44 @@
+package com.fsck.k9.mail.store.imap
+
+import java.io.Serializable
+
+/**
+ * Represents a single response from the IMAP server.
+ *
+ * Tagged responses have a non-null tag.
+ * Untagged responses have a null tag.
+ * Continuation requests are identified with a `+`.
```
rafaeltonholo, line 10: Just missing part of the old java docs:

```suggestion
 * Continuation requests are identified with a `+`.
 * The object will contain all of the available tokens at the time the response is received.
```

PR #9280, opened 2025-06-06
Bump Gradle 8.14 -> 8.14.2

wmontwe:
While discussing and testing the build performance we tried some improvements. This is just a tiny tiny bit faster.


rafaeltonholo (approved):
LGTM

PR #9291, opened 2025-06-07
Convert ImapMessage class from java to kotlin

### LINKED ISSUE #9290: Convert ImapMessage class from Java to Kotlin
We should migrate the `com.fsck.k9.mail.store.imap.ImapMessage` class to Kotlin to align with the ongoing modernization efforts and Kotlin-first development approach across the codebase.

shamim-emon:
- Fixes #9290


mail/protocols/imap/src/main/java/com/fsck/k9/mail/store/imap/ImapMessage.kt
```diff
@@ -0,0 +1,13 @@
+package com.fsck.k9.mail.store.imap
+
+import com.fsck.k9.mail.internet.MimeMessage
+
+class ImapMessage(uid: String?) : MimeMessage() {
```
wmontwe: Does the uid need to be nullable?

shamim-emon: @wmontwe not necessarily. Made it non-nullable

wmontwe (approved):
Thank you!

PR #9293, opened 2025-06-07
Convert ImapResponseParserException class from java to kotlin

### LINKED ISSUE #9292: Convert ImapResponseParserException class from Java to Kotlin
We should migrate the `com.fsck.k9.mail.store.imap.ImapResponseParserException` class to Kotlin to align with the ongoing modernization efforts and Kotlin-first development approach across the codebase.

shamim-emon:
- Fixes #9292


mail/protocols/imap/src/main/java/com/fsck/k9/mail/store/imap/ImapResponseParserException.kt
```diff
@@ -0,0 +1,7 @@
+package com.fsck.k9.mail.store.imap
+
+class ImapResponseParserException : RuntimeException {
+    constructor(message: String?) : super(message)
+
+    constructor(message: String?, cause: Throwable?) : super(message, cause)
+}
```
wmontwe: You could write it more Kotlin idiomatic:

```suggestion
class ImapResponseParserException @JvmOverloads constructor(
    override val message: String? = null,
    override val cause: Throwable? = null,
) : RuntimeException(message, cause)
```

shamim-emon: Done

Issue #9314, opened 2025-06-10
Refactor Sync debug logging for new logging


asoucar:
After #9281 , rewritting the new sync debug logging to use and extend the updated logging

PR #9341, opened 2025-06-17
Create Log to File and migrate sync debug functionality

### LINKED ISSUE #9144: Add uncaught exception handler to flush log to file
Add uncaught exception handler in case of a crash, to flush persistent logging if enabled

asoucar:
Because the Sync Debug work was done almost concurrently with the new logging setup, it makes sense to migrate immediately to avoid future tech debt. 
Functionality:

- Log to file of choice
- export file log to external URI
- Add logging and log flushing at uncaught exception
- Add file logging at some existing sync related points

Closes #9144 , #9314


core/logging/impl-file/src/commonMain/kotlin/net/thunderbird/core/logging/file/FileLogSink.kt
```diff
@@ -0,0 +1,35 @@
...
+import net.thunderbird.core.logging.LogLevel
+import net.thunderbird.core.logging.LogSink
+
+interface FileLogSink : LogSink {
+    /**
+     * Exports from the logging method to the requested external file
+     *
+     * @param uriString The [String] for the URI to export the log to
+     *
+     * **/
```
wmontwe: The leading `* ` could be removed

core/logging/impl-file/src/commonMain/kotlin/net/thunderbird/core/logging/file/FileLogSink.kt
```diff
@@ -0,0 +1,35 @@
...
+    fun export(uriString: String)
+
+    /**
+     * On a crash, flushes buffer to file fo avoid log loss
+     *
+     * **/
```
wmontwe: The leading `* ` could be removed

core/logging/impl-file/src/commonMain/kotlin/net/thunderbird/core/logging/file/FileLogSink.kt
```diff
@@ -0,0 +1,35 @@
...
+    fun flushAndCloseBuffer()
+}
+
+/**
+ * A [LogSink] implementation that logs messages to a specified internal file.
+ *
+ * This sink uses the platform-specific implementations to handle logging.
+ *
+ * @param level The minimum [LogLevel] for messages to be logged.
+ * @param fileName The [String] fileName to log to
```
wmontwe, line 28: Please add the remaining property description too.

legacy/core/src/main/java/com/fsck/k9/K9.kt
```diff
@@ -32,7 +35,8 @@ object K9 : KoinComponent {
     private val generalSettingsManager: RealGeneralSettingsManager by inject()
     private val telemetryManager: TelemetryManager by inject()
     private val featureFlagProvider: FeatureFlagProvider by inject()
-    private val context: Context by inject()
+    private val syncDebugCompositeSink: CompositeLogSink by inject(named("syncDebug"))
+    private val syncDebugFileLogSink: FileLogSink by inject(named("syncDebug"))
```
wmontwe, line 39: 👍

app-common/src/main/kotlin/net/thunderbird/app/common/BaseApplication.kt
```diff
@@ -65,6 +66,7 @@ abstract class BaseApplication : Application(), WorkManagerConfiguration.Provide
         messagingListenerProvider.listeners.forEach { listener ->
             messagingController.addListener(listener)
         }
+        Thread.setDefaultUncaughtExceptionHandler(ExceptionHandler())
```
wmontwe: Could you please check if this doesn't break the crash reporting with the default handler that will upload uncaught crashes to Google if Google Play Services are present on the device. I think you need to the get the current handler and pass the exception to it in your `ExceptionHandler`.

```suggestion
        val originalHandler = Thread.getDefaultUncaughtExceptionHandler()
        Thread.setDefaultUncaughtExceptionHandler(ExceptionHandler(originalHandler))
```

and

```kotlin
class ExceptionHandler(
    private val defaultHandler: Thread.UncaughtExceptionHandler?
) : Thread.UncaughtExceptionHandler, KoinComponent {
    private val syncDebugFileLogSink: FileLogSink by inject(named("syncDebug"))

    override fun uncaughtException(t: Thread, e: Throwable) {
        Log.e("UncaughtException", e.toString(), e)
        syncDebugFileLogSink.flushAndCloseBuffer()
        // Forward to default handler
        defaultHandler?.uncaughtException(t, e)
    }
}
```

core/logging/impl-file/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
+plugins {
+    id(ThunderbirdPlugins.Library.kmp)
```
wmontwe, line 2: Add Mokkery here, in case you want to use mocks.

```suggestion
    id(ThunderbirdPlugins.Library.kmp)
    alias(libs.plugins.dev.mokkery)
```

core/logging/impl-file/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+kotlin {
+    sourceSets {
+        commonMain.dependencies {
+            implementation(projects.core.logging.api)
+        }
+        commonTest.dependencies {
+            implementation(libs.bundles.shared.jvm.test)
+        }
```
wmontwe: Remove this declaration and switch to [Mokkery](https://mokkery.dev/) please, which is fully KMP ready.

core/logging/impl-file/src/androidMain/kotlin/net/thunderbird/core/logging/file/AndroidFileLogSink.kt
```diff
@@ -0,0 +1,92 @@
...
+) : FileLogSink {
+
+    private val coroutineScope = CoroutineScope(coroutineContext + SupervisorJob())
+    private val writeFile = File(fileLocation, "$fileName.txt")
+    private val accumulatedLogs = ConcurrentHashMap<String, String>()
+
+    override fun log(event: LogEvent) {
+        try {
+            accumulatedLogs[convertLongToTime(event.timestamp)] = "priority = ${event.level}, ${event.message}"
+            createLogFile()
```
wmontwe: The file is written on every log event, this creates a lot of disk i/o. Maybe the writes could be done in batches or using a buffer with a flush interval. The `bufferedWriter` doesn't prevent disk i/o.

I also don't see where the `accumulatedLogs` are cleared after writing to file.

Should the file be appended to or always overwritten with the accumulated logs? The later could accumulate some memory over time.

core/logging/impl-file/src/androidMain/kotlin/net/thunderbird/core/logging/file/AndroidFileLogSink.kt
```diff
@@ -0,0 +1,92 @@
...
+                it.write(accumulatedLogs.entries.joinToString("\n") { it2 -> it2.key + " " + it2.value })
+            }
+        }
+        if (result.isFailure) {
+            result.exceptionOrNull()?.printStackTrace()
+        }
+    }
+
+    override fun flushAndCloseBuffer() {
+        writeFile.bufferedWriter().close()
```
wmontwe: The use of `bufferedWriter().use {` in the logsink already closes all used instances. This should rather write the `accumulatedLogs` to file ignoring any write delays, bufferings etc.

core/logging/impl-file/src/androidUnitTest/kotlin/AndroidFileLogSinkTest.android.kt
wmontwe, line 1: I think there is no need for integration tests. You could write to a temporary file and test the output when triggering logs. The `ContentResolver` and `Uri` are currently preventing tests and would require either more mocks or Robolectric. A better approach is to replace direct Android dependencies by fascades, for example a `FileSystemManager`. This would allow to write tests and help to indentify the corner cases to stabilize the current implementation.

Please also move the file to the `androidUnitTest/kotlin/net/thunderbird/core/logging/file` folder.

core/logging/impl-file/src/androidMain/kotlin/net/thunderbird/core/logging/file/AndroidFileLogSink.kt
```diff
@@ -0,0 +1,92 @@
...
+import java.util.concurrent.ConcurrentHashMap
+import kotlin.coroutines.CoroutineContext
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.launch
+import net.thunderbird.core.logging.LogEvent
+import net.thunderbird.core.logging.LogLevel
+
+private const val ANDROID_LOG_TIME_FORMAT = "MM-dd-yy hh:mm:ss.SSS"
```
wmontwe: The format should follow a specification, for example ISO 8601 (yyyy-MM-dd'T'HH:mm:ss.SSS'Z'). KotlinX DateTime already has support, see below.

core/logging/impl-file/src/androidMain/kotlin/net/thunderbird/core/logging/file/AndroidFileLogSink.kt
```diff
@@ -0,0 +1,92 @@
...
+    }
+    private fun createLogFile() =
+        coroutineScope.launch {
+            writeToLogFile()
+        }
+    private fun convertLongToTime(long: Long): String {
+        val date = Date(long)
+        val format = SimpleDateFormat(ANDROID_LOG_TIME_FORMAT, Locale.US)
+        return format.format(date)
+    }
```
wmontwe: This should use [kotlinx-datetime](https://github.com/Kotlin/kotlinx-datetime).

```suggestion
    private fun convertLongToTime(long: Long): String {
        val instant = Instant.fromEpochMilliseconds(long)
        val dateTime = instant.toLocalDateTime(TimeZone.currentSystemDefault())
        return LocalDateTime.Formats.ISO.format(dateTime)
    }
```

asoucar:
@wmontwe I addressed all your comments I think. I am aware there are 2 commented out tests at the moment, but I wanted to get the rest of the code up because I know the tests are failing due to coroutine testing finessing

core/logging/impl-file/src/jvmMain/kotlin/net/thunderbird/core/logging/file/FileLogSink.jvm.kt
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * A [LogSink] implementation that logs messages to a specified internal file.
+ *
+ * This sink uses the platform-specific implementations to handle logging.
+ *
+ * @param level The minimum [LogLevel] for messages to be logged.
+ * @param fileName The [String] fileName to log to
+ */
+
```
wmontwe: This newline could be removed

legacy/core/build.gradle.kts
```diff
@@ -7,11 +7,14 @@ dependencies {
     api(projects.mail.common)
     api(projects.backend.api)
     api(projects.library.htmlCleaner)
+    api(libs.android.material)
```
wmontwe: Is the material dependency needed?

core/logging/impl-file/build.gradle.kts
wmontwe: The test doesn't require mocks anymore, so this could be removed.

core/logging/impl-file/src/androidMain/kotlin/net/thunderbird/core/logging/file/AndroidFileLogSink.kt
wmontwe, line 1: I think there are threading issues that could cause issues with `writeToLogFile`. You could use a Mutex to protect the sensible operations around `accumulatedLogs` and file operations.

The test could be fixed by always using `Timezone.UTC`.

I'm not a huge fan of using a coroutine job. This is hard to test and could cause race conditions. Some ideas to prevent immediate log writes as alternative to using a job:

- size based write, collect 50 log events before writing to file
- time based write, use a timestamp and a write interval to delay writing
- time and size based write 
- use a MutableSharedFlow that buffers a certain amount of logs

wmontwe: For a time based approach you could use `Clock` and `TestClock` to control the time during tests.

gradle/libs.versions.toml
```diff
@@ -171,6 +173,7 @@ androidx-glance-appwidget-preview = { module = "androidx.glance:glance-appwidget
 androidx-glance-preview = { module = "androidx.glance:glance-preview", version.ref = "androidxGlance" }
 androidx-glance-material3 = { module = "androidx.glance:glance-material3", version.ref = "androidxGlanceMaterial3" }
 androidx-lifecycle-common = { module = "androidx.lifecycle:lifecycle-common", version.ref = "androidxLifecycle" }
+androidx-lifecycle-process = { module = "androidx.lifecycle:lifecycle-process" }
```
wmontwe: Please add the `version.ref` here too, similarly to the other declarations.

gradle/libs.versions.toml
```diff
@@ -107,6 +108,7 @@ timber = "5.0.1"
 turbine = "1.2.0"
 xmlpull = "1.0"
 zxing = "3.5.3"
+lifecycleProcess = "2.9.1"
```
wmontwe: This could be removed as there is already a androidxLifecycle version declaration.

gradle/libs.versions.toml
```diff
@@ -274,6 +278,7 @@ timber = { module = "com.jakewharton.timber:timber", version.ref = "timber" }
 turbine = { module = "app.cash.turbine:turbine", version.ref = "turbine" }
 xmlpull = { module = "com.github.cketti:xmlpull-extracted-from-android", version.ref = "xmlpull" }
 zxing = { module = "com.google.zxing:core", version.ref = "zxing" }
+lifecycle-process = { group = "androidx.lifecycle", name = "lifecycle-process", version.ref = "lifecycleProcess" }
```
wmontwe: This could be removed as there is already a androidxLifecycle version declaration.

app-common/build.gradle.kts
```diff
@@ -40,6 +41,8 @@ dependencies {
     implementation(projects.mail.protocols.imap)
 
     implementation(libs.androidx.work.runtime)
+    implementation(libs.androidx.lifecycle.process)
+    implementation(libs.lifecycle.process)
```
wmontwe: This could be removed as there is already a androidx lifecycle dependency.

app-common/src/main/kotlin/net/thunderbird/app/common/feature/LoggerLifeCycleObserver.kt
```diff
@@ -0,0 +1,17 @@
+package net.thunderbird.app.common.feature
+
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
+import net.thunderbird.core.logging.file.FileLogSink
+
+class LoggerLifeCycleObserver(val fileLogSink: FileLogSink?) : DefaultLifecycleObserver {
```
wmontwe: Nit: `LoggerLifeCycleObserver` -> `LoggerLifecycleObserver`

core/logging/impl-file/src/jvmMain/kotlin/net/thunderbird/core/logging/file/JvmFileSystemManager.kt
```diff
@@ -0,0 +1,13 @@
+package net.thunderbird.core.logging.file
+
+import kotlinx.io.RawSink
+
+/**
+ * Android implementation of [FileSystemManager] that uses [ContentResolver] to perform file operations.
+ */
+class JvmFileSystemManager() : FileSystemManager {
+    override fun openSink(uriString: String, mode: String): RawSink? {
+        // TODO: Implementation
```
wmontwe: Could you please add an issue to add the JVM implementation and add the link here.

legacy/core/build.gradle.kts
```diff
@@ -12,6 +12,8 @@ dependencies {
     api(projects.core.android.account)
     api(projects.core.preference.impl)
     api(projects.core.android.logging)
+    api(projects.core.logging.implFile)
+    api(projects.core.logging.implComposite)
```
wmontwe, line 16: Why adding the `implComposite` as `api` and `implementation` dependency?

core/logging/impl-file/src/androidMain/kotlin/net/thunderbird/core/logging/file/AndroidFileLogSink.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    override fun log(event: LogEvent) {
+        try {
+            coroutineScope.launch {
+                mutex.withLock {
+                    accumulatedLogs.add(
+                        "${convertLongToTime(event.timestamp)} priority = ${event.level}, ${event.message}",
+                    )
+                }
+                if (accumulatedLogs.size > 4) {
```
wmontwe: This should be a constant.

core/logging/impl-file/src/androidMain/kotlin/net/thunderbird/core/logging/file/AndroidFileLogSink.kt
```diff
@@ -0,0 +1,153 @@
...
+            sink.close()
+            outputStream.close()
+        }
+    }
+
+    override fun flushAndCloseBuffer() {
+        runBlocking {
+            writeToLogFile()
+        }
+    }
```
wmontwe: Hmm, I suggested `runBlocking`, but I realize now that this would be executed on the main thread, which is not ideal. I think it's better to remove it and change this to a suspend function.

Then the `LoggerLifeCycleObserver` needs to be adjusted as:

```kotlin
class LoggerLifeCycleObserver(val fileLogSink: FileLogSink?) : DefaultLifecycleObserver {
    override fun onStop(owner: LifecycleOwner) {
        super.onStop(owner)
        fileLogSink?.let {
            owner.lifecycleScope.launch {
                withContext(Dispatchers.IO) {
                    it.flushAndCloseBuffer()
                }
            }
        }
    }
}
```

and the `ExceptionHandler`:

```kotlin
class ExceptionHandler(
    private val defaultHandler: Thread.UncaughtExceptionHandler?,
) : Thread.UncaughtExceptionHandler, KoinComponent {
    private val syncDebugFileLogSink: FileLogSink by inject(named("syncDebug"))

    override fun uncaughtException(t: Thread, e: Throwable) {
        Log.e("UncaughtException", e.toString(), e)
        runBlocking {
            syncDebugFileLogSink.flushAndCloseBuffer()
        }
        defaultHandler?.uncaughtException(t, e)
    }
}
```

app-common/src/main/kotlin/net/thunderbird/app/common/feature/LoggerLifeCycleObserver.kt
```diff
@@ -0,0 +1,17 @@
...
+    override fun onStop(owner: LifecycleOwner) {
+        super.onStop(owner)
+        fileLogSink?.flushAndCloseBuffer()
+    }
+
+    override fun onDestroy(owner: LifecycleOwner) {
```
wmontwe: I think `onDestroy` is not needed as `onStop` is called before and for the case the app crashes there is the exception handler.

wmontwe (changes requested):
Thank you, I just added some small adjustments to make it more robust.

legacy/core/build.gradle.kts
asoucar: Doing it as implementation makes the KoinModule not able to find it

wmontwe (approved):
Thank you!

Issue #9229, opened 2025-05-27
Change feature search to Kotlin


wmontwe:
The `feature:search` is still written in Java, this needs to be changed to allow further changes to search to finally decouple `MessageList` from `LocalSearch`. This should help to fix the drawer cashes when the local state doesn't match the app state.

Additionally it should be split into api/implementation and changed to a Kotlin Multiplatform module.

PR #9358, opened 2025-06-20
Refactor search to kotlin - part 2

wmontwe:
These changes are the foundation for the drawer with the unified account.

Part of [#9229](https://github.com/thunderbird/thunderbird-android/issues/9229).

Remaining `LocalMessageSearch` will be converted to Kotlin in a follow-up pull-request.

- `LocalSearch` renamed to `LocalMessageSearch` and `SearchSpecification` to `MessageSearchSpecification` as they are about message related searches
- `ConditionsTreeNode` renamed to `SearchConditionTreeNode` and refactored to a builder pattern for easier use and now tested with new not operation.
- `SearchField` renamed to `MessageSearchField` and introduced new `SearchField` and `SearchFieldType` to abstract from message specifics.
- `SqlQueryBuilder` renamed to `SqlWhereClause` and refactored to a builder pattern and now tested.
- `MessageSearchField` remained an enum to provide Java interop, in the future it could be a sealed class instead


feature/search/src/main/java/net/thunderbird/feature/search/api/MessageSearchSpecification.kt
```diff
@@ -0,0 +1,28 @@
...
+ * This interface is used to encapsulate the details of a search operation,
+ * including which accounts to search and the conditions that must be met
+ * for messages to be included in the search results.
+ */
+interface MessageSearchSpecification : Parcelable {
+    /**
+     * Get all the uuids of accounts this search acts on.
+     * @return Set of uuids.
+     */
+    val accountUuids: Set<String>
```
rafaeltonholo, line 19: nit: I believe we should start using a value class `AccountUuid` for these kind of data. We already have one defined in the codebase, maybe it would be a case to move it to `:feature:account:api`.

feature/search/src/main/java/net/thunderbird/feature/search/sql/SqlWhereClause.kt
```diff
@@ -0,0 +1,165 @@
...
+            selectionArgs.add(selectionArg)
+        }
+    }
+
+    companion object Companion {
+        // TODO: This is a workaround for ambiguous column names in the selection. Find a better solution.
+        fun addPrefixToSelection(columnNames: Array<String>, prefix: String?, selection: String): String {
+            var result = selection
+            for (columnName in columnNames) {
+                result = result.replace(("(?<=^|[^\\.])\\b" + columnName + "\\b").toRegex(), prefix + columnName)
```
rafaeltonholo: You could use string interpolation instead

feature/search/src/main/java/net/thunderbird/feature/search/SearchConditionTreeNode.kt
```diff
@@ -0,0 +1,177 @@
...
+                val rightStr = right?.toString() ?: "null"
+                "($leftStr ${operator.name} $rightStr)"
+            }
+
+            Operator.CONDITION -> condition.toString()
+            Operator.NOT -> "(NOT ${left?.toString() ?: "null"})"
+        }
+    }
+
+    class Builder(
```
rafaeltonholo, line 108: nit: It would be interesting having a DSL builder function to make it more Kotlin idiomatic.

rafaeltonholo (approved):
LGTM. I have left just a few comments. Few free to address them or not

feature/search/src/main/java/net/thunderbird/feature/search/SearchConditionTreeNode.kt
wmontwe: Good idea, but currently this code is mainly used by the remaining Java code base.

feature/search/src/main/java/net/thunderbird/feature/search/api/MessageSearchSpecification.kt
wmontwe: It should use the new `AccountId` class, but that is even more refactoring. Maybe for the next pull-request.

PR #9359, opened 2025-06-20
Logging - Add log auto tagging

rafaeltonholo:
After our recent changes to the Logging API, logs that do not explicitly receive a tag as an argument are all being tagged with `AndroidConsoleLogSink`. 

Additionally, the introduction of a `Log` object, designed to maintain legacy logging behaviour, has led to an issue where nearly all logs are being tagged with `AndroidConsoleLogSink`, making it harder to filter the logs.

This PR reinstates the automatic tagging code from commit 0a7d5964fb1560bf8b6f08dfea008619c8568853, with some modifications to allow both Android and JVM code to utilize the same shared code. This approach ensures that we do not compromise support for other platforms that may be added in the future.

### DIFF
--- a/app-common/build.gradle.kts
+++ b/app-common/build.gradle.kts
@@ -4,6 +4,10 @@ plugins {
 
 android {
     namespace = "net.thunderbird.app.common"
+
+    buildFeatures {
+        buildConfig = true
+    }
 }
 
 dependencies {
--- a/app-common/src/main/kotlin/net/thunderbird/app/common/core/AppCommonCoreModule.kt
+++ b/app-common/src/main/kotlin/net/thunderbird/app/common/core/AppCommonCoreModule.kt
@@ -1,5 +1,6 @@
 package net.thunderbird.app.common.core
 
+import net.thunderbird.app.common.BuildConfig
 import net.thunderbird.core.logging.DefaultLogger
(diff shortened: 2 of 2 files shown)


rafaeltonholo:
> Nice, I was just wondering if this should be part of the `ConsoleLogSink`, or better located in the logger itself as a `LogTagComposer`:
> 
> ```kotlin
> package net.thunderbird.core.logging
> 
> /**
>  * Interface for composing log tags.
>  *
>  * This interface is responsible for determining the tag to use for a log event.
>  * Implementations can extract tags from stack traces or use other strategies.
>  */
> interface LogTagComposer {
>     /**
>      * Composes a tag for the given [LogEvent].
>      *
>      * @param event The [LogEvent] to compose a tag for.
>      * @return The composed tag, or null if no tag could be determined.
>      */
>     fun composeTag(event: LogEvent): String?
> }
> ```
> 
> What do you think?

I considered that option, but it would mean we need to implement the `composeTag` function/interface for every other platform as well in the future. I recall that @toddheasley mentioned tags are not very useful in the iOS ecosystem, so I'm unsure if implementing it outside the JVM environment would yield any benefits. That’s why I chose to go with that implementation.

wmontwe (approved):
LGTM

PR #9360, opened 2025-06-20
feat(drawer): change drawer layout to dropdown ui

### LINKED ISSUE #9189: Dropdown Drawer - change the account selection UI to dropdown list
The account selection should open a dropdown list of accounts instead of the siderail.

Image

Image

wmontwe:
Resolves #9189

Depends on #9358

This changes the drawer ui to a true dropdown ui solution, hidden behind a feature flag.

## Main Features

1. **Unified Folder Support**: 
   - Added a unified account view that aggregates content from multiple accounts
   - Implemented a unified inbox that shows messages from all accounts in one place

2. **Drawer Layout Improvements**:
   - Added support for two drawer UI modes: dropdown drawer and side rail drawer to keep changes minimal when the feature flag is not enabled
   - Improved folder selection and navigation between accounts
   - Enhanced the account list view with better selection states

3. **UI/UX Enhancements**:
   - Fixed selection state for nested folders
   - Improved avatar color selection for dark mode

## Technical Changes

1. **Architecture**:
   - Created a dedicated repository for unified folders (`UnifiedFolderRepository`)
   - Implemented search functionality to support unified inbox views
   - Added proper message counting for unified folders
   - Created new domain entities: `UnifiedDisplayAccount`, `UnifiedDisplayFolder`, and `UnifiedDisplayFolderType`

2. **Bug Fixes**:
   - Fixed issues with folder selection in nested folders
   - Fixed account selection to only trigger when the account is different
   - Removed deselect functionality that was preventing folders from being selected
   - Fixed management of folders when unified account is shown

3. **Feature Flags**:
   - Added feature flags to control the drawer functionality:
     - `enable_dropdown_drawer`: Controls the overall drawer feature (dropdown vs. side rail)
     - `enable_dropdown_drawer_ui`: Controls the UI mode (dropdown with expandable folders vs. side rail with expandable folders)
     
| Name       | Screenshot                                  |
|-----------------------|----------------------------------------------------|
| Unified Account          | Unified Account |
| Account List          | Account List |
| Account View         | Account View |
| Account View with nested selection         | Account View |

### DIFF
--- a/core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/organism/drawer/NavigationDrawerItem.kt
+++ b/core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/organism/drawer/NavigationDrawerItem.kt
@@ -9,6 +9,16 @@ import androidx.compose.ui.text.style.TextOverflow
 import app.k9mail.core.ui.compose.designsystem.atom.text.TextLabelLarge
 import androidx.compose.material3.NavigationDrawerItem as Material3NavigationDrawerItem
 
+/**
+ * A navigation drawer item that can be used in a navigation drawer.
+ *
+ * @param label The content of the label to be displayed in the item as a String.
+ * @param selected Whether this item is selected.
+ * @param onClick The callback to be invoked when this item is clicked.
+ * @param modifier The [Modifier] to be applied to this item.
+ * @param icon An optional composable that represents an icon for this item.
+ * @param badge An optional composable that represents a badge for this item.
+ */
 @Composable
 fun NavigationDrawerItem(
     label: String,
@@ -36,6 +46,16 @@ fun NavigationDrawerItem(
     )
 }
 
+/**
+ * A navigation drawer item that can be used in a navigation drawer.
+ *
+ * @param label The content of the label to be displayed in the item as AnnotatedString.
+ * @param selected Whether this item is selected.
+ * @param onClick The callback to be invoked when this item is clicked.
+ * @param modifier The [Modifier] to be applied to this item.
+ * @param icon An optional composable that represents an icon for this item.
+ * @param badge An optional composable that represents a badge for this item.
+ */
 @Composable
 fun NavigationDrawerItem(
     label: AnnotatedString,
@@ -53,6 +73,37 @@ fun NavigationDrawerItem(
                 maxLines = 2,
             )
         },
+        selected = selected,
+        onClick = onClick,
+        modifier = Modifier
+            .padding(NavigationDrawerItemDefaults.ItemPadding)
+            .then(modifier),
+        icon = icon,
+        badge = badge,
+    )
+}
+
+/**
+ * A navigation drawer item that can be used in a navigation drawer.
+ *
+ * @param label The content of the label to be displayed in the item.
+ * @param selected Whether this item is selected.
+ * @param onClick The callback to be invoked when this item is clicked.
+ * @param modifier The [Modifier] to be applied to this item.
+ * @param icon An optional composable that represents an icon for this item.
+ * @param badge An optional composable that represents a badge for this item.
+ */
+@Composable
+fun NavigationDrawerItem(
+    label: @Composable () -> Unit,
+    selected: Boolean,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    icon: (@Composable () -> Unit)? = null,
+    badge: (@Composable () -> Unit)? = null,
+) {
+    Material3NavigationDrawerItem(
+        label = label,
         selected = selected,
         onClick = onClick,
         modifier = Modifier
--- a/core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/organism/drawer/NavigationDrawerItem.kt
+++ b/core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/organism/drawer/NavigationDrawerItem.kt
@@ -28,7 +28,7 @@ fun NavigationDrawerItem(
     icon: (@Composable () -> Unit)? = null,
     badge: (@Composable () -> Unit)? = null,
 ) {
-    Material3NavigationDrawerItem(
+    NavigationDrawerItem(
         label = {
             TextLabelLarge(
                 text = label,
@@ -38,9 +38,7 @@ fun NavigationDrawerItem(
         },
         selected = selected,
         onClick = onClick,
-        modifier = Modifier
-            .padding(NavigationDrawerItemDefaults.ItemPadding)
-            .then(modifier),
+        modifier = modifier,
         icon = icon,
         badge = badge,
     )
@@ -65,7 +63,7 @@ fun NavigationDrawerItem(
     icon: (@Composable () -> Unit)? = null,
     badge: (@Composable () -> Unit)? = null,
 ) {
-    Material3NavigationDrawerItem(
+    NavigationDrawerItem(
         label = {
             TextLabelLarge(
                 text = label,
@@ -75,9 +73,7 @@ fun NavigationDrawerItem(
         },
         selected = selected,
         onClick = onClick,
-        modifier = Modifier
-            .padding(NavigationDrawerItemDefaults.ItemPadding)
-            .then(modifier),
+        modifier = modifier,
         icon = icon,
         badge = badge,
     )
--- a/core/ui/compose/theme2/common/src/main/kotlin/app/k9mail/core/ui/compose/theme2/SelectThemeColorScheme.kt
+++ b/core/ui/compose/theme2/common/src/main/kotlin/app/k9mail/core/ui/compose/theme2/SelectThemeColorScheme.kt
@@ -1,12 +1,12 @@
 package app.k9mail.core.ui.compose.theme2
 
-import android.annotation.SuppressLint
 import android.content.Context
 import androidx.compose.material3.ColorScheme
 import androidx.compose.material3.dynamicDarkColorScheme
 import androidx.compose.material3.dynamicLightColorScheme
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.compositeOver
 import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.platform.LocalContext
 import com.google.android.material.color.MaterialColors
@@ -115,6 +115,23 @@ private fun ColorScheme.toDynamicThemeColorScheme(
     )
 }
 
+/**
+ * The color roles of a theme accent color. They are used to define the main accent color and its complementary colors
+ * in a Material Design theme.
+ *
+ * These roles are used to create a harmonious color scheme that works well together.
+ *
+ * The roles are:
+ * - `accent`: The main accent color.
+ * - `onAccent`: The color used for text and icons on top of the accent color.
+ * - `accentContainer`: A container color that complements the accent color.
+ * - `onAccentContainer`: The color used for text and icons on top of the accent container color.
+ *
+ * @param accent The main accent color.
+ * @param onAccent The color used for text and icons on top of the accent color.
+ * @param accentContainer A container color that complements the accent color.
+ * @param onAccentContainer The color used for text and icons on top of the accent container color.
+ */
 data class ColorRoles(
     val accent: Color,
     val onAccent: Color,
@@ -122,8 +139,24 @@ data class ColorRoles(
     val onAccentContainer: Color,
 )
 
+/**
+ * Returns a harmonized color that is derived from the given color and the target color.
+ *
+ * This function uses Material Colors to harmonize the two colors.
+ *
+ * @param target The target color to harmonize with.
+ * @return A new color that is harmonized with the target color.
+ */
 fun Color.toHarmonizedColor(target: Color) = Color(MaterialColors.harmonize(toArgb(), target.toArgb()))
 
+/**
+ * Returns a [ColorRoles] object that contains the accent colors derived from the given color.
+ *
+ * This function uses Material Colors to retrieve the accent colors based on the provided color.
+ *
+ * @param context The context to use for retrieving the color roles.
+ * @return A [ColorRoles] object containing the accent colors.
+ */
 fun Color.toColorRoles(context: Context): ColorRoles {
     val colorRoles = MaterialColors.getColorRoles(context, this.toArgb())
     return ColorRoles(
@@ -134,8 +167,16 @@ fun Color.toColorRoles(context: Context): ColorRoles {
     )
 }
 
-@SuppressLint("RestrictedApi")
-fun Color.toSurfaceContainer(context: Context): Color {
-    val surfaceContainer = MaterialColors.getSurfaceContainerFromSeed(context, this.toArgb())
-    return Color(surfaceContainer)
+/**
+ * Returns a surface container color that is a composite of the given color and the theme surface container color.
+ *
+ * The alpha value is applied to the given color before compositing.
+ *
+ * @param alpha The alpha value to apply to the color.
+ * @return A new color that is a composite of the given color and the theme surface container color.
+ */
+@Composable
+fun Color.toSurfaceContainer(alpha: Float): Color {
+    val color = copy(alpha = alpha)
+    return color.compositeOver(MainTheme.colors.surfaceContainer)
 }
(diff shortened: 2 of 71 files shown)


core/ui/compose/designsystem/src/main/kotlin/app/k9mail/core/ui/compose/designsystem/organism/drawer/NavigationDrawerItem.kt
```diff
@@ -62,3 +82,34 @@ fun NavigationDrawerItem(
...
+ *
+ * @param label The content of the label to be displayed in the item.
+ * @param selected Whether this item is selected.
+ * @param onClick The callback to be invoked when this item is clicked.
+ * @param modifier The [Modifier] to be applied to this item.
+ * @param icon An optional composable that represents an icon for this item.
+ * @param badge An optional composable that represents a badge for this item.
+ */
+@Composable
+fun NavigationDrawerItem(
```
rafaeltonholo, line 93: With the introduction of this Composable, the previous two must be using this instead of `Material3NavigationDrawerItem`

core/ui/compose/theme2/common/src/main/kotlin/app/k9mail/core/ui/compose/theme2/SelectThemeColorScheme.kt
```diff
@@ -132,3 +133,9 @@ fun Color.toColorRoles(context: Context): ColorRoles {
         onAccentContainer = Color(colorRoles.onAccentContainer),
     )
 }
+
+@SuppressLint("RestrictedApi")
+fun Color.toSurfaceContainer(context: Context): Color {
+    val surfaceContainer = MaterialColors.getSurfaceContainerFromSeed(context, this.toArgb())
+    return Color(surfaceContainer)
+}
```
rafaeltonholo: I wouldn't use a restricted api as the library can remove it without further notice. I would prefer copying the function's contents rather than using it directly.

Additionally, you might be able to achieve the same effect you want by using the conjunction of `color.copy(alpha)` and `compositeOver(MainTheme.colors.surfaceContainer)`

feature/account/avatar/impl/src/main/kotlin/net/thunderbird/feature/account/avatar/ui/AvatarOutlined.kt
```diff
@@ -51,19 +51,19 @@ fun AvatarOutlined(
...
     color: Color,
-    colorRoles: ColorRoles,
+    borderColor: Color,
     modifier: Modifier = Modifier,
     onClick: (() -> Unit)? = null,
     content: @Composable BoxScope.() -> Unit,
 ) {
     Surface(
-        color = colorRoles.accentContainer,
+        color = color,
```
rafaeltonholo, line 59: There is a small distortion in the avatar's border that we can fix by applying the `CircleShape` to the `Surface`:

image

You can also remove the `.clip(CircleShape)` after changing the surface's shape.

```suggestion
        color = color,
        shape = CircleShape,
```

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/account/AccountAvatar.kt
```diff
@@ -24,7 +27,9 @@ internal fun AccountAvatar(
     onClick: ((DisplayAccount) -> Unit)? = null,
 ) {
     val context = LocalContext.current
-    val accountColor = calculateAccountColor(account.color)
+    val name = getDisplayAccountName(account)
+    val color = getDisplayAccountColor(account)
+    val accountColor = calculateAccountColor(color.toArgb())
```
rafaeltonholo: I would encapsulate this within a remember function. The `calculateAccountColor` function performs color calculations that it would be good to avoid recalculating whenever a recomposition occurs.

```suggestion
    val accountColor = remember(color) {
        calculateAccountColor(color.toArgb())
    }
```

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/domain/entity/DisplayAccount.kt
```diff
@@ -1,10 +1,7 @@
...
 
-internal data class DisplayAccount(
-    val id: String,
-    val name: String,
-    val email: String,
-    val color: Int,
-    val unreadMessageCount: Int,
-    val starredMessageCount: Int,
-)
+interface DisplayAccount {
```
rafaeltonholo: I would use a `sealed interface` here

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/account/AccountListItem.kt
```diff
@@ -1,30 +1,82 @@
...
-        contentAlignment = Alignment.Center,
+    val color = getDisplayAccountColor(account)
+    val name = getDisplayAccountName(account)
+
+    NavigationDrawerItem(
+        label = { AccountLabel(account = account) },
+        selected = selected,
+        onClick = { onClick(account) },
+        modifier = modifier.fillMaxWidth()
+            .height(MainTheme.sizes.large),
```
rafaeltonholo, line 40: Do we need to fix the height on `64.dp`? Additionally, I would reduce it if possible. The component is slightly big on small screens. 

image

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/account/AccountListItem.kt
```diff
@@ -1,30 +1,82 @@
...
+        icon = {
+            AvatarOutlined(
+                color = color,
+                name = name,
+                size = AvatarSize.MEDIUM,
```
rafaeltonholo, line 45: The current `AvatarSize.MEDIUM` is resolving in `48.dp`, which is slightly bigger then the proposed by the design `40.dp`. Having extra `8.dp` on small screens can compromise layout space.

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/account/AccountListItemBadge.kt
```diff
@@ -0,0 +1,64 @@
...
+}
+
+@Composable
+private fun AccountCountAndStarredBadge(
+    unreadCount: Int,
+    starredCount: Int,
+    showStarredCount: Boolean,
+    modifier: Modifier = Modifier,
+) {
+    Row(
```
rafaeltonholo: I couldn't see that in the designs, but I would consider using a `Column` instead of a `Row` in this case. On small screens, the Account title is getting squished because of not enough room to display:

image

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/setting/AccountSettingList.kt
```diff
@@ -0,0 +1,29 @@
...
+    Column(
+        modifier = modifier
+            .padding(vertical = MainTheme.spacings.default)
+            .fillMaxWidth(),
+    ) {
+        SettingListItem(
+            label = stringResource(id = R.string.navigation_drawer_dropdown_action_sync_all_accounts),
+            onClick = onSyncAllAccountsClick,
+            icon = Icons.Outlined.Sync,
+        )
```
rafaeltonholo: nit: I would consider adding another option: "Add new account"

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/siderail/ui/account/SideRailAccountIndicator.kt
```diff
@@ -18,7 +21,7 @@ internal fun AccountIndicator(
             .defaultMinSize(
                 minHeight = MainTheme.spacings.default,
             ),
-        color = calculateAccountColor(accountColor),
+        color = calculateAccountColor(accountColor.toArgb()),
```
rafaeltonholo: I would encapsulate this within a remember function. The `calculateAccountColor` function performs color calculations that it would be good to avoid recalculating whenever a recomposition occurs.

```suggestion
        color = remember(accountColor) {
            calculateAccountColor(accountColor.toArgb())
        },
```

rafaeltonholo (commented):
Added a first round of review. I will come back later to complete the review

core/ui/compose/theme2/common/src/main/kotlin/app/k9mail/core/ui/compose/theme2/SelectThemeColorScheme.kt
wmontwe: I used an alpha value previously, but that caused issues due to the brackground shining through. I'll check out `compositeOver`.

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/account/AccountListItem.kt
wmontwe: So far the layout is not optimized for landscape at all. Have to check what we could do there.

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/account/AccountListItem.kt
wmontwe: I applied the minimum touch target size of 48.dp here, as the avatar is also used as a touch item in the siderail.

I'm not 100% following the design as it still has a couple of rough edges, especially around responsive layouts and the design team was occupied with other tasks. The prototype linked in the design is more accurate as it was updated.

The design is currently not optimized at all for small screens. The new Bolt design system takes this into account and we will add compact, default and relaxed as general spacing guides and will introduce breakpoints to slightly adjust spacings and sizes when within a certain Window Size Class. Despite these changes, we might need to add a special case for very small screens.

rafaeltonholo: I see, thanks for explaining that

feature/navigation/drawer/siderail/src/main/kotlin/net/thunderbird/feature/navigation/drawer/siderail/SideRailDrawer.kt
```diff
@@ -79,6 +79,7 @@ class SideRailDrawer(
     override fun selectUnifiedInbox() {
         drawerState.update {
             it.copy(
+                selectedAccountUuid = "unified_account",
```
rafaeltonholo, line 82: As this is being used in multiple places, I believe this could be moved to a constant

rafaeltonholo (changes requested):
Great work on the new Drawer implementation! I'm excited to see this new version live soon!

I have just left a few comments that need to be addressed before merging. Let me know you have any questions

rafaeltonholo:
I would like to give two suggestions that could be included in this PR or in a different PR if you feel it would make sense to add that:

1. Add an option to "Add new account" in the Settings options when the account list is visible. That will improve the user experience, giving a fast way to add a new account
2. Enable account switching by swiping down the `AccountView` when it is displaying a selected account.

These features would enhance the user experience for me, but I understand they may not be necessary for everyone. Feel free to disregard them if you think they are not a suitable use case.

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/account/AccountListItemBadge.kt
wmontwe: I think using a column is a very good idea 👍

feature/account/avatar/impl/src/main/kotlin/net/thunderbird/feature/account/avatar/ui/AvatarOutlined.kt
wmontwe: Good catch!

feature/navigation/drawer/siderail/src/main/kotlin/net/thunderbird/feature/navigation/drawer/siderail/SideRailDrawer.kt
wmontwe: This is the old drawer implementation that doesn't know the new `UnifiedDisplayAccount`. This needs to be reworked as the accounts will need to use a true UUID. But I would rather switch to a specific account type rather to keep these hard coded ids around. Could be a specific task.

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/siderail/ui/account/SideRailAccountIndicator.kt
wmontwe: I removed the `toArgb()` calculation. But `remember` would require more changes, as it's not a Composable context.

rafaeltonholo: I believe you just need a Composable context because of the fallback color when `accountColor` is 0 and the surface color, is that correct?

Would you be able to pass that as a Color parameter to `calculateAccountColor` instead? If you prefer, you could also create something like:
```kotlin
@Composable
fun rememberCalculatedAccountColor(accountColor: Int, fallbackColor: Color = MainTheme.colors.primary): Color {
    val surface = MainTheme.colors.surface
    return remember {
        if (accountColor == 0) {
            fallbackColor
        } else {
            Color(accountColor).toHarmonizedColor(surface)
        }
    }
}
```

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/account/AccountListItem.kt
wmontwe: I reworked the settings and show them in a row when there is not enough space

feature/navigation/drawer/siderail/src/main/kotlin/net/thunderbird/feature/navigation/drawer/siderail/SideRailDrawer.kt
wmontwe: But I replaced it in the `MessageList`

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/setting/AccountSettingList.kt
wmontwe: Added 👍

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/siderail/ui/account/SideRailAccountIndicator.kt
wmontwe: I think that works, thanks

wmontwe:
> I would like to give two suggestions that could be included in this PR or in a different PR if you feel it would make sense to add that:
> 
>     1. Add an option to "Add new account" in the Settings options when the account list is visible. That will improve the user experience, giving a fast way to add a new account
> 
>     2. Enable account switching by swiping down the `AccountView` when it is displaying a selected account.
> 
> 
> These features would enhance the user experience for me, but I understand they may not be necessary for everyone. Feel free to disregard them if you think they are not a suitable use case.

I updated the pull-request and added your first suggestion. The 2nd is a good idea, but out of scope for these changes. Maybe as separate task.

PR #9378, opened 2025-06-24
Fix issue 8376 - Scrolling in a short email may trigger left/right swipe

### LINKED ISSUE #8376: Scrolling in a short email may trigger left/right swipe
### Checklist

### App

Thunderbird for Android

### App version

14.26.6

Google Play

### Android version

Android 14 / REDMAGICOS9.0.20_NX769J_GB

### Device model

REDMAGIC 9 Pro

### Steps to reproduce

1. Open an email that is short enough to fit in whole on the screen
2. Scroll up/down quickly

### Expected behavior

Nothing happens

### Actual behavior

If the scrolling action wasn't perfectly vertical, it will mistake the gesture for a left/right swipe and switch to the previous/next email.

If improving the gesture detection isn't an option, an easy fix would be to add an option to the settings to disable the left/right gesture within emails.

### Logs

_No response_

joohnq:
Fixes #8376 

I used the fling listener to reduce the sensitivity on the X axis. This way, it's less likely that the user will accidentally swipe to another email while scrolling.

Now, I was thinking of creating an option that the user can enable or disable. What do you think?

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessageViewContainerFragment.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessageViewContainerFragment.kt
@@ -7,21 +7,24 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.RecyclerView
 import androidx.viewpager2.adapter.FragmentStateAdapter
 import androidx.viewpager2.widget.MarginPageTransformer
 import androidx.viewpager2.widget.ViewPager2
 import app.k9mail.legacy.message.controller.MessageReference
+import com.fsck.k9.K9
 import com.fsck.k9.ui.R
 import com.fsck.k9.ui.base.extensions.withArguments
 import com.fsck.k9.ui.messagelist.MessageListItem
 import com.fsck.k9.ui.messagelist.MessageListViewModel
+import net.thunderbird.core.preference.GeneralSettingsManager
 
 /**
  * A fragment that uses [ViewPager2] to allow the user to swipe between messages.
  *
  * Individual messages are displayed using a [MessageViewFragment].
  */
-class MessageViewContainerFragment : Fragment() {
+class MessageViewContainerFragment: Fragment() {
     var isActive: Boolean = false
         set(value) {
             field = value
@@ -29,6 +32,8 @@ class MessageViewContainerFragment : Fragment() {
         }
 
     private var showAccountChip: Boolean = true
+
+    private var isUseLeftRightGestureNavigation: Boolean = true
 
     lateinit var messageReference: MessageReference
         private set
@@ -68,6 +73,8 @@ class MessageViewContainerFragment : Fragment() {
 
         showAccountChip = arguments?.getBoolean(ARG_SHOW_ACCOUNT_CHIP) ?: showAccountChip
 
+        isUseLeftRightGestureNavigation = arguments?.getBoolean(ARG_IS_USE_LEFT_RIGHT_GESTURE_NAVIGATION) ?: isUseLeftRightGestureNavigation
+
         adapter = MessageViewContainerAdapter(this, showAccountChip)
     }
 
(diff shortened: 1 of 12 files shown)


joohnq:
I’d like to know if there’s a global object or class in the codebase that stores constants or configuration values. For example, I added a value that multiplies the X velocity, and I want to know if this kind of value should stay hardcoded or be placed in a shared/configurable location.

kewisch:
I think we should have a sensible default here instead of adding an option. I'll leave the code review and remaining questions to Wolf. I'm going to add the magic note to your original comment so the issue will close automatically.

joohnq:
Thanks! At first, I considered letting the user choose a value for this, but later I realized it might not be necessary, and a sensible default should be enough.
Just to clarify one last point for whoever reviews the code: is there already an object or a preferred place in the codebase to store values like this?
Something like:
```
object Variables {
    const val SENSITIVITY_VIEWPAGER_X_MULTIPLIER = 0.5f
}
```
Or should I just keep it inline?

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessageViewContainerFragment.kt
```diff
@@ -109,6 +115,19 @@ class MessageViewContainerFragment : Fragment() {
...
+            rv.onFlingListener = object : RecyclerView.OnFlingListener() {
+                override fun onFling(velocityX: Int, velocityY: Int): Boolean {
+                    val scaledVelocity = (velocityX * VIEW_PAGER_FLING_VELOCITY_SCALE).toInt()
+                    val adjustedVelocityX = scaledVelocity.coerceIn(VIEW_PAGER_MIN_FLING_VELOCITY_X, VIEW_PAGER_MAX_FLING_VELOCITY_X)
+
+                    return rv.fling(adjustedVelocityX, velocityY)
+                }
+            }
+        }
+
```
wmontwe: I think a `ItemTouchHelper` gives more control of the overall behavior:

```suggestion
        recyclerView?.let { rv ->
            val itemTouchHelper = ItemTouchHelper(object : ItemTouchHelper.SimpleCallback(0, ItemTouchHelper.LEFT or ItemTouchHelper.RIGHT) {
                    override fun onMove(
                        recyclerView: RecyclerView,
                        viewHolder: RecyclerView.ViewHolder,
                        target: RecyclerView.ViewHolder,
                    ): Boolean {
                        return false
                    }

                    override fun onSwiped(viewHolder: RecyclerView.ViewHolder, direction: Int) {
                        if (direction == ItemTouchHelper.LEFT) {
                            viewPager.currentItem = viewPager.currentItem + 1
                        } else if (direction == ItemTouchHelper.RIGHT) {
                            viewPager.currentItem = viewPager.currentItem - 1
                        }
                    }

                    override fun getSwipeThreshold(viewHolder: RecyclerView.ViewHolder): Float {
                        return VIEW_PAGER_SWIPE_THRESHOLD
                    }

                    override fun getSwipeVelocityThreshold(defaultValue: Float): Float {
                        return VIEW_PAGER_SWIPE_VELOCITY_THRESHOLD
                    }
                })

            itemTouchHelper.attachToRecyclerView(rv)
        }
        
        ...
        
        companion object {
        private const val VIEW_PAGER_SWIPE_THRESHOLD = 0.4f
        private const val VIEW_PAGER_SWIPE_VELOCITY_THRESHOLD = 0.8f
        
        ...

```

joohnq: But... why are you suggesting the use of ItemTouchHelper? Isn't it typically used for side actions like delete or mark as read, rather than for controlling page navigation like in ViewPager2?

wmontwe: Unfortunately, `ViewPager2` does not provide an official API for customizing swipe behavior. As a result, many workarounds rely on reflection to modify internal parameters controlling swipe gestures, which may not be the most reliable or maintainable approach.

A more viable alternative is creating a custom implementation of `CustomViewPager2` by overriding the `onTouchEvent` method, allowing for adjustments to the coordinate transformations to reduce sensitivity.

But as `ViewPager2` wraps a `RecyclerView`, using the built-in features of the `ItemTouchHelper` can simplify managing swipe behavior. This helper provides robust event handling that can be used to react to specific touch events only (e.g., swipe vs. move, direction left or right).

However, I've also explored using the `OnFlingListener` approach with varying values, but unfortunately, these efforts have shown either unsatisfactory results or caused application crashes.

joohnq:
All right, based on your suggestion, I’ve updated the code and split this feature into a separate function. If the function structure doesn’t look good to you, feel free to let me know.

Also, I didn’t remove the settings toggle that enables or disables this feature. If you think it can really be discarded, just let me know.

wmontwe:
Thanks for the update, as mentioned in the issue comments already, we would not like to introduce another option to the settings.

wmontwe (approved):
Thank you!

idl0r:
Is this already part of a beta by chance? If so, I'm on 12.0b7 and it doesn't seems to be (much) better, if at all.

Issue #9194, opened 2025-05-20
Authentication fails after some time with a Fastmail account


Elbullazul:
### Checklist

- [x] I have used the search function to see if someone else has already submitted the same bug report.
- [x] I will describe the problem with as much detail as possible.

### App

Thunderbird for Android

### App version

11b02

### Where did you get the app from?

F-Droid

### Android version

Calyx OS 6.7.20

### Device model

Google Pixel 6 Pro

### Steps to reproduce

1. Add a [Fastmail](https://www.fastmail.com) account. The login process will use OAuth to sign me in.
2. Enable notifications and push sync for the "Inbox" folder. Allow the app to set alarms and reminders as well.
3. Allow unrestricted power usage for Thunderbird Beta.
4. After a few hours, a notification with "Authentication failed" will appear, and the app won't be able to sync from the upstream IMAP server.

### Expected behavior

The expected behaviour, and what still happens in Thunderbird 10.0, is that Thunderbird can fetch updates in the background and the OAuth authorization doesn't expire.

### Actual behavior

The OAuth authorization expires a lot faster than before, and the app will continuously prompt me to log in, work for a few hours, and prompt me to log in again.

### Logs

[thunderbird.log](https://github.com/user-attachments/files/20353760/thunderbird.log)

The displayed notification is this:
![Image](https://github.com/user-attachments/assets/11f94d8b-fd94-4600-96c4-0802e94c6cbb)

plett:
I came here to report the same issue. I've been using OAuth with Fastmail since the feature launched (in 8.0?) and it's always worked fine. I've been using 10b1 successfully, but it breaks after a few hours when upgrading to 11b1 or 11b2. Re-authing works fine and I can force close the app and re-open it, or reboot the phone and it still works, but it stops working again after a few more hours. Downgrading back to 10b1 fixes the issue.

Installed from F-Droid, Pixel 7a with Android 15

![Image](https://github.com/user-attachments/assets/dd06a2e2-a65f-4605-b0aa-285edbf124b5)

kewisch:
Confirming based on multiple reports. We'll take a look at the logs soon.

kewisch:
For folks experiencing this on 11, can you maybe downgrade to 10 and confirm it stops happening?

Elbullazul:
I can confirm, I've been using v10 for a month and I haven't gotten this error at all.

neff:
[...]

PR #9390, opened 2025-06-26
Revert "PR #9089 - Fixes Outlook outgoing server autoconfig failure"

rafaeltonholo:
Reverts #9089 PR changes that caused the #9194 issue as a side effect.

I will come up with a different solution for #8024 that won't cause a regression.


wmontwe (approved):
LGTM

Issue #9366, opened 2025-06-21
scrolling only works in the center content area on Welcome & New-account screens (Landscape orientation)


mhf1998:
### Checklist

- [x] I have used the search function to see if someone else has already submitted the same bug report.
- [x] I will describe the problem with as much detail as possible.

### App

K-9 Mail

### App version

K9MAIL_10_0

### Where did you get the app from?

Other

### Android version

10

### Device model

SAMSUNG A6

### Steps to reproduce

1. Launch K9-Mail
2. On the Welcome screen (or tap “Get started” to go to the account-setup flow), rotate the device to landscape orientation.
3. Try to swipe up or down anywhere on the screen to scroll.

### Expected behavior

The entire screen surface (including the blank left/right margins) should respond to drag gestures and scroll the content up or down,

### Actual behavior

Only the central content panel where text and form fields appear accepts scroll gestures.

Swiping on the empty areas at the left or right edges of the screen does not scroll; the gesture is ignored.

### Additional information

- On taller-screen devices, the empty left/right margins become much larger, so an even bigger area is non-scrollable.  
- On shorter-screen devices, those margins shrink, but the scroll-dead zone still remains outside the central panel.  
- In all cases, only the central content area accepts vertical drag gestures; swiping on any blank margin (regardless of its size) does nothing.

kewisch:
Could you maybe create a screen share? I want to make sure I understand correctly.

mhf1998:
https://github.com/user-attachments/assets/ddeb64cd-0d03-4447-a0c3-b8132bd090bc

 This video was recorded on a Samsung A32 running Android 13.

mhf1998:
From a technical perspective, the bug is related to the ResponsiveWidthContainer composable, which limits the width of the content in those screens. Then, when we try to set scrollable content inside it from higher-level classes, the scroll behavior only applies to the content area, and the side padding area becomes non-scrollable.

kewisch:
Thanks, I see what you mean. Is this issue something you could look into? If you say this is also on the welcome screen, I know there were some changes recently to make the logo on the welcome screen resize well when changing orientation and on different screen sizes, we should make sure that continues to work.  #8458

wmontwe:
[...]

PR #9396, opened 2025-06-27
Add proxy scroll to AccountAutoDiscoveryContent

terofeev:
Fixes https://github.com/thunderbird/thunderbird-android/issues/9366

**Details:**
One way to enable scrolling outside the actual scrolling column is to capture vertical drag gestures and manually scroll the inner content.

1. First, we catch vertical drag events and call `scrollBy()` with the drag amount. However, the content stops immediately after lifting the finger, without inertia.
2. Next, we use a decay animation to simulate natural scrolling. `VelocityTracker` calculates the initial movement speed. However, if the user tries to scroll the inner content while the manual scroll is still in progress, a glitch occurs due to two simultaneous scroll events.
3. Finally, we use `NestedScrollConnection` to cancel the simulated natural scrolling.

**Test:**
We can highlight two areas for testing:
- The outer box
- The inner (scrollable) column

by using different background colors via the `background()` method.

**Sample video:**
<details><summary>Demo</summary>
<p>

https://github.com/user-attachments/assets/551d2307-889b-4fc1-8f63-602f76e05cd3

</p>
</details>

### DIFF
--- a/core/ui/compose/common/src/main/kotlin/app/k9mail/core/ui/compose/common/scroll/ScrollModifier.kt
+++ b/core/ui/compose/common/src/main/kotlin/app/k9mail/core/ui/compose/common/scroll/ScrollModifier.kt
@@ -1,0 +1,56 @@
+package app.k9mail.core.ui.compose.common.scroll
+
+import androidx.compose.animation.core.Animatable
+import androidx.compose.animation.splineBasedDecay
+import androidx.compose.foundation.ScrollState
+import androidx.compose.foundation.gestures.detectVerticalDragGestures
+import androidx.compose.foundation.gestures.scrollBy
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.input.pointer.pointerInput
+import androidx.compose.ui.input.pointer.util.VelocityTracker
+import androidx.compose.ui.input.pointer.util.addPointerInputChange
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+
+fun Modifier.proxyScroll(
+    coroutineScope: CoroutineScope,
+    scrollState: ScrollState,
+    onCancelFling: () -> Unit,
+    onCreateFling: (Job) -> Unit,
+): Modifier {
+    val velocityTracker = VelocityTracker()
+    return pointerInput(Unit) {
+        val decay = splineBasedDecay<Float>(this)
+        detectVerticalDragGestures(
+            onDragStart = {
+                velocityTracker.resetTracking()
+                onCancelFling()
+            },
+            onDragEnd = {
+                onCancelFling()
+                val velocity = -velocityTracker.calculateVelocity().y
+                velocityTracker.resetTracking()
+                onCreateFling(
+                    coroutineScope.launch {
+                        var previous = 0f
+                        val animatable = Animatable(0f)
+                        animatable.animateDecay(
+                            initialVelocity = velocity,
+                            animationSpec = decay,
+                        ) {
+                            val delta = value - previous
+                            previous = value
+                            coroutineScope.launch { scrollState.scrollBy(delta) }
+                        }
+                    },
+                )
+            },
+            onVerticalDrag = { change, dragAmount ->
+                coroutineScope.launch { scrollState.scrollBy(-dragAmount) }
+                velocityTracker.addPointerInputChange(change)
+                change.consume()
+            },
+        )
+    }
+}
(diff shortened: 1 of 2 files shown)


feature/account/setup/src/main/kotlin/app/k9mail/feature/account/setup/ui/autodiscovery/AccountAutoDiscoveryContent.kt
```diff
@@ -42,10 +53,28 @@ internal fun AccountAutoDiscoveryContent(
...
+        flingJob = null
+    }
+    val nestedScrollConnection = remember {
+        object : NestedScrollConnection {
+            override fun onPreScroll(available: Offset, source: NestedScrollSource): Offset {
+                cancelFling()
+                return Offset.Zero
+            }
+        }
+    }
```
rafaeltonholo, line 69: This should be part of the `proxyScroll` modifier.

core/ui/compose/common/src/main/kotlin/app/k9mail/core/ui/compose/common/scroll/ScrollModifier.kt
```diff
@@ -0,0 +1,56 @@
...
+                )
+            },
+            onVerticalDrag = { change, dragAmount ->
+                coroutineScope.launch { scrollState.scrollBy(-dragAmount) }
+                velocityTracker.addPointerInputChange(change)
+                change.consume()
+            },
+        )
+    }
+}
```
rafaeltonholo, line 56: If we intend to use the `proxyScroll` modifier to resolve this issue, it should also incorporate the NestedScrollConnection logic; otherwise, the solution will be incomplete.

You can make it a `@Composable` modifier:
```suggestion
@Composable
fun Modifier.proxyScroll(
    scrollState: ScrollState,
): Modifier {
    val coroutineScope = rememberCoroutineScope()
    var flingJob by remember { mutableStateOf<Job?>(null) }
    val nestedScrollConnection = remember {
        object : NestedScrollConnection {
            override fun onPreScroll(available: Offset, source: NestedScrollSource): Offset {
                flingJob?.cancel()
                flingJob = null
                return Offset.Zero
            }
        }
    }

    DisposableEffect(scrollState) {
        onDispose {
            flingJob?.cancel()
            flingJob = null
        }
    }

    val velocityTracker = remember(scrollState) { VelocityTracker() }
    return this then Modifier
        .nestedScroll(nestedScrollConnection)
        .pointerInput(scrollState) {
            val decay = splineBasedDecay<Float>(this)
            detectVerticalDragGestures(
                onDragStart = {
                    velocityTracker.resetTracking()
                    flingJob?.cancel()
                    flingJob = null
                },
                onDragEnd = {
                    flingJob?.cancel()
                    val velocity = -velocityTracker.calculateVelocity().y
                    velocityTracker.resetTracking()
                    flingJob = coroutineScope.launch {
                        var previous = 0f
                        val animatable = Animatable(0f)
                        animatable.animateDecay(
                            initialVelocity = velocity,
                            animationSpec = decay,
                        ) {
                            val delta = value - previous
                            previous = value
                            coroutineScope.launch { scrollState.scrollBy(delta) }
                        }
                    }
                },
                onVerticalDrag = { change, dragAmount ->
                    coroutineScope.launch { scrollState.scrollBy(-dragAmount) }
                    velocityTracker.addPointerInputChange(change)
                    change.consume()
                },
            )
        }
}
```

I would also consider making it by using the `Modifier.Node` API for better performance.

rafaeltonholo (changes requested):
Hi @terofeev,

Thank you for your contribution to addressing issue #9366. However, the current change will not resolve the issue, as the fix is only applied to one of the screens affected. The root of the problem, as noted by the OP, lies with the `ResponsiveWidthContainer` Composable. Consequently, any screen wrapped by it and using vertical scrolling will experience the same problem.

Additionally, I have a few concerns about this proposed solution:

1. Creating a new `Modifier` that requires a `CoroutineScope`, a `NestedScrollConnection`, and a `PointerInput` modifier seems a bit overkill. I would like to explore other options first before adding this modifier.
2. In case we opt to add a new `Modifier` to fix this issue, I strongly recommend using the [`Modifier.Node` API](https://developer.android.com/develop/ui/compose/custom-modifiers#implement-custom) instead.
3. I am unclear about the rationale behind the `ResponsiveWidthContainer` (maybe @wmontwe can give me more context around it). If it is indeed necessary, I would suggest exploring an alternative solution, such as using a `PaddingValue`. This could be exposed to the child composable in a manner similar to how `Scaffold` operates, allowing the child composable to decide how to adjust its size instead of imposing a specific width.
4. Lastly, the fix must be applied to all composables that use `ResponsiveWidthContainer` and have a `Modifier.verticalScroll()`, otherwise the issue won't be totally fixed.

Thank you for your understanding.

rafaeltonholo:
I'm closing this PR in favor of #9403. Please, also check @wmontwe's comment regarding the `ResponsiveWidthContainer` on #9366.

PR #9401, opened 2025-06-28
fix(test): do not use loopback IP for IMAP connectivity test

yantarou:
When running a local IMAP server on the development host, the `open() with connection error should throw` IMAP test will succeed connecting to the IMAP server, thus making the test fail.

This patch changes the use of a loopback IP address (127.1.2.3) to a non-routable RFC-5737 TEST-NET-1 address (192.0.2.123).

Fixes: https://github.com/thunderbird/thunderbird-android/issues/9400


wmontwe (approved):
Thank you!

PR #9443, opened 2025-07-11
chore(i18n): Fix release i18n placeholder and pluralization issues

kewisch:
This should fix the localization errors on release:

* I added many where other was missing and vice versa
* I fixed a few xliff errors in Tamil

### DIFF
--- a/legacy/ui/legacy/src/main/res/values-ta-rIN/strings.xml
+++ b/legacy/ui/legacy/src/main/res/values-ta-rIN/strings.xml
@@ -258,8 +258,8 @@
     <string name="idle_refresh_period_2min">ஒவ்வொரு 2 நிமிடங்களுக்கும்</string>
     <string name="account_setup_options_mail_display_count_250">250 செய்திகள்</string>
     <plurals name="notification_new_messages_title">
-        <item quantity="one">&lt;xliff: g ஐடி = \"new_message_count\"&gt;%d &lt;/xliff: g&gt; புதிய செய்தி</item>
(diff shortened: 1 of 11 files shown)


legacy/ui/legacy/src/main/res/values-ta-rIN/strings.xml
```diff
@@ -371,7 +371,7 @@
     <string name="account_notify_contacts_mail_only_summary">அறியப்பட்ட தொடர்புகளின் செய்திகளுக்கு மட்டுமே அறிவிப்புகளைக் காட்டுங்கள்</string>
     <string name="account_settings_ignore_chat_messages_summary">மின்னஞ்சல் அரட்டையைச் சேர்ந்த செய்திகளுக்கான அறிவிப்புகளைக் காட்ட வேண்டாம்</string>
     <string name="empty_subject">பொருள் காலியாக உள்ளது, எப்படியும் அனுப்ப மீண்டும் சொடுக்கு செய்க</string>
-    <string name="message_view_recipients_format">பெறுநர் &lt;xliff: g ஐடி = \"பெறுநர்கள்\"&gt;%s &lt;/xliff: g&gt;</string>
+    <string name="message_view_recipients_format">பெறுநர் <xliff:g id="பெறுநர்கள்">%s</xliff:g></string>
```
rafaeltonholo: Although I believe the id doesn't matter much, the correct id here should be `recipients`:
```suggestion
    <string name="message_view_recipients_format">பெறுநர் <xliff:g id="recipients">%s</xliff:g></string>
```

PR #9448, opened 2025-07-13
Split GeneralSettings and GeneralSettingsManager in smaller classes/interfaces

### LINKED ISSUE #9432: Split `GeneralSettings` and `GeneralSettingsManager` in smaller classes/interfaces
During the replacement of the direct use of K9 object, we ended up creating a settings manager that defeats the single responsibility principle, risking the creation of another K9 class.

We should step back and change the approach we are taking so far before it is too hard to maintain. Here are a few steps we need to take:

- [ ] Analyze each property inside `GeneralSettings` and verify if it belongs to a group of settings. We could group each of the properties by respecting the grouping we already have in the Settings screen (e.g., Display, Interaction, Notification, etc), and/or by areas (e.g., Account list, Message List, Message Composer, etc.)
- [ ] Split the `GeneralSettingsManager` interface/implementation into smaller interfaces, each one handling the group they are responsible for. The `GeneralSettingsManager` must still implement all other `SettingsPreferenceManager` interfaces.
- [ ] Refactor the `RealGeneralSettingsManager` implementation:
    - [ ] Rename it from `RealGeneralSettingsManager` to `DefaultGeneralSettingsManager`, following our code naming convention
    - [ ] Implement either `GeneralSettingsManager` and the other `SettingsPreferenceManager`s
    - [ ] Receive all the other `SettingsPreferenceManager` interfaces as a composition, so `RealGeneralSettingsManager` can fulfill all the interface requirements without requiring to implementation of their methods by using the [Delegation pattern](https://kotlinlang.org/docs/delegation.html)

Notes: 
- Prefer using `PreferenceManager<T>` instead of creating a completely new interface for Preference management
- Avoid creating a `set` method for each Settings property. Instead, you could use an extension function such as:
   ```kotlin
    inline fun <reified T> PreferenceManager<T>.update(updater: (T) -> T) {
        val config = getConfig()
        save(updater(config))
    }
   ```
   And directly calling the `PreferenceManager<T>.update()` to update the settings value, as exemplified next:
   ```kotlin
    class GeneralSettingsDataStore(
        ...
        private val privacySettingsPreference: PreferenceManager<PrivacySettings>,
    ) {
        override fun putBoolean(key: String, value: Boolean) {
            when (key) {
                ....
                "privacy_hide_timezone" -> {
                    skipSaveSettings = true
                    privacySettingsPreference.update { it.copy(isHideTimeZone = value) }
                }
            }
        }
    }
   ```

shamim-emon:
- Fixes #9432


legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
```diff
@@ -0,0 +1,151 @@
...
+ * Retrieve and modify general settings.
+ *
+ * Currently general settings are split between [K9] and [GeneralSettings]. The goal is to move everything over to
+ * [GeneralSettings] and get rid of [K9].
+ *
+ * The [GeneralSettings] instance managed by this class is updated with state from [K9] when [K9.saveSettingsAsync] is
+ * called.
+ */
+// TODO(#9432): Split GeneralSettings and GeneralSettingsManager in smaller classes/interfaces
+@Suppress("TooManyFunctions")
```
rafaeltonholo: You can remove both the TODO comment and the suppress, since it was both addressed in this PR.
```suggestion
```

config/detekt/detekt-baseline-legacy-core.xml
```diff
@@ -90,7 +90,7 @@
     <ID>TooManyFunctions:NotifierMessageStore.kt$NotifierMessageStore : MessageStore</ID>
     <ID>TooManyFunctions:Preferences.kt$Preferences : AccountManager</ID>
     <ID>TooManyFunctions:PushController.kt$PushController</ID>
-    <ID>TooManyFunctions:RealGeneralSettingsManager.kt$RealGeneralSettingsManager : GeneralSettingsManager</ID>
+    <ID>TooManyFunctions:DefaultGeneralSettingsManager.kt$DefaultGeneralSettingsManager : GeneralSettingsManager</ID>
```
rafaeltonholo: You can remove this as it was addressed already

core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/GeneralSettings.kt
```diff
@@ -1,45 +1,25 @@
...
  *
  * When adding a setting here, make sure to also add it in these places:
  * - [GeneralSettingsManager] (write function)
- * - [RealGeneralSettingsManager.loadGeneralSettings]
- * - [RealGeneralSettingsManager.writeSettings]
+ * - [DefaultGeneralSettingsManager.loadGeneralSettings]
  * - [GeneralSettingsDescriptions]
  */
 // TODO: Move over settings from K9
+// TODO(#9432): Split in smaller classes per group of settings.
```
rafaeltonholo: You can remove this TODO message as you have addressed all the others

core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/GeneralSettings.kt
```diff
@@ -59,3 +39,8 @@ enum class SubTheme {
     DARK,
     USE_GLOBAL,
 }
+enum class BackgroundOps {
```
rafaeltonholo, line 41: nit: add an empty space between SubTheme and BackgroundOps for consistency
```suggestion

enum class BackgroundOps {
```

core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/GeneralSettingsManager.kt
```diff
@@ -1,39 +1,26 @@
 package net.thunderbird.core.preference
 
 import kotlinx.coroutines.flow.Flow
-import net.thunderbird.core.preference.privacy.PrivacySettingsManager
 
 /**
  * Retrieve and modify general settings.
  *
  * TODO: Add more settings as needed.
```
rafaeltonholo: Could you please remove this TODO? As we are not using the approach of having multiple setters here, it has no meaning anymore

core/preference/impl/src/commonMain/kotlin/net/thunderbird/core/preference/display/DefaultDisplaySettingsPreferenceManager.kt
```diff
@@ -0,0 +1,160 @@
...
+    private fun loadConfig(): DisplaySettings = DisplaySettings(
+        fixedMessageViewTheme = storage.getBoolean(
+            KEY_FIXED_MESSAGE_VIEW_THEME,
+            DISPLAY_SETTINGS_FIXED_MESSAGE_VIEW_THEME,
+        ),
+        isShowUnifiedInbox = storage.getBoolean(KEY_SHOW_UNIFIED_INBOX, DISPLAY_SETTINGS_IS_SHOW_UNIFIED_INBOX),
+        showRecentChanges = storage.getBoolean(KEY_SHOW_RECENT_CHANGES, DISPLAY_SETTINGS_SHOW_RECENT_CHANGES),
+        appTheme = storage.getEnumOrDefault(KEY_THEME, AppTheme.FOLLOW_SYSTEM),
+        messageViewTheme = storage.getEnumOrDefault(KEY_MESSAGE_VIEW_THEME, SubTheme.USE_GLOBAL),
+        messageComposeTheme = storage.getEnumOrDefault(KEY_MESSAGE_COMPOSE_THEME, SubTheme.USE_GLOBAL),
```
rafaeltonholo: Ideally, all the settings default values, including Enums, should ideally be stored in a DEFAULT top-level property. This approach can help prevent inconsistent defaults for the same settings throughout the code.

core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/network/NetworkSettings.kt
```diff
@@ -0,0 +1,7 @@
+package net.thunderbird.core.preference.network
+
+import net.thunderbird.core.preference.BackgroundOps
+
+data class NetworkSettings(
```
rafaeltonholo, line 7: Could you please move the enum value to a top-level property? This will help to prevent inconsistent defaults for the same settings throughout the code.
```suggestion
val NETWORK_SETTINGS_DEFAULT_BACKGROUND_OPS = BackgroundOps.ALWAYS

data class NetworkSettings(
    val backgroundOps: BackgroundOps = NETWORK_SETTINGS_DEFAULT_BACKGROUND_OPS,
)
```

core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/display/DisplaySettings.kt
```diff
@@ -0,0 +1,47 @@
...
+const val DISPLAY_SETTINGS_IS_SHOW_CONTACT_PICTURE = true
+const val DISPLAY_SETTINGS_IS_CHANGE_CONTACT_NAME_COLOR = true
+const val DISPLAY_SETTINGS_IS_COLORIZE_MISSING_CONTACT_PICTURE = false
+const val DISPLAY_SETTINGS_IS_USE_BACKGROUND_AS_INDICATOR = false
+const val DISPLAY_SETTINGS_IS_SHOW_COMPOSE_BUTTON_ON_MESSAGE_LIST = true
+const val DISPLAY_SETTINGS_IS_THREAD_VIEW_ENABLED = true
+const val DISPLAY_SETTINGS_IS_USE_MESSAGE_VIEW_FIXED_WIDTH_FONT = false
+const val DISPLAY_SETTINGS_IS_AUTO_FIT_WIDTH = true
+
+data class DisplaySettings(
```
rafaeltonholo, line 28: Could you please move the enum values to a top-level property? This will help to prevent inconsistent defaults for the same settings throughout the code.

core/preference/impl/src/commonMain/kotlin/net/thunderbird/core/preference/notification/DefaultNotificationPreferenceManager.kt
```diff
@@ -0,0 +1,67 @@
...
+import kotlinx.coroutines.sync.withLock
+import net.thunderbird.core.logging.Logger
+import net.thunderbird.core.preference.storage.Storage
+import net.thunderbird.core.preference.storage.StorageEditor
+
+private const val TAG = "DefaultNotificationPreferenceManager"
+
+class DefaultNotificationPreferenceManager(
+    private val logger: Logger,
+    private val storage: Storage,
```
rafaeltonholo: nit: The `Storage` parameter is only used during the initialization of the `configState`, thus it doesn't need to be a property:

```suggestion
    storage: Storage,
```

core/ui/theme/manager/src/main/java/net/thunderbird/core/ui/theme/manager/ThemeManager.kt
```diff
@@ -80,9 +81,13 @@ class ThemeManager(
...
-            generalSettingsManager.setMessageViewTheme(SubTheme.LIGHT)
+            generalSettingsManager.update { settings ->
+                settings.copy(display = settings.display.copy(messageViewTheme = SubTheme.LIGHT))
+            }
         } else {
-            generalSettingsManager.setMessageViewTheme(SubTheme.DARK)
+            generalSettingsManager.update { settings ->
+                settings.copy(display = settings.display.copy(messageViewTheme = SubTheme.DARK))
+            }
         }
```
rafaeltonholo, line 91: nit: This could be simplified to:
```suggestion
        generalSettingsManager.update { settings ->
            settings.copy(
                display = settings.display.copy(
                    messageViewTheme = if (messageViewTheme === Theme.DARK) {
                        SubTheme.LIGHT
                    } else {
                        SubTheme.DARK
                    },
                ),
            )
        }
```

feature/mail/message/list/src/main/kotlin/net/thunderbird/feature/mail/message/list/ui/dialog/SetupArchiveFolderDialogViewModel.kt
```diff
@@ -30,7 +31,7 @@ internal class SetupArchiveFolderDialogViewModel(
     private val resourceManager: StringsResourceManager,
     private val generalSettingsManager: GeneralSettingsManager,
 ) : ViewModel(
-    initialState = if (generalSettingsManager.getSettings().shouldShowSetupArchiveFolderDialog) {
+    initialState = if (generalSettingsManager.getConfig().display.shouldShowSetupArchiveFolderDialog) {
```
rafaeltonholo, line 34: nit: The benefit of having the `GeneralSettingsManager` split in smaller managers is that for classes that only consume/update a single group, we could use directly that group manager instead of the `GeneralSettingsManager`.

I would like to suggest the following:
```suggestion
    private val displaySettingsManager: DisplaySettingsPreferenceManager,
) : ViewModel(
    initialState = if (displaySettingsManager.getConfig().shouldShowSetupArchiveFolderDialog) {
```

feature/mail/message/list/src/main/kotlin/net/thunderbird/feature/mail/message/list/ui/dialog/SetupArchiveFolderDialogViewModel.kt
```diff
@@ -125,7 +126,13 @@ internal class SetupArchiveFolderDialogViewModel(
...
     private fun onDismissClicked() {
         viewModelScope.launch {
-            generalSettingsManager.setSetupArchiveShouldNotShowAgain(state.value.isDoNotShowDialogAgainChecked.not())
+            generalSettingsManager.update { settings ->
+                settings.copy(
+                    display = settings.display.copy(
+                        shouldShowSetupArchiveFolderDialog = state.value.isDoNotShowDialogAgainChecked.not(),
+                    ),
+                )
+            }
```
rafaeltonholo, line 135: nit: If you choose to use the `DisplaySettingsPreferenceManager` instead of the `GeneralSettingsManager`, you can use the following:
```suggestion
            displaySettingsManager.update { settings ->
                settings.copy(
                    shouldShowSetupArchiveFolderDialog = state.value.isDoNotShowDialogAgainChecked.not(),
                )
            }
```

legacy/core/src/main/java/com/fsck/k9/controller/push/AutoSyncManager.kt
```diff
@@ -6,18 +6,22 @@ import android.content.Context
...
+import net.thunderbird.core.preference.BackgroundOps
+import net.thunderbird.core.preference.GeneralSettingsManager
 
 /**
  * Listen for changes to the system's auto sync setting.
  */
-internal class AutoSyncManager(private val context: Context) {
+internal class AutoSyncManager(
+    private val context: Context,
+    private val generalSettingsManager: GeneralSettingsManager,
```
rafaeltonholo, line 18: nit: You could use `NetworkSettingsPreferenceManager` directly instead:
```suggestion
    private val networkSettingsManager: NetworkSettingsPreferenceManager,
```

legacy/core/src/main/java/com/fsck/k9/controller/push/AutoSyncManager.kt
```diff
@@ -6,18 +6,22 @@ import android.content.Context
...
+) {
     val isAutoSyncDisabled: Boolean
         get() = respectSystemAutoSync && !ContentResolver.getMasterSyncAutomatically()
 
     val respectSystemAutoSync: Boolean
-        get() = K9.backgroundOps == K9.BACKGROUND_OPS.WHEN_CHECKED_AUTO_SYNC
+        get() = generalSettingsManager.getConfig().network.backgroundOps == BackgroundOps.WHEN_CHECKED_AUTO_SYNC
```
rafaeltonholo, line 24: nit: If you choose to use directly the `NetworkSettingsPreferenceManager`, you could apply the following:
```suggestion
        get() = networkSettingsManager.getConfig().backgroundOps == BackgroundOps.WHEN_CHECKED_AUTO_SYNC
```

legacy/core/src/main/java/com/fsck/k9/job/MailSyncWorker.kt
```diff
@@ -4,17 +4,19 @@ import android.content.ContentResolver
...
 import net.thunderbird.core.android.account.LegacyAccount
 import net.thunderbird.core.logging.legacy.Log
+import net.thunderbird.core.preference.BackgroundOps
+import net.thunderbird.core.preference.GeneralSettingsManager
 
 // IMPORTANT: Update K9WorkerFactory when moving this class and the FQCN no longer starts with "com.fsck.k9".
 class MailSyncWorker(
     private val messagingController: MessagingController,
     private val preferences: Preferences,
+    private val generalSettingsManager: GeneralSettingsManager,
```
rafaeltonholo, line 19: nit: You could use `NetworkSettingsPreferenceManager` directly instead:
```suggestion
    private val networkSettingsManager: NetworkSettingsPreferenceManager,
```

legacy/core/src/main/java/com/fsck/k9/job/MailSyncWorker.kt
```diff
@@ -57,10 +59,10 @@ class MailSyncWorker(
     }
 
     private fun isBackgroundSyncDisabled(): Boolean {
-        return when (K9.backgroundOps) {
-            K9.BACKGROUND_OPS.NEVER -> true
-            K9.BACKGROUND_OPS.ALWAYS -> false
-            K9.BACKGROUND_OPS.WHEN_CHECKED_AUTO_SYNC -> !ContentResolver.getMasterSyncAutomatically()
+        return when (generalSettingsManager.getConfig().network.backgroundOps) {
```
rafaeltonholo, line 62: nit: If you choose to use directly the `NetworkSettingsPreferenceManager`, you could apply the following:
```suggestion
        return when (networkSettingsManager.getConfig().backgroundOps) {
```

legacy/core/src/main/java/com/fsck/k9/job/MailSyncWorkerManager.kt
```diff
@@ -7,17 +7,19 @@ import androidx.work.NetworkType
...
 import net.thunderbird.core.logging.Logger
 import net.thunderbird.core.logging.legacy.Log
+import net.thunderbird.core.preference.BackgroundOps
+import net.thunderbird.core.preference.GeneralSettingsManager
 
 class MailSyncWorkerManager(
     private val workManager: WorkManager,
     val clock: Clock,
     val syncDebugLogger: Logger,
+    val generalSettingsManager: GeneralSettingsManager,
```
rafaeltonholo, line 25: nit: You could use `NetworkSettingsPreferenceManager` directly instead:
```suggestion
    private val networkSettingsManager: NetworkSettingsPreferenceManager,
```

legacy/core/src/main/java/com/fsck/k9/job/MailSyncWorkerManager.kt
```diff
@@ -67,7 +69,8 @@ class MailSyncWorkerManager(
         }
     }
 
-    private fun isNeverSyncInBackground() = K9.backgroundOps == K9.BACKGROUND_OPS.NEVER
+    private fun isNeverSyncInBackground() =
+        generalSettingsManager.getConfig().network.backgroundOps == BackgroundOps.NEVER
```
rafaeltonholo, line 76: nit: If you choose to use directly the `NetworkSettingsPreferenceManager`, you could apply the following:
```suggestion
        networkSettingsManager.getConfig().backgroundOps == BackgroundOps.NEVER
```

legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
```diff
@@ -0,0 +1,151 @@
...
+    private val displaySettingsSettingsPreferenceManager: DisplaySettingsPreferenceManager,
+    private val networkSettingsPreferenceManager: NetworkSettingsPreferenceManager,
+    private val backgroundDispatcher: CoroutineDispatcher = Dispatchers.IO,
+) : GeneralSettingsManager {
+    val mutex = Mutex()
+
+    // TODO(#9232): Should be removed when K9 settings is completely migrated
+    // This fallback is required until we finalize the split of the GeneralSettings class and Manager.
+    // The GeneralSettings must be composed by other smaller Managers flows.
+    private val k9GeneralSettingsFallback = MutableStateFlow(value = loadGeneralSettings())
```
rafaeltonholo: You can remove this now as it is not being used anymore.
```suggestion
```

legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
```diff
@@ -0,0 +1,151 @@
...
+
+    override fun getConfig(): GeneralSettings = generalSettings.value
+
+    override fun getConfigFlow(): Flow<GeneralSettings> = generalSettings
+
+    @Synchronized
+    fun loadSettings() {
+        K9.loadPrefs(preferences.storage)
+        // TODO(#9232): Should be removed when K9 settings is completely migrated
+        k9GeneralSettingsFallback.update { loadGeneralSettings() }
```
rafaeltonholo: You can remove this now as it is not being used anymore.
```suggestion
```

legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Deprecated(message = "This only exists for collaboration with the K9 class")
+    fun saveSettingsAsync() {
+        coroutineScope.launch(backgroundDispatcher) {
+            val settings = updateGeneralSettingsWithStateFromK9()
+            save(config = settings)
+        }
```
rafaeltonholo: You can remove the `updateGeneralSettingsWithStateFromK9()`:
```suggestion
        coroutineScope.launch(backgroundDispatcher) {
            save(config = getConfig())
        }
```

legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Synchronized
+    @Deprecated("This only exists for collaboration with the K9 class and should be removed after #9232")
+    private fun updateGeneralSettingsWithStateFromK9(): GeneralSettings {
+        return getSettings().also { generalSettings ->
+            k9GeneralSettingsFallback.update { generalSettings }
+        }
+    }
```
rafaeltonholo: We can remove this method:
```suggestion
```

legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
```diff
@@ -0,0 +1,151 @@
...
+        val editor = preferences.createStorageEditor()
+        K9.save(editor)
+        editor.commit()
+        changePublisher.publish()
+    }
+
+    private fun loadGeneralSettings(): GeneralSettings {
+        val settings = GeneralSettings()
+        return settings
+    }
```
rafaeltonholo: I guess we don't need a method to create an instance from `GeneralSettings` as we are just instantiating it.
```suggestion
```

legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
```diff
@@ -0,0 +1,151 @@
...
+    // This fallback is required until we finalize the split of the GeneralSettings class and Manager.
+    // The GeneralSettings must be composed by other smaller Managers flows.
+    private val k9GeneralSettingsFallback = MutableStateFlow(value = loadGeneralSettings())
+
+    private val generalSettings = MutableStateFlow(value = loadGeneralSettings())
+        .combine(privacySettingsPreferenceManager.getConfigFlow()) { generalSettings, privacySettings ->
+            generalSettings.copy(
+                privacy = privacySettings,
+            )
+        }
```
rafaeltonholo: As there are no other dependencies to create the `GeneralSettings`, you don't need to create a `MutableStateFlow`. Instead, you could directly call the `PrivacySettingsPreferenceManager` and map its result to the `GeneralSettings` as shown below:

```suggestion
    private val generalSettings = privacySettingsPreferenceManager.getConfigFlow()
        .map { privacy ->
            GeneralSettings(privacy = privacy)
        }
```

legacy/core/src/main/java/com/fsck/k9/preferences/GeneralSettingsDescriptions.java
rafaeltonholo, line 1: The idea of having the DEFAULT consts in the configurations is to allow us to have consistent defaults for the same settings throughout the code.

This also means you will need to update the default values here to use the consts instead.

legacy/core/src/main/java/com/fsck/k9/preferences/GeneralSettingsDescriptions.java
```diff
@@ -308,11 +309,7 @@ class GeneralSettingsDescriptions {
             new V(90,
                 new EnumSetting<>(PostMarkAsUnreadNavigation.class, PostMarkAsUnreadNavigation.ReturnToMessageList))
         ));
-        s.put(
-            RealGeneralSettingsManagerKt.KEY_SHOULD_SHOW_SETUP_ARCHIVE_FOLDER_DIALOG,
-            Settings.versions(
-                new V(105, new BooleanSetting(true))
-            )
+        s.put("shouldShowSetupArchiveFolderDialog", Settings.versions(new V(105, new BooleanSetting(true)))
```
rafaeltonholo: Please revert this change. The purpose of having the `KEY` constants is to ensure consistent key values for the same settings throughout the code, which helps prevent typo errors.

Since the `RealGeneralSettingsManager` has changed and the constant `KEY_SHOULD_SHOW_SETUP_ARCHIVE_FOLDER_DIALOG` has been migrated to the `DefaultDisplaySettingsPreferenceManager` class, we need to update the code to use it correctly rather than reverting to hardcoded strings.

Please move all the key constants from the companion object in `DefaultDisplaySettingsPreferenceManager` to the companion object in `DisplaySettingsPreferenceManager`, making them public, and use them in the code here.

The same should apply to the other `PreferenceManager`s you created here.

You don't need to update all the other hardcoded keys present in this file right now, as we can address in a follow-up task.

config/detekt/detekt-baseline-legacy-core.xml
shamim-emon: Done

core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/GeneralSettings.kt
shamim-emon: Done

core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/GeneralSettingsManager.kt
shamim-emon: Done

core/preference/impl/src/commonMain/kotlin/net/thunderbird/core/preference/display/DefaultDisplaySettingsPreferenceManager.kt
shamim-emon: Done

core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/display/DisplaySettings.kt
shamim-emon: Done

legacy/core/src/main/java/com/fsck/k9/preferences/GeneralSettingsDescriptions.java
shamim-emon: Done

shamim-emon:
@rafaeltonholo thank you for your feedback, I have resolved them.
Regarding the **nits**, can I do them in next PR?

rafaeltonholo (approved):
LGTM

legacy/core/src/main/java/com/fsck/k9/preferences/GeneralSettingsDescriptions.java
wmontwe: The descriptions are used for the import/export functionality. So we need to be carefull when we couple the key with the description. If a key is renamed we need to replace the key references by a hard coded version of the old key name and write a migration.

I think coupling the keys is not a good idea. But I don't have a good idea either at the moment, besides decoupling the import/export related functionality into a dedicated module.

shamim-emon: @wmontwe so can we continue coupling key and decouple import/export related functionality into a dedicated module in another PR?

wmontwe: @shamim-emon I think they should not be coupled, we used the same names but they should be specific to the import/export functionality. With the new settings structure it will be easier to couple the import/export to the actual data objects instead of the property names.

wmontwe (approved):
Thank you!

Issue #9429, opened 2025-07-07
[Android 15] Enable edge to edge on daily


kewisch:
Let's enable edge to edge support and targeting the new API on daily, then give it a review to see what screens need adjustment.

Alternatively, enable it on a debug build and then post screenshots of all screens that require work so we know what needs to be done.

asoucar:
@kewisch to be clear, does this involve fixing the issues created or just enabling?

kewisch:
Just enabling and making an assessment of what the screens look like would be great. What I would recommend to get started is to enable it locally and then screenshot all the screens we have to see how good/bad it is. You might need to enable safe drawing areas in the one or other view, these should ideally be quick fixes we can do in the scope of this week.

asoucar:
Update: Enabled locally to get some screenshots. MessageList already handles edge to edge luckily. The issues on other screens are fairly minor but do cause navigation issues. See nav bar in below image overlapping the status bar:

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/59c0f858-22ec-4f1a-9865-fc7966779665" />

All other screen issues I found are similar to this one. My quick fix plan for this would be to mimic the changes made for MessageList to add a minor inset to fix post edge to edge navigation. @laurelterlesky is this ok with design as a first steps solution?

laurelterlesky:
Yes. That sounds great @asoucar

Issue #9459, opened 2025-07-17
Handle Edge to Edge cutting off toolbar


asoucar:
Fix below cut off across all screens that have a top bar
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/8a13a091-c685-4abe-96e3-44a5e2f5e633" />

Update API target

Issue #9338, opened 2025-06-17
The “Download Complete Message” button is located behind the bottom navigation bar and cannot be selected


vitacola45:
### Checklist

- [x] I have used the search function to see if someone else has already submitted the same bug report.
- [x] I will describe the problem with as much detail as possible.

### App

Thunderbird for Android

### App version

11.0.b3

### Where did you get the app from?

F-Droid

### Android version

14 - One-UI 6.1

### Device model

Samsung A52s 5G  SM-A528B/DS

### Steps to reproduce

1. open a message
2. scroll down

### Expected behavior

the "Load complete message" button should appear above the lower navigation bar and be clickable

### Actual behavior

<img src="https://github.com/user-attachments/assets/8fc9dfd2-9a30-4246-95f9-84fa26e0a538" width="45%"/>
<img src="https://github.com/user-attachments/assets/339039c4-16b7-4db2-96dd-65d4d3b11d41" width="45%"/>
<img src="https://github.com/user-attachments/assets/617d1550-a92f-4105-b797-905eadca661a" width="45%"/>

michaelnum:
the same for me (Android 15, Moto G34)

rafaeltonholo:
Thanks @vitacola45 and @michaelnum for the report! I also reproduced this using my Redmi A3, with Android 15, while using the 3-button navigation bar.

spymorass:
Same for me (Android 15, OnePlus 12)

vitacola45:
The same for my Tablet (Samsing Tab A7 SM-T500; Android 12; One-UI 4.1

tarmacalastair:
Also experiencing this

joohnq:
I found the cause of this, and I was able to reproduce it on multiple devices here.
I'll send a pull request with the solution.

The UI was being overlapped by system UI elements (like navigation bars and display cutouts) because WindowInsets were not properly applied to the root container.

| Before | After |
| ------ | ----- |
| ![Image](https://github.com/user-attachments/assets/39050253-18ed-4cd5-8661-07d36b7d6a03) | ![Image](https://github.com/user-attachments/assets/863e1d44-cee6-49bc-bffd-6be11ca2df3d) |

rafaeltonholo:
> I found the cause of this, and I was able to reproduce it on multiple devices here. I'll send a pull request with the solution.
> 
[...]

Issue #9165, opened 2025-05-14
transparent 3 button navigation bar (issue 9157) not yet solved


HelmutL:
### Checklist

- [x] I have used the search function to see if someone else has already submitted the same bug report.
- [x] I will describe the problem with as much detail as possible.

### App

Thunderbird for Android

### App version

10.0b2

### Where did you get the app from?

None

### Android version

15

### Device model

Xiaomi 14

### Steps to reproduce

Settings in TB:
Use dark mode (system mode)
+ use light mode for messages

### Expected behavior

navigation bar should be clearly visible

### Actual behavior

Navigation not distinctive enough 
practically invisible in message view

--images removed for privacy reasons---

kewisch:
9157 has Thunderbird 11 as the milestone, have you tested this in 11.0b2 ?

HelmutL:
Yes, I tried with 11.9b2.
10.0b2 is a typo. Sorry for that

HelmutL:
Sorry, should be 11.0b2

kewisch:
Thanks for letting us know, I'll check with the team if we can confirm. Where did you download 11.0b2 from?

HelmutL:
I was using FF updater - I think it downloads from github

HelmutL:
Just retried with download from github - same problem

ramonagavrilescu:
Hello Philip,

We have checked this request on TB Android 11.0b2, we have updated the app from google play.
We have tested on multiple devices Samsung Galaxy Tab9(Android14), Samsung Galaxy Fold 4(Android15) , Pixel 7(Android14) OPPO A58(Android 13).
We have fiddled with Light/ Dark/Use System default theme in combination with Message view theme Light/Dark/Use app theme, on view message list as well as  read messages.
The only device we were able to reproduce this issue is on OPPO A58(Android13) in Dark theme.
We have provided screenshots for clarity.

![Image](https://github.com/user-attachments/assets/6ff6d76c-e21f-469f-8f76-afe34bad254a)
![Image](https://github.com/user-attachments/assets/d33ced16-ac67-4360-8929-4e0d9219d3b9)
![Image](https://github.com/user-attachments/assets/3631e716-8796-40a0-b141-51021511001c)
![Image](https://github.com/user-attachments/assets/f9549a13-82bd-42ab-b6c3-b1d5271169e9)

kewisch:
@marcRDZ has reproduced and will take a look at this

[...]

PR #9460, opened 2025-07-17
Update targetSdk to 35 and use enableEdgeToEdge

asoucar:
Necessary update prior to Aug 31, 2025 per Google Play Store requirements
Resolves: #9429 #9459

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
@@ -12,14 +12,9 @@ import android.view.MenuItem
 import android.view.View
 import android.view.animation.AnimationUtils
 import android.widget.ProgressBar
-import androidx.activity.enableEdgeToEdge
 import androidx.appcompat.app.ActionBar
 import androidx.appcompat.view.ActionMode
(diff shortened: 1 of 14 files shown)


legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
```diff
@@ -204,46 +197,6 @@ open class MessageList :
...
-    private fun initializeContainerInsets() {
-        val container = findViewById<View>(R.id.container)
-
-        ViewCompat.setOnApplyWindowInsetsListener(container) { v, windowsInsets ->
-            val insets = windowsInsets.getInsets(displayCutout() or navigationBars())
-            v.setPadding(insets.left, 0, insets.right, insets.bottom)
-
-            windowsInsets
-        }
-    }
```
rafaeltonholo: You still need to include the `insets.bottom` in the `MessageList` container, otherwise, this will cause regression on two bugs we fixed before: #9338, #9165 

| (Before) Download button hidden | (After) Download button |
| --- | --- |
| Image | Image |
| (Before) Navigation bar | (After) Navigation bar |
| Image | Image |

rafaeltonholo (changes requested):

PR #9469, opened 2025-07-18
feat(avatar): relax the data mapper logic

### LINKED ISSUE #9468: The avatar mapping it too restrictive and should apply defaults
The current mapping logic for avatars is to restrictive and needs to provide defaults instead of throwing exceptions.

wmontwe:
Relax the data mapper logic by adding a default value instead of throwing exceptions

Fixes #9468


asoucar (approved):

coreycb:
Hi, It looks like this may be a feature, in which case it should ride the train: https://github.com/thunderbird/thunderbird-android/blob/main/docs/ci/RELEASE.md#uplift-criteria - We can cut a new beta next Monday if I'm mistaken. For now, I'm going to remove the beta label.

wmontwe:
@coreycb This is fixing the top crash for current 12.0b4. I think it's worth uplifting.

coreycb:
@wmontwe Got it, in that case lets uplift

PR #9481, opened 2025-07-22
Refactor search to kotlin - part 3

### LINKED ISSUE #9229: Change feature search to Kotlin

wmontwe:
This finishes the Java to Kotlin rewrite of the search functionality.

The remaining `LocalMessageSearch` was converted to Kotlin and the most important change is replacing `Parcelable` with `Serializable`.

The search module is now a full kmp module and the implemetation has been moved under `impl-legacy` to prepare for a better api design in the future.

Resolves #9229


wmontwe:
@asoucar Found a bug and need to fix it first.

wmontwe:
> @asoucar Found a bug and need to fix it first.

Fixed

PR #9506, opened 2025-07-25
feat(notification): add notification action intent creator

rafaeltonholo:
Changeset split from https://github.com/thunderbird/thunderbird-android/pull/9415.
Part of https://github.com/thunderbird/thunderbird-android/issues/9245.

PR #9415, opened 2025-07-02
feat(in-app-notification): Create NotificationSender to manage notification providers

rafaeltonholo:
Resolves #9245 

This pull request introduces a `NotificationSender` responsible for managing and dispatching notifications through various providers. As the application now supports both in-app and system notifications, this manager simplifies the process of triggering a notification by abstracting the provider-specific logic.

The `NotificationSender` receives a notification object and determines the appropriate provider to use for delivery.

### Changes Included

- `NotificationSender` Implementation: Core manager for handling notification logic.
- `InAppNotificationNotifier`: Implementation for displaying notifications within the app UI.
- `SystemNotificationNotifier`: Implementation for triggering native system notifications.
- Enhanced Notification Features:
    - Added support for notification actions, allowing for interactive notifications.
    - Included notification icons and custom styling capabilities.
- Debug Tools: A new secret debug settings screen has been added to facilitate testing of the notification system and other options in the future.
- UI Components:
    - Added a horizontal divider atom to the design system.
    - Resolved rendering issues with theme previews.

PR #9508, opened 2025-07-25
feat(notification): trigger in-app notification using `InAppNotificationNotifier`

### LINKED ISSUE #9245: In-App Error Notifications - Create NotificationSender
As we are going to have at least two different notification providers (system and in-app), it is important to have a manager to simplify the notification triggering.

The `NotificationSender` will be responsible for receiving a notification object and triggering the notification using the appropriate provider.

rafaeltonholo:
Depends on #9500, #9506 
Changeset split from #9415.
Resolve #9245.

- Implement the `InAppNotificationCommand` and `InAppNotificationNotifier`
- Add `InAppNotificationEventBus` and `InAppNotificationReceiver` to broadcast the notification over the application without using Android `BroadcastReceiver`.
- Add `InAppNotificationReceiverCompat` to enable Java classes to consume the `InAppNotificationReceiver` without requiring RxJava or other `Flow` converters.

Difference between #9245 and this PR:
- Dropped the usage of `BroadcastReceiver` in favour of `InAppNotificationEventBus`
- "Show" in-app notification within the Debug secret screen via the "Notification status log" section.

https://github.com/user-attachments/assets/e1dbbfa5-e059-4135-8062-16dc2d2a5ad7

### DIFF
--- a/feature/notification/impl/src/commonTest/kotlin/net/thunderbird/feature/notification/impl/command/InAppNotificationCommandTest.kt
+++ b/feature/notification/impl/src/commonTest/kotlin/net/thunderbird/feature/notification/impl/command/InAppNotificationCommandTest.kt
@@ -1,0 +1,141 @@
+package net.thunderbird.feature.notification.impl.command
+
+import assertk.all
+import assertk.assertThat
+import assertk.assertions.hasMessage
+import assertk.assertions.isEqualTo
+import assertk.assertions.isInstanceOf
+import assertk.assertions.prop
+import dev.mokkery.matcher.any
+import dev.mokkery.spy
+import dev.mokkery.verify.VerifyMode.Companion.exactly
+import dev.mokkery.verifySuspend
+import kotlin.test.Test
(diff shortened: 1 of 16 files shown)


feature/notification/impl/src/commonTest/kotlin/net/thunderbird/feature/notification/impl/command/InAppNotificationCommandTest.kt
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+    @Test
+    fun `execute should return Success when display_in_app_notifications feature flag is Enabled`() =
+        runTest {
+            // Arrange
+            val notification = FakeNotification(
+                severity = NotificationSeverity.Information,
+            )
+            val notifier = spy(FakeInAppNotificationNotifier())
```
wmontwe, line 103: I really like the use of spy, makes the fakes way simples as you don't need to record any invocations within the fake.

PR #9507, opened 2025-07-25
feat(notification): trigger system notification using `SystemNotificationNotifier`

rafaeltonholo:
Depends on #9500, #9506 
Changeset split from #9415.
Part of #9245.

- Implement the `SystemNotificationCommand` and `SystemNotificationNotifier`
- Add `AlarmPermissionMissingNotificationTapActionIntentCreator` to handle the notification tap action for `AlarmPermissionMissingNotification`
- trigger system notification using `SystemNotificationNotifier`

Difference between #9245 and this PR:
- Dropped the `SystemNotificationIntentCreator` in favour of `NotificationActionIntentCreator`, removing unnecessary duplication
- Dropped the `NotificationIdFactory` in favour of `NotificationRegistry`

Screen recording:

https://github.com/user-attachments/assets/24cdc65e-3888-45dd-9f77-584d787ba4ed

PR #9519, opened 2025-07-28
feat(notification): add notification command exception

rafaeltonholo:
Part of #9245.


rafaeltonholo:
Merged as part of #9507

PR #9472, opened 2025-07-18
fix(message-list): app crash when threaded view is enabled

rafaeltonholo:
Fixes #9470

PR #9433, opened 2025-07-07
fix(message-list): back button not closing app when message list is in inbox after returning from another folder and pressing back

rafaeltonholo:
Fixes #9407
| Before fix | After fix |
| --------- | ---------|
| <video src="https://github.com/user-attachments/assets/b1565548-7962-4c24-92f3-4fbc1c422cd7"/> | <video src="https://github.com/user-attachments/assets/1faad30b-96bf-458d-9561-9bf1d5a9e652"/> |

PR #9521, opened 2025-07-28
Uplift: fix back button not closing app when message list is in inbox after returning from another folder and pressing back

rafaeltonholo:
Uplift #9472 to beta. 

The changes on #9472 supersede the changes on #9433, correctly fixing the navigate back issue without crashing the app when trying to open a Threaded view.

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageList.kt
@@ -1453,11 +1453,11 @@ open class MessageList :
         this.search = search
         singleFolderMode = false
 
+        val folderIds = search.folderIds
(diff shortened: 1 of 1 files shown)


coreycb:
> Risk to taking this patch (and alternatives if risky): N/A

There is always risk in taking a patch that affects executable code. This does not need to be an in-depth analysis.. a few sentences is enough. The patch should be evaluated for what could go wrong. This helps evaluate the risk of taking the patch vs the risk of not taking the patch.

In the future, please add a thoughtful response for this field. I know it can feel cumbersome to follow the uplift process, however there is not much point in having a process if we don't follow it.

Considering this patch has been tested on daily by the reporter and we have 5 weeks until the release of 12.0, I am going to accept the uplift in this case.

PR #9241, opened 2025-05-31
Handle UTF8 strings sent using literal syntax

arnt:
RFC 6855/9755 permits IMAP servers to send UTF8 in strings such as "this", but does not ban sending literals such as {4}
this
instead, and some servers do, and #8772 did not handle that.

Fixes #9212

PR #9523, opened 2025-07-28
Uplift - Handle UTF8 strings sent using literal syntax

wmontwe:
Uplift #9241 to beta

[Approval Request]

Original Issue/Pull request: #9241

Regression caused by (issue #): N/A

User impact if declined: User will face malformed folder names when server supports UTF-8

Testing completed (on daily, etc.): daily and main

Risk to taking this patch (and alternatives if risky): Not known yet.


coreycb:
> Risk to taking this patch (and alternatives if risky): Not known yet.

For this field, the patch should be evaluated for "what could go wrong". This helps evaluate the risk of taking the patch vs the risk of not taking the patch. It's been tested on daily so that lowers the regression risk, even though it has only been on daily for a week.

coreycb:
Wolf, In the future please put more thought into the "Risk of taking this patch" field, as mentioned above.

Considering we're 5 weeks away from 12.0 and this has been tested on daily, I'm going to accept the uplift this time.

coreycb (approved):

PR #9562, opened 2025-08-04
Fix database migration missing avatar type

wmontwe:
Resolves #9547

PR #9563, opened 2025-08-04
Uplift: feat(avatar): relax the data mapper logic

wmontwe:
Uplift #9469 to beta

For a full fix it also requires #9562


coreycb:
Looks good. Thanks Wolf!

PR #9576, opened 2025-08-05
chore: remove foojay resolver

wmontwe:
There is no need to use the foojay resolver plugin, hence we could remove it.


rafaeltonholo (approved):
LGTM

PR #9577, opened 2025-08-05
Replace deprecated MessagingException

NWuensche:
Replaced all K9-MessagingException with Thunderbird-MessagingException. Also made MessagingException#isPermanentFailure public because it's access from the outside now


wmontwe (approved):
Thank you! Just run `./gradlew spotlessApply` to fix the import order.

NWuensche:
@wmontwe I've fixed everything

PR #9588, opened 2025-08-09
Disable edge to edge in Preferences

### LINKED ISSUE #9587: Edge-to-edge Regression - Preferences hide behind three button navigation
### Checklist

### App

Thunderbird for Android

### App version

current main

Other

### Android version

Android 16

### Device model

Pixel 8a

### Steps to reproduce

1. Open Settings
2. Click on E-Mail Account
3. Click on Reading Mail

### Expected behavior

Elements are all clickable

### Actual behavior

Image

### Logs

_No response_

NWuensche:
Fix #9587 

Take safe area into account to not hide items under Android 16


asoucar:
Thank you for working on this fix. Because this issue effects multiple screens across legacy and new code we are going with a different approach to solve this and other screens simultaneously.

Issue #8496, opened 2024-11-04
Why the Breton language has been removed?


ILPlais:
### Checklist

- [X] I have used the search function to see if someone else has already submitted the same feature request.
- [X] I will describe the problem with as much detail as possible.
- [X] This issue only contains a request for one single feature, **not** multiple (related) features.

### App version

8.0

### Problem you are trying to solve

I noticed that the Breton language (brezhoneg) had been removed from the application. Even though it's still present on [Weblate](https://hosted.weblate.org/projects/tb-android/app-strings/br/).
Is there a reason for this?

### Suggested solution

Put back in place the Breton translation of the application.

kewisch:
https://hosted.weblate.org/projects/tb-android/-/br/ is where you can help. We'd love to see you as a contributor for Breton, let us know if you need help getting started.

PR #9593, opened 2025-08-11
Add Breton to supported languages

ILPlais:
Since the number of translated channels exceeds 70% (as specified in issue #8496), I propose to re-add support for the Breton language.


rafaeltonholo:
Hi @ILPlais,

Thanks for the PR. We have two other new languages that reached the threshold of 70%, so I will add them as well, together with the Breton language, in another PR.

ILPlais:
Is there a planned date for adding new languages?

PR #9599, opened 2025-08-11
feat(message-list): add new Message List UI components

### LINKED ISSUE #9596: Create new Message List UI components
Add the following UI elements to the design system:

Image

Image

rafaeltonholo:
Resolves #9596.

| Light mode | Dark mode |
| ---- | ----- |
| <video src="https://github.com/user-attachments/assets/5c892987-6ffc-4cd9-9eec-e2f969078b47" /> | <video src="https://github.com/user-attachments/assets/44c05cb5-9332-4536-ac14-972c84c2f7f6" /> |

### DIFF
--- a/app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/items/message/CatalogMessageItems.kt
+++ b/app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/items/message/CatalogMessageItems.kt
@@ -1,0 +1,423 @@
+package net.thunderbird.ui.catalog.ui.page.organism.items.message
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.border
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.ColumnScope
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.lazy.grid.LazyGridScope
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.Slider
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.unit.dp
+import app.k9mail.core.ui.compose.designsystem.atom.DividerHorizontal
+import app.k9mail.core.ui.compose.designsystem.atom.text.TextLabelSmall
+import app.k9mail.core.ui.compose.designsystem.atom.text.TextTitleMedium
+import app.k9mail.core.ui.compose.designsystem.atom.text.TextTitleSmall
+import app.k9mail.core.ui.compose.designsystem.atom.textfield.TextFieldOutlined
+import app.k9mail.core.ui.compose.designsystem.molecule.input.CheckboxInput
+import app.k9mail.core.ui.compose.theme2.MainTheme
+import kotlin.math.roundToInt
+import kotlin.random.Random
+import kotlin.time.Clock
+import kotlin.time.ExperimentalTime
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.toLocalDateTime
+import net.thunderbird.core.ui.compose.designsystem.organism.message.ActiveMessageItem
+import net.thunderbird.core.ui.compose.designsystem.organism.message.JunkMessageItem
+import net.thunderbird.core.ui.compose.designsystem.organism.message.NewMessageItem
+import net.thunderbird.core.ui.compose.designsystem.organism.message.ReadMessageItem
+import net.thunderbird.core.ui.compose.designsystem.organism.message.UnreadMessageItem
+import net.thunderbird.ui.catalog.ui.page.common.list.fullSpanItem
+import net.thunderbird.ui.catalog.ui.page.common.list.sectionHeaderItem
+import net.thunderbird.ui.catalog.ui.page.common.list.sectionSubtitleItem
+
+@OptIn(ExperimentalMaterial3Api::class)
+fun LazyGridScope.messageItems() {
+    sectionHeaderItem("Message Item")
+
+    sectionSubtitleItem("Configuration")
+    fullSpanItem {
+        var config by remember {
+            mutableStateOf(
+                value = MessageItemConfiguration(
+                    sender = "Sender Name",
+                    subject = "The subject",
+                    preview = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam eleifend, leo at " +
+                        "elementum luctus, felis nisl placerat enim, quis aliquam erat nibh gravida eros. Nunc ac " +
+                        "elit mauris. Vivamus tristique, nisi eget rutrum condimentum, dui neque bibendum tortor, " +
+                        "id fringilla nisi sem eget velit. In euismod leo luctus, tristique ante et, vulputate " +
+                        "metus. Integer volutpat pulvinar dictum. Suspendisse et orci quis diam convallis accumsan " +
+                        "in non justo.",
+                    hideSection = false,
+                    hideAvatar = false,
+                    swapSenderAndSubject = false,
+                    randomizeAttachment = false,
+                    maxPreviewLines = 2,
+                ),
+            )
+        }
+        Column {
+            MessageItemConfiguration(
+                config = config,
+                onSenderChange = { config = config.copy(sender = it) },
+                onSubjectChange = { config = config.copy(subject = it) },
+                onPreviewChange = { config = config.copy(preview = it) },
+                onHideSectionChange = { config = config.copy(hideSection = it) },
+                onHideAvatarChange = { config = config.copy(hideAvatar = it) },
+                onSwapSenderAndSubjectChange = { config = config.copy(swapSenderAndSubject = it) },
+                onRandomizeAttachmentChange = { config = config.copy(randomizeAttachment = it) },
+                onMaxPreviewLines = { config = config.copy(maxPreviewLines = it) },
+            )
+            DividerHorizontal(modifier = Modifier.padding(MainTheme.spacings.default))
+            CatalogMessageItems(config = config)
+        }
+    }
+}
+
+private data class MessageItemConfiguration(
+    val sender: String,
+    val subject: String,
+    val preview: String,
+    val hideSection: Boolean,
+    val hideAvatar: Boolean,
+    val swapSenderAndSubject: Boolean,
+    val randomizeAttachment: Boolean,
+    val maxPreviewLines: Int,
+)
+
+@Composable
+private fun MessageItemConfiguration(
+    config: MessageItemConfiguration,
+    modifier: Modifier = Modifier,
+    onSenderChange: (String) -> Unit = {},
+    onSubjectChange: (String) -> Unit = {},
+    onPreviewChange: (String) -> Unit = {},
+    onHideSectionChange: (Boolean) -> Unit = {},
+    onHideAvatarChange: (Boolean) -> Unit = {},
+    onSwapSenderAndSubjectChange: (Boolean) -> Unit = {},
+    onRandomizeAttachmentChange: (Boolean) -> Unit = {},
+    onMaxPreviewLines: (Int) -> Unit = {},
+) {
+    Column(
+        modifier = modifier,
+        verticalArrangement = Arrangement.spacedBy(MainTheme.spacings.half),
+    ) {
+        CheckboxInput(
+            text = "Hide sections",
+            onCheckedChange = onHideSectionChange,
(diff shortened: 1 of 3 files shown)


app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/items/message/CatalogMessageItems.kt
```diff
@@ -0,0 +1,487 @@
...
+
+@OptIn(ExperimentalMaterial3Api::class)
+fun LazyGridScope.messageItems() {
+    sectionHeaderItem("Message Item")
+
+    sectionSubtitleItem("Configuration")
+    fullSpanItem {
+        var sender by remember { mutableStateOf("Sender Name") }
+        var subject by remember { mutableStateOf("The subject") }
+        var preview by remember { mutableStateOf("The message preview") }
```
wmontwe: I would add a long text as default to easily review the impact of the preview lines setting.

app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/items/message/CatalogMessageItems.kt
```diff
@@ -0,0 +1,487 @@
...
+    modifier: Modifier = Modifier,
+    onSender: (String) -> Unit = {},
+    onSubject: (String) -> Unit = {},
+    onPreview: (String) -> Unit = {},
+    onHideSection: (Boolean) -> Unit = {},
+    onHideAvatar: (Boolean) -> Unit = {},
+    onSwapSenderAndSubject: (Boolean) -> Unit = {},
+    onMaxPreviewLines: (Int) -> Unit = {},
+) {
+    Column(modifier = modifier) {
```
wmontwe: I think the items should have some spacing.

```suggestion
    Column(
        modifier = modifier,
        verticalArrangement = Arrangement.spacedBy(MainTheme.spacings.half)
    ) {
```

core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/atom/icon/filled/NewMailBadge.kt
```diff
@@ -0,0 +1,54 @@
...
+
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.graphics.vector.path
+import androidx.compose.ui.unit.dp
+import app.k9mail.core.ui.compose.designsystem.atom.icon.Icons
+
+@Suppress("MagicNumber", "UnusedReceiverParameter")
+val Icons.Filled.NewMailBadge: ImageVector
```
wmontwe, line 14: I think it would be beneficial to add a preview for the icons to inspect them. Or do you know a better alternative?

app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/CatalogMoleculeContent.kt
wmontwe, line 1: Just noticed that the file name is wrong, should be `CatalogOrganismContent`

core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/organism/message/MessageItem.kt
```diff
@@ -0,0 +1,272 @@
...
+ *  Defaults to `false`.
+ * @param maxPreviewLines The maximum number of lines to display for the message preview.
+ *  Defaults to `2`.
+ * @param contentPadding The padding to be applied to the content of the message item.
+ *  Defaults to [MessageItemDefaults.defaultContentPadding].
+ * @see MessageItemDefaults
+ */
+@Composable
+internal fun MessageItem(
+    trailing: @Composable () -> Unit,
```
wmontwe: This should be `leading`

app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/items/message/CatalogMessageItems.kt
wmontwe, line 1: I just noticed that when you entered a long preview text and change the preview lines. The message item height is not adjusted.

core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/message/MessageItemSenderText.kt
```diff
@@ -0,0 +1,93 @@
...
+        modifier = modifier,
+        color = color,
+    )
+}
+
+@Composable
+private fun MessageItemSenderText(
+    sender: String,
+    subject: String,
+    swapSenderWithSubject: Boolean,
```
wmontwe, line 76: Shouldn't this be decided in the `MessageItem`? This would reduce the properties to just providing the text to display.

core/ui/compose/designsystem/src/debug/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/message/MessageItemSenderTextPreview.kt
```diff
@@ -0,0 +1,89 @@
...
+            swapSenderWithSubject = true,
+            threadCount = 10,
+        ),
+    ),
+)
+
+@Preview
+@Composable
+private fun MessageItemSenderTextPreview(
+    @PreviewParameter(MessageItemSenderTextPreviewCol::class) params: MessageItemSenderTextPreviewParams,
```
wmontwe, line 69: I like defining the variants as pure data. The only issue I have with the `PreviewParameter` is that it renders the collection as `params 0 .. 5` which doesn't give you context of what exactly your looking at:



But I didn't find a more expressive way to define these.

wmontwe (approved):
Looks good just some little findings.

app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/CatalogMoleculeContent.kt
rafaeltonholo: I will take the opportunity to update it in this PR

core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/atom/icon/filled/NewMailBadge.kt
rafaeltonholo: I completely agree that having a preview for the icons is important. However, I would prefer to have them included in the same file rather than placed in the debug folder. 

Since they don’t strictly require a theme to be previewed, I believe it makes sense to add the preview within the same file.

app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/items/message/CatalogMessageItems.kt
rafaeltonholo: It seems to work fine to me... Tested on both emulator and real devices.
| Emulator | Real device |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/93653b38-f5c4-4974-817e-c3e0d127309d" /> | <video src="https://github.com/user-attachments/assets/43ce04cf-492b-46aa-904b-c2619db0c94a" /> |

Am I missing something?

core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/message/MessageItemSenderText.kt
rafaeltonholo: The issue with allowing the `MessageItem` (the base component) to determine its states is that it would require the `MessageItem` to have knowledge of these states.

The `New` and `Unread` states use the `titleSmall` text style, while the `Active`, `Junk`, and `Read` states utilize the `bodyMedium` text style.

While this composable could be integrated within each `<State>MessageItem`, doing so would result in code duplication, as each state would essentially perform the same function, differing only in the text style used.

Furthermore, Detekt raises concerns about the composable being too large, which could lead to multiple instances of the "same" `MessageItemSenderText` for each state.

Let me know if that answers your question

rafaeltonholo: Also, as this isn't useful outside the scope of the `<State>MessageItem`, I will change its visibility to `internal`

core/ui/compose/designsystem/src/debug/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/message/MessageItemSenderTextPreview.kt
rafaeltonholo: Unfortunately, that is a limitation of the `ParameterPreviewCollection`. We may see some changes soon, as https://issuetracker.google.com/issues/241699422 has been moved to `In Progress`. In the meantime, I could add a message to the preview that will display the `MessageItemSenderTextPreviewParams.toString()`.

Example:
image
image

app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/items/message/CatalogMessageItems.kt
wmontwe: Sorry, I meant that the whole list item doesn't fully resize. I had a look with the layout inspector and the trailing icon is visible but there is another one invisible below. Looking at the layout it seems to be the attachment placeholder. But this prevents the item to collapse and adds unused space when only showing 1 or 2 preview lines and no attachment present.


Screenshot 2025-08-25 at 11 04 39

wmontwe:
I also noticed that the star is not yellow when selected.

rafaeltonholo:
> I also noticed that the star is not yellow when selected.

The star isn't yellow because we don't have the actual colour in the design system. It would require the introduction of the `onSurfaceVariantYellow`

app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/items/message/CatalogMessageItems.kt
rafaeltonholo: That space is intended, and it is following the layout spec when there is a one-line preview. The empty space present there is to make sure the message item keeps its minimum height as expected by the layout definition.
I have tagged you in the Figma where we can see that.

wmontwe: When we assume 2 preview lines are the default as with the design, then the current implementation results in unbalanced items where the visible top padding is different to the bottom padding. I guess this is caused by having a min touch target for the star and also for the attachment which are not present in the design.

core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/message/MessageItemSenderText.kt
wmontwe: Okay, got it.

wmontwe:
> > I also noticed that the star is not yellow when selected.
> 
> The star isn't yellow because we don't have the actual colour in the design system. It would require the introduction of the `onSurfaceVariantYellow`

This could be added as a local color for the component until there is a theme property.

app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/items/message/CatalogMessageItems.kt
rafaeltonholo: Oh, now I see what you mean! The issue is indeed the Attachment spacer that is causing the uneven padding. The problem of not having that spacer is that we will have uneven message list items when attachments are present...

There is also another problem that might be causing that issue, which is the lack of padding between the subject line and the preview line.

Let me make some changes to see if I can overcome that issue.

rafaeltonholo:
> > > I also noticed that the star is not yellow when selected.
> > 
> > 
> > The star isn't yellow because we don't have the actual colour in the design system. It would require the introduction of the `onSurfaceVariantYellow`
> 
> This could be added as a local color for the component until there is a theme property.

Sure, I can hardcode the colour if you prefer

app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/organism/items/message/CatalogMessageItems.kt
rafaeltonholo: I have made a few adjustments to the layout, and I believe now it is as expected. Here are a few changes:

| Preview with 1 line | Preview with 2 lines |
| ----- | ---- |
| image | image |

I have also added padding to the Avatar/selected button icon, so it matches the design specs better. Finally, I have hardcoded the fav icon colour until we have the new colour in the theme.

image

wmontwe: Looks great!

wmontwe (approved):
LGTM

PR #9634, opened 2025-08-20
fix(logs): debug logs ignored when user enable it

### LINKED ISSUE #9561: Debug logging does not include IMAP/POP3 commands in beta versions
### Checklist

### App

Thunderbird for Android

### App version

12.0b4

Other

### Android version

13 (xiaomi.eu 14.0.3.0.TJUMIXM)

### Device model

Xiaomi POCO X3 Pro (global)

### Steps to reproduce

1. Go to Settings -> "General settings" -> Debugging
2. Enable all options (debug logging, sync debug logging, log sensitive information)
3. Refresh a folder
4. Export logs or observe logcat output

### Expected behavior

Logs include IMAP/POP3 commands sent to the mail server and received responses, as well as other debugging information (number of unsynced messages, etc.)

### Actual behavior

There is no useful debugging information in debug logs (see below for an example). This makes debugging problems with message synchronisation much harder.

This problem does not exist in debug builds, but most people who submit bug reports will most likely run a release version. In my case a debug build (from beta branch) synchronises correctly, but the release (12.0b4) downloaded from archive.mozilla.org doesn't, so I only have debug logs from a version where everything is ok...

#9250 introduced logging changes that prevented messages with level lower than INFO (VERBOSE and DEBUG) from being logged at all. #9359 changed the minimum log level to VERBOSE, but only for debug builds. Most log messages enabled by the "Enable debug logging" option are DEBUG or VERBOSE, so enabling this option does almost nothing.

### Logs

```
08-04 14:22:58.350 32664 14487 I MessagingController: Running command 'synchronizeMailbox', seq = 119 (background priority)
08-04 14:22:58.352 32664 14487 I ImapSync: Synchronizing folder konto.pl:INBOX
08-04 14:22:58.731 32664 14458 I rd.android.beta: This is non sticky GC, maxfree is 8388608 minfree is 2097152
08-04 14:23:03.167 32664 14458 I rd.android.beta: This is sticky GC, maxfree is 8388608 minfree is 2097152
08-04 14:23:03.668 32664 14487 I ImapSync: Done synchronizing folder konto.pl:INBOX
08-04 14:23:03.668 32664 14487 I MessagingController:  Command 'synchronizeMailbox' completed
08-04 14:23:03.668 32664 14487 I MessagingController: Running command 'sendPendingMessages', seq = 120 (background priority)
08-04 14:23:03.669 32664 14487 I MessagingController:  Command 'sendPendingMessages' completed
```

rafaeltonholo:
Fixes #9561.

- Fixes the "Enable debug logging" preference not logging any useful data outside debuggable builds
- Refactor:
  - Move `CompositeLogSinkManager` to `:core:logging:api`
  - Replaced hardcoded string qualifier `syncDebug` with the `LogSync.Qualifier.SyncDebug` enum qualifier for improved discoverability.

### DIFF
--- a/app-common/src/main/kotlin/net/thunderbird/app/common/core/AppCommonCoreModule.kt
+++ b/app-common/src/main/kotlin/net/thunderbird/app/common/core/AppCommonCoreModule.kt
@@ -3,6 +3,8 @@ package net.thunderbird.app.common.core
 import android.content.Context
 import kotlin.time.ExperimentalTime
 import net.thunderbird.app.common.core.logging.DefaultLogLevelManager
+import net.thunderbird.core.common.inject.getList
+import net.thunderbird.core.common.inject.singleListOf
 import net.thunderbird.core.logging.DefaultLogger
 import net.thunderbird.core.logging.LogLevel
 import net.thunderbird.core.logging.LogLevelManager
@@ -23,16 +25,14 @@ val appCommonCoreModule: Module = module {
         DefaultLogLevelManager()
     }.bind<LogLevelProvider>()
 
-    single<List<LogSink>> {
-        listOf(
-            ConsoleLogSink(level = LogLevel.VERBOSE),
-        )
-    }
+    singleListOf<LogSink>(
+        { ConsoleLogSink(level = LogLevel.VERBOSE) },
+    )
 
     single<CompositeLogSink> {
         CompositeLogSink(
             logLevelProvider = get(),
-            sinks = get(),
+            sinks = getList(),
         )
     }
 
@@ -46,7 +46,7 @@ val appCommonCoreModule: Module = module {
     single<CompositeLogSink>(named(SYNC_DEBUG_LOG)) {
         CompositeLogSink(
             logLevelProvider = get(),
-            sinks = get(),
+            sinks = getList(),
         )
     }
 
--- a/core/common/src/commonMain/kotlin/net/thunderbird/core/common/inject/KoinMultibindingCollection.kt
+++ b/core/common/src/commonMain/kotlin/net/thunderbird/core/common/inject/KoinMultibindingCollection.kt
@@ -1,0 +1,64 @@
+package net.thunderbird.core.common.inject
+
+import org.koin.core.definition.Definition
+import org.koin.core.module.KoinDslMarker
+import org.koin.core.module.Module
+import org.koin.core.parameter.parametersOf
+import org.koin.core.qualifier.Qualifier
+import org.koin.core.qualifier.named
+import org.koin.core.scope.Scope
+
+// This file must be deleted once https://github.com/InsertKoinIO/koin/pull/1951 is merged to
+// Koin and released on 4.2.0
+
+/**
+ * Defines a singleton list of elements of type [T].
+ *
+ * This function creates a singleton definition for a mutable list of elements.
+ * Each element in the list is resolved from the provided [items] definitions.
+ *
+ * @param T The type of elements in the list.
+ * @param items Vararg of [Definition]s that will be resolved and added to the list.
+ * @param qualifier Optional [Qualifier] to distinguish this list from others of the same type.
+ *                  If null, a default qualifier based on the type [T] will be used.
+ */
+@KoinDslMarker
+inline fun <reified T> Module.singleListOf(vararg items: Definition<T>, qualifier: Qualifier? = null) {
+    single(qualifier ?: defaultListQualifier<T>(), createdAtStart = true) {
+        items.map { definition -> definition(this, parametersOf()) }
+    }
+}
+
+/**
+ * Resolves a [List] of instances of type [T].
+ * This is a helper function for Koin's multibinding feature.
+ *
+ * It uses the [defaultListQualifier] if no [qualifier] is provided.
+ *
+ * @param T The type of instances in the list.
+ * @param qualifier An optional [Qualifier] to distinguish between different lists of the same type.
+ * @return The resolved [MutableList] of instances of type [T].
+ */
+inline fun <reified T> Scope.getList(qualifier: Qualifier? = null) =
+    get<List<T>>(qualifier ?: defaultListQualifier<T>())
+
+/**
+ * Creates a qualifier for a set of a specific type.
+ *
+ * This is used to differentiate between different sets of the same type when injecting dependencies.
+ *
+ * @param T The type of the elements in the set.
+ * @return A qualifier for the set.
+ */
+inline fun <reified T> defaultListQualifier() =
+    defaultCollectionQualifier<List<T>, T>()
+
+/**
+ * Creates a default [Qualifier] for a collection binding.
+ *
+ * @param TCollection The type of the collection (e.g., `List`, `List`).
+ * @param T The type of the elements in the collection.
+ * @return A [Qualifier] that can be used to identify the specific collection binding.
+ */
+inline fun <reified TCollection : Collection<T>, reified T> defaultCollectionQualifier() =
+    named("${TCollection::class.qualifiedName}<${T::class.qualifiedName}>")


core/logging/api/src/commonMain/kotlin/net/thunderbird/core/logging/LogSink.kt
```diff
@@ -32,4 +32,9 @@ interface LogSink {
     fun log(
         event: LogEvent,
     )
+
+    enum class Qualifier {
```
wmontwe: I think this needs a doc to explain what this is used for.

core/logging/api/src/commonMain/kotlin/net/thunderbird/core/logging/LogSink.kt
```diff
@@ -32,4 +32,9 @@ interface LogSink {
...
+        DebuggingLogsEnabledByUser,
```
wmontwe: What about renaming it to `DebugLogsEnabledByUser`?

core/logging/impl-composite/src/commonMain/kotlin/net/thunderbird/core/logging/composite/DefaultCompositeLogSink.kt
```diff
@@ -14,6 +14,10 @@ internal class DefaultCompositeLogSink(
         manager.addAll(sinks)
     }
 
+    override fun canLog(level: LogLevel): Boolean {
```
wmontwe: These changes will ignore the composite own log level and so the globally configured max level.

The tests need to be updated to cover the new cases.

wmontwe (commented):
The `CompositeLogSink` has it's own limiting `LogLevel`, this was introduced to prevent individual loggers to exceed the configured level. These changes will ignore the limiting `LogLevel`.

The issue arrises as we want to log debug logs while the global log level is set to `LogLevel.INFO` for non debuggable builds, which prevents any `LogLevel.DEBUG` or `LogLevel.VERBOSE` from being collected.

We could either introduce a diagnostic composite where the `LogLevel` is set to `LogLevel.DEBUG` OR change the `LogLevel` to be fully user configurable and add a `LogLevelProvider` for dynamic updates based on the user settings.

What do you think?

rafaeltonholo:
> The `CompositeLogSink` has it's own limiting `LogLevel`, this was introduced to prevent individual loggers to exceed the configured level. These changes will ignore the limiting `LogLevel`.
> 
> The issue arrises as we want to log debug logs while the global log level is set to `LogLevel.INFO` for non debuggable builds, which prevents any `LogLevel.DEBUG` or `LogLevel.VERBOSE` from being collected.
> 
> We could either introduce a diagnostic composite where the `LogLevel` is set to `LogLevel.DEBUG` OR change the `LogLevel` to be fully user configurable and add a `LogLevelProvider` for dynamic updates based on the user settings.
> 
> What do you think?

Thanks for explaining that, but I'm still "confused" about the meaning of the `LogLevel` for `CompositeLogSink`. What I had as understanding for the `CompositeLogSink` was that it would be an aggregator for `LogSink`s and not a limitator.

With your explanation, it seems to me that it is both an Aggregator and a Limitator, which is fine, however, that introduces a few problems:

- We end up having multiple `CompositeLogSink`s with a single `LogSink` (`SyncDebug CompositeLogSink` and potentially `Diagnostic CompositeLogSink`).
- Having to create different `Logger`s per `CompositeLogSink`, since the `Logger` has only one `LogSink`.

Additionally, this brings the challenge of updating it at runtime, as we use `Logger` as a singleton and its sink is currently immutable.

That brings me to question whether we really need a log level inside `CompositeLogSink` or even the `CompositeLogSink` itself.

Regarding having a `LogLevelProvider`, that was my initial idea, but I reverted it because it felt like it was going in the opposite direction from the logger api.

core/logging/api/src/commonMain/kotlin/net/thunderbird/core/logging/LogSink.kt
rafaeltonholo: Would ktdocs be enough, or do you also believe it would be beneficial to add it to the logging documentation file?

wmontwe: Just in place docs are enough.

wmontwe:
> > The `CompositeLogSink` has it's own limiting `LogLevel`, this was introduced to prevent individual loggers to exceed the configured level. These changes will ignore the limiting `LogLevel`.
> > The issue arrises as we want to log debug logs while the global log level is set to `LogLevel.INFO` for non debuggable builds, which prevents any `LogLevel.DEBUG` or `LogLevel.VERBOSE` from being collected.
> > We could either introduce a diagnostic composite where the `LogLevel` is set to `LogLevel.DEBUG` OR change the `LogLevel` to be fully user configurable and add a `LogLevelProvider` for dynamic updates based on the user settings.
> > What do you think?
> 
> Thanks for explaining that, but I'm still "confused" about the meaning of the `LogLevel` for `CompositeLogSink`. What I had as understanding for the `CompositeLogSink` was that it would be an aggregator for `LogSink`s and not a limitator.
> 
> With your explanation, it seems to me that it is both an Aggregator and a Limitator, which is fine, however, that introduces a few problems:
> 
>     * We end up having multiple `CompositeLogSink`s with a single `LogSink` (`SyncDebug CompositeLogSink` and potentially `Diagnostic CompositeLogSink`).
> 
>     * Having to create different `Logger`s per `CompositeLogSink`, since the `Logger` has only one `LogSink`.
> 
> 
> Additionally, this brings the challenge of updating it at runtime, as we use `Logger` as a singleton and its sink is currently immutable.
> 
> That brings me to question whether we really need a log level inside `CompositeLogSink` or even the `CompositeLogSink` itself.
> 
> Regarding having a `LogLevelProvider`, that was my initial idea, but I reverted it because it felt like it was going in the opposite direction from the logger api.

The `LogLevel` on the `LogSink` is used to limit the level of details logged by a specific log sink. The `CompositeLogSink` currently behaves like any other sink and uses its own level, so it can reject events before children see them, even if some children would accept those events. This was intended to guarantee that a `LogSink` can't log more verbose events than configured.

Now we have the issue that we use a static global `LogLevel.INFO` that is injected at construction, which prevents dynamic changes and blocks selective DEBUG/VERBOSE changes at runtime.

The changes to `CompositeLogSink` solve the original issue and ensure is just aggregates. But this introduces another issue that unintended log levels are beeing recorded.

I guess we need to add a filter log sink to change the globally allowed log level based on a provider to allow dynamic changes.


Example:

```kotlin
fun interface LevelProvider { fun current(): LogLevel }
```

```kotlin
class LevelFilteredSink(
    private val delegate: LogSink,
    private val provider: LevelProvider
) : LogSink {
    override val level: LogLevel
        get() = provider.current()

    override fun canLog(level: LogLevel): Boolean {
        return this.level <= level && delegate.canLog(level)
    }

    override fun log(event: LogEvent) {
        if (canLog(event.level)) delegate.log(event)
    }
}
```

wmontwe (approved):
Looks great 👍

wmontwe:
@rafaeltonholo I think this should be uplifted to release, I added the task label.

core/preference/impl/src/commonMain/kotlin/net/thunderbird/core/preference/debugging/DefaultDebuggingSettingsPreferenceManager.kt
```diff
@@ -58,4 +61,12 @@ class DefaultDebuggingSettingsPreferenceManager(
             }
         }
     }
+
+    private fun updateDebugLogLevel(config: DebuggingSettings) {
+        if (config.isDebugLoggingEnabled) {
+            logLevelManager.override(LogLevel.DEBUG)
```
krzysdz, line 67: _Fun_ fact: now enabling debug logging in debug builds, raises minimum verbosity level from `VERBOSE` (default) to `DEBUG`.

rafaeltonholo: Will be addressed in #9731

PR #9658, opened 2025-08-26
Uplift to Release - fix(logs): debug logs ignored when user enable it

rafaeltonholo:
Uplift to Release #9634.

### DIFF
--- a/legacy/core/src/main/java/com/fsck/k9/K9.kt
+++ b/legacy/core/src/main/java/com/fsck/k9/K9.kt
@@ -17,6 +17,8 @@ import net.thunderbird.core.common.action.SwipeAction
 import net.thunderbird.core.common.action.SwipeActions
 import net.thunderbird.core.featureflag.FeatureFlagProvider
 import net.thunderbird.core.featureflag.toFeatureFlagKey
+import net.thunderbird.core.logging.LogLevel
+import net.thunderbird.core.logging.LogLevelManager
 import net.thunderbird.core.logging.composite.CompositeLogSink
 import net.thunderbird.core.logging.file.FileLogSink
 import net.thunderbird.core.preference.storage.Storage
@@ -35,6 +37,7 @@ object K9 : KoinComponent {
     private val featureFlagProvider: FeatureFlagProvider by inject()
     private val syncDebugCompositeSink: CompositeLogSink by inject(named("syncDebug"))
     private val syncDebugFileLogSink: FileLogSink by inject(named("syncDebug"))
+    private val loggerManager: LogLevelManager by inject()
 
     /**
      * If this is `true`, various development settings will be enabled.
@@ -413,7 +416,10 @@ object K9 : KoinComponent {
     private fun updateLoggingStatus() {
         Timber.uprootAll()
         if (isDebugLoggingEnabled) {
+            loggerManager.override(LogLevel.VERBOSE)
             Timber.plant(DebugTree())
+        } else {
+            loggerManager.restoreDefault()
         }
     }
 


coreycb:
I have similar questions to the other uplift. Has this been tested on daily? It also hasn't made it to beta yet, but it will be released in 13.0b1 next Tues. The next chance to uplift to release would be 12.1 on Sep 15. If this rides the train it would get released on Sep 29 in 13.0.

rafaeltonholo:
Tested on daily build 13.0a1 (22401652). See screen recording below:

https://github.com/user-attachments/assets/37213d7c-3de7-4709-9f65-0c0cc069e67b

coreycb:
Thanks for testing and sharing the screen recording. This is an important fix. Not having verbose logging is going to make it harder for developers to triage bugs. However, I'm concerned that it hasn't spent time on beta where it would be exposed to a larger number of users, and it landed on main 2-3 days ago. Let's give this some exposure on beta to make sure there are no regressions introduced. It will be included in 13.0b1 on Tues, Sept 2nd. Hopefully we can make it without verbose logging for the 2 weeks after that, and we can plan to uplift this to 12.1. That's scheduled for Mon, Sept 15th. Let's leave the PR open until then.

rafaeltonholo:
> Thanks for testing and sharing the screen recording. This is an important fix. Not having verbose logging is going to make it harder for developers to triage bugs. However, I'm concerned that it hasn't spent time on beta where it would be exposed to a larger number of users, and it landed on main 2-3 days ago. Let's give this some exposure on beta to make sure there are no regressions introduced. It will be included in 13.0b1 on Tues, Sept 2nd. Hopefully we can make it without verbose logging for the 2 weeks after that, and we can plan to uplift this to 12.1. That's scheduled for Mon, Sept 15th. Let's leave the PR open until then.

Sounds good to me

krzysdz:
I think this PR is missing something. I can't find `logLevelManager.override()` call anywhere other than in `.restoreDefault()`. The change to [`DefaultDebuggingSettingsPreferenceManager.kt`](https://github.com/thunderbird/thunderbird-android/pull/9634/files#diff-9155b98a3c8a5ec627cecf4bf2b72f8a46c7dbd8a31368b723632b6af813d661) from #9634 seems to be missing, because this file doesn't exist in the `release` branch.

rafaeltonholo:
> I think this PR is missing something. I can't find `logLevelManager.override()` call anywhere other than in `.restoreDefault()`. The change to [`DefaultDebuggingSettingsPreferenceManager.kt`](https://github.com/thunderbird/thunderbird-android/pull/9634/files#diff-9155b98a3c8a5ec627cecf4bf2b72f8a46c7dbd8a31368b723632b6af813d661) from #9634 seems to be missing, because this file doesn't exist in the `release` branch.

Good catch! As the `release` branch still doesn't have the `DefaultDebuggingSettingsPreferenceManager`, I will need to introduce a commit to make the log level change in the `K9` object.

Thanks for pointing that out!

legacy/core/src/main/java/com/fsck/k9/K9.kt
```diff
@@ -413,7 +416,10 @@ object K9 : KoinComponent {
     private fun updateLoggingStatus() {
         Timber.uprootAll()
         if (isDebugLoggingEnabled) {
+            loggerManager.override(LogLevel.DEBUG)
```
krzysdz: While it's just uplift of the same change from #9634 and less of an issue in `release`, this override technically **decreases verbosity** in debug builds, which have `LogLevel.VERBOSE` as default.

rafaeltonholo: In debug builds, we log at the `VERBOSE` level by default. Changing the log level to `VERBOSE` when "Enable debug logging" is selected will not alter this behaviour, as it remains the same. However, setting it to `DEBUG` and reducing the verbosity in debug mode allows us to confirm that the option is functioning as intended, without needing to build the app in release mode.

krzysdz: There are some verbose logs that are emitted only when "Enable debug logging" is selected and will never be shown, even in debug builds, because of minimum `DEBUG` level when this setting is enabled:

- Guarded by `K9MailLib.isDebug()`:
  - 6 in `RealImapConnection.kt`
  - 1 in `XOAuth2ChallengeParser.java`
  - 1 in `ImapResponseParser.java`
- [`StorageEditor.putStringWithLogging`](https://github.com/thunderbird/thunderbird-android/blob/a3e80598c5ecbeb9fc1fc86ed83f7ad3b3e50880/legacy/core/src/main/java/com/fsck/k9/preferences/GeneralSettingsWriter.kt#L40-L59) if `isDebugLoggingEnabled` argument is `true` (all callers get this from `generalSettingsManager.getConfig().debugging.isDebugLoggingEnabled`)
- Directly controlled by `generalSettingsManager.getConfig().debugging.isDebugLoggingEnabled`:
  - [`LockableDatabase`](https://github.com/thunderbird/thunderbird-android/blob/a3e80598c5ecbeb9fc1fc86ed83f7ad3b3e50880/legacy/core/src/main/java/com/fsck/k9/mailstore/LockableDatabase.java#L144-L169)
  - [`SettingsUpgradeHelper`](https://github.com/thunderbird/thunderbird-android/blob/a3e80598c5ecbeb9fc1fc86ed83f7ad3b3e50880/legacy/core/src/main/java/com/fsck/k9/preferences/SettingsUpgradeHelper.kt#L99-L102)

rafaeltonholo: Oh, I see your point now. I wasn't aware of the `K9MailLib.isDebug()`. Then what appears to me is that when enabling the "Enable debug logging", we should set the `LogLevel` to `VERBOSE` and not to `DEBUG`. I will apply the change accordingly.

rafaeltonholo: I will also create an issue to apply that change in the main branch. Thanks for bringing that to our attention

wmontwe (approved):
LGTM

rafaeltonholo:
The uplift won't be necessary as a new release is coming soon.

PR #9679, opened 2025-08-28
chore(ci): fix cache use of the android workflow

### LINKED ISSUE #9680: Android workflow doesn't use cache
I observed that the android workflow is not consuming the Gradle cache produced by the
gradle-cache workflow. This is because of GitHub changes on how caches work, see: [Using the cache read-only](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#using-the-cache-read-only) and [Restrictions for accessing a cache](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache)

As the android workflow is never excecuted on the main thread it will never write to the Gradle cache. We need to fix this.

wmontwe:
This should fix the caching issue for the android workflow by allowing execution on the main branch to fill the cache.

See: [Using the cache read-only](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#using-the-cache-read-only) and [Restrictions for accessing a cache](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache)

Results should be visible once this got merged.

Resolves #9680

### DIFF
--- a/.github/ci-gradle.properties
+++ b/.github/ci-gradle.properties
@@ -1,7 +1,7 @@
 org.gradle.daemon=false
 org.gradle.parallel=true
 org.gradle.workers.max=4
-org.gradle.jvmargs=-Xmx10g -XX:MaxMetaspaceSize=1g -Dfile.encoding=UTF-8 -XX:+UseParallelGC -XX:+HeapDumpOnOutOfMemoryError
+org.gradle.jvmargs=-Xmx10g -XX:MaxMetaspaceSize=2g -Dfile.encoding=UTF-8 -XX:+UseParallelGC -XX:+HeapDumpOnOutOfMemoryError
 
-kotlin.incremental=false
+kotlin.incremental=true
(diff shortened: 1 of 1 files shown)


.github/workflows/android.yml
```diff
@@ -1,6 +1,9 @@
 name: Android CI
 
 on:
+  push:
+    branches:
+      - main
   pull_request:
```
coreycb, line 7: Wouldn't this already get run on pull requests to main without this change?

wmontwe: The `pull_request` workflow runs against the pull request branch, and all caches are tied to that branch. However, we need this workflow to run on commits to `main` (merge commits), since the build cache can only be written during direct executions on main. The previous approach of using a separate workflow to populate the build cache is no longer viable because of GitHub’s restrictions on cache access.

coreycb (approved):

PR #9695, opened 2025-08-30
Fix #9694 - Fix the misleading label for "Your Name" field

### LINKED ISSUE #9694: Confusing (possibly misleading) label in "Account Display Options" for en
### Checklist

### App

Thunderbird for Android

### App version

11.1

Google Play

### Android version

14

### Device model

Oneplus

### Steps to reproduce

1. Set language as English - GB in your phone settings
2. Setup any GMail account in Thunderbird
3. After OAuth, we see a Display Options screen that has a field "Display Name"
4. The default value in other languages seems to be "Your Name"

### Expected behavior

The field should have "Your name" in en-rGB language too.

### Actual behavior

The field has a label "Display Name". (See Screenshot Below)

This is confusing to any new user.  

They may mistake it for "Display name" of the account and give something like "My Personal GMail".  Now, all the emails they send from Thunderbird will go from a person named "My Personal GMail"

Image

### Logs

_No response_

vivganes:
This PR fixes #9694

### DIFF
--- a/feature/account/setup/src/main/res/values-en-rGB/strings.xml
+++ b/feature/account/setup/src/main/res/values-en-rGB/strings.xml
@@ -22,10 +22,10 @@
     <string name="account_setup_options_account_name_label">Account name</string>
     <string name="account_setup_auto_discovery_status_header_title_configuration_found">Configuration found</string>
     <string name="account_setup_auto_discovery_validation_error_email_address_required">Email address is required.</string>
-    <string name="account_setup_options_display_name_label">Display name</string>
+    <string name="account_setup_options_display_name_label">Your name</string>
     <string name="account_setup_auto_discovery_loading_error">Failed to load email configuration</string>
     <string name="account_setup_auto_discovery_result_approval_checkbox_label">I trust this configuration</string>
-    <string name="account_setup_options_display_name_error_required">Display name is required.</string>
+    <string name="account_setup_options_display_name_error_required">Your name is required.</string>
     <string name="account_setup_create_account_error">An error occurred while trying to create the account</string>
     <string name="account_setup_error_network">Network error. Please check your connection status and try again.</string>
     <string name="account_setup_auto_discovery_validation_error_email_address_invalid">This is not recognised as a valid email address.</string>


feature/account/setup/src/main/res/values-en-rGB/strings.xml
```diff
@@ -22,10 +22,10 @@
     <string name="account_setup_options_account_name_label">Account name</string>
     <string name="account_setup_auto_discovery_status_header_title_configuration_found">Configuration found</string>
     <string name="account_setup_auto_discovery_validation_error_email_address_required">Email address is required.</string>
-    <string name="account_setup_options_display_name_label">Display name</string>
```
wmontwe, line 25: `account_setup_options_display_name_label` -> `account_setup_options_your_name_label`

feature/account/setup/src/main/res/values-en-rGB/strings.xml
```diff
@@ -22,10 +22,10 @@
...
+    <string name="account_setup_options_display_name_label">Your name</string>
     <string name="account_setup_auto_discovery_loading_error">Failed to load email configuration</string>
     <string name="account_setup_auto_discovery_result_approval_checkbox_label">I trust this configuration</string>
-    <string name="account_setup_options_display_name_error_required">Display name is required.</string>
```
wmontwe, line 28: `account_setup_options_display_name_error_required` -> `account_setup_options_your_name_error_required`

wmontwe (changes requested):
Thank you for the fix. Since the meaning of the string has changed, the string key should also be renamed to provide translators with the proper context.

feature/account/setup/src/main/res/values-en-rGB/strings.xml
vivganes: Done!

feature/account/setup/src/main/res/values-en-rGB/strings.xml
vivganes: Done

wmontwe (approved):
Thank you!

I trimmed your changes down to the `en` strings, since that’s what Weblate uses as the basis for translations. No further changes are needed.

vivganes:
@wmontwe - I am seeing an error in CI.  Looks like having extra strings causes a build failure.

```
"account_setup_options_display_name_label" is translated here but not found in default locale [ExtraTranslation]
```

wmontwe:
> @wmontwe - I am seeing an error in CI. Looks like having extra strings causes a build failure.
> 
> ```
> "account_setup_options_display_name_label" is translated here but not found in default locale [ExtraTranslation]
> ```

I fixed it, our instructions on how to deal with `en` source strings need an update.

PR #9743, opened 2025-09-07
fix: remove Gmail prefix when updating folder list

### LINKED ISSUE #9742: [Gmail]/ prefix appears after refreshing folder list
### Checklist

### App

Thunderbird for Android

### App version

283fada2f39b97f05398352c7bb04e13433d975a (`main` branch as of 2025-09-07)

Other

### Android version

13 (xiaomi.eu 14.0.3.0.TJUMIXM)

### Device model

Xiaomi POCO X3 Pro (global)

### Steps to reproduce

1. Add 2 accounts, 1 of them Gmail
2. Refresh folders on the Gmail account (Manage folders -> 3 dots -> Refresh folder list) (this updates folders to include the `[Gmail]/` prefix) - **The bug is here**
3. Switch to a different account (can be Unified; causes the app to regenerate folder tree)
4. Switch to the Gmail account again - there should be a _[Gmail]_ [placeholder folder](https://github.com/thunderbird/thunderbird-android/blob/94c9063858f999cc4e57538ec0eb2fcd0c35a124/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/domain/usecase/GetDisplayTreeFolder.kt#L73)

### Expected behavior

Folders list on the Gmail account is the same as when the app was opened - flat with no _[Gmail]_ folder.

### Actual behavior

A _[Gmail]_ folder appears and folders other than _Inbox_ and _Outbox_ are inside it. In "Manage folders" those folders are prefixed with `[Gmail]`.

### Additional details

The problem was originally reported @platform34 by in #9644. The crash when clicking the `[Gmail]` folder was resolved in #9716, but why the prefix appears remained an unsolved problem.

#9213 added code that removes `[Gmail]/` and `[Google Mail]/` prefixes from folders when creating the message store and prevents creation of new folders with those prefixes in `CreateFolderOperations.createFolders()`. When refreshing folder list if a folder with the same `serverId` already exists `UpdateFolderOperations.changeFolder()` is used to update the folder's name and type. `UpdateFolderOperations` does not sanitise the new name causing the prefix to appear.

### Logs

_No response_

krzysdz:
The missing part of #9213.

Fixes #9742

### DIFF
--- a/legacy/storage/src/main/java/com/fsck/k9/storage/messages/UpdateFolderOperations.kt
+++ b/legacy/storage/src/main/java/com/fsck/k9/storage/messages/UpdateFolderOperations.kt
@@ -11,7 +11,7 @@ internal class UpdateFolderOperations(private val lockableDatabase: LockableData
     fun changeFolder(folderServerId: String, name: String, type: FolderType) {
         lockableDatabase.execute(false) { db ->
             val values = ContentValues().apply {
-                put("name", name)
+                put("name", name.replace("\\[(Gmail|Google Mail)]/".toRegex(), ""))
                 put("type", type.toDatabaseFolderType())
             }
 


asoucar:
Thank you for fixing this!

Issue #9654, opened 2025-08-26
Replace MessageList items with new UI components


wmontwe:
The new Composable MessageList item UI should replace the XML-based layouts.

PR #9770, opened 2025-09-10
Refactor message list fragment

wmontwe:
Depends on #9763

Preparation to for the new message list item ui #9654


legacy/core/src/main/java/com/fsck/k9/mailstore/LocalStoreProvider.kt
```diff
@@ -17,18 +19,36 @@ class LocalStoreProvider {
         val generalSettingsManager = DI.get(GeneralSettingsManager::class.java)
         val accountUuid = account.uuid
 
+        return getInstanceById(accountUuid) {
+            LocalStore.createInstance(account, context, generalSettingsManager)
+        }
+    }
+
+    @Throws
```
rafaeltonholo: Please add what kind of exception this method throws:
```suggestion
    @Throws(MessagingException::class)
```

wmontwe:
@rafaeltonholo I fixed some issues by removing the generic class injections which should help koin to inject the right implementation.

But still need to fix the archive folder creation.

wmontwe:
I updated the dependency injection to remove the star-projections as koin is not capable to inject the right dependency at runtime. Neither is the `DependencyInjectionTest` able to detect those issues.

This required a larger effort and I moved feature specific injections to `app-common`.

app-common/src/test/kotlin/net/thunderbird/app/common/feature/mail/BaseAccountBackendStorageFactoryTest.kt
```diff
@@ -0,0 +1,56 @@
...
+        val dto = FakeData.legacyAccountDto
+        val legacyFactory = FakeLegacyAccountDtoBackendStorageFactory()
+        val mapper = FakeLegacyAccountDataMapper()
+        val factory = BaseAccountBackendStorageFactory(legacyFactory, mapper)
+
+        // Act
+        factory.createBackendStorage(dto)
+
+        // Assert
+        assertThat(legacyFactory.lastAccount).isSameAs(dto)
```
rafaeltonholo: To avoid compile error when updating assertk dependency, please use the suggested replacement instead.

```suggestion
        assertThat(legacyFactory.lastAccount).isSameInstanceAs(dto)
```

Please, apply the same for the other parts of the code that uses `isSame(instance)` instead.

app-common/src/test/kotlin/net/thunderbird/app/common/feature/mail/BaseAccountImapBackendFactoryTest.kt
```diff
@@ -0,0 +1,110 @@
...
+        val account = createLegacyAccountDto(protocol = Protocols.IMAP)
+        val legacyFactory = FakeLegacyBackendFactory()
+        val mapper = FakeLegacyAccountDataMapper().apply { toDtoResult = account }
+        val factory = BaseAccountImapBackendFactory(legacyFactory, mapper)
+
+        // Act
+        factory.createBackend(account)
+
+        // Assert
+        assertThat(legacyFactory.lastAccount).isSameAs(account)
```
rafaeltonholo: To avoid compile error when updating assertk dependency, please use the suggested replacement instead.

```suggestion
        assertThat(legacyFactory.lastAccount).isSameInstanceAs(account)
```

Please, apply the same for the other parts of the code that uses `isSame(instance)` instead.

app-common/src/test/kotlin/net/thunderbird/app/common/feature/mail/BaseAccountSpecialFolderUpdaterFactoryTest.kt
```diff
@@ -0,0 +1,108 @@
...
+        val dto = createLegacyAccountDto()
+        val legacyFactory = FakeLegacyAccountDtoSpecialFolderUpdaterFactory()
+        val mapper = FakeLegacyAccountDataMapper().apply { toDtoResult = dto }
+        val factory = BaseAccountSpecialFolderUpdaterFactory(legacyFactory, mapper)
+
+        // Act
+        val updater = factory.create(dto)
+
+        // Assert
+        assertThat(legacyFactory.lastAccount).isSameAs(dto)
```
rafaeltonholo: To avoid compile error when updating assertk dependency, please use the suggested replacement instead.

```suggestion
        assertThat(legacyFactory.lastAccount).isSameInstanceAs(dto)
```

Please, apply the same for the other parts of the code that uses `isSame(instance)` instead.

app-common/src/test/kotlin/net/thunderbird/app/common/feature/mail/RemoteFolderCreatorResolverTest.kt
```diff
@@ -0,0 +1,44 @@
...
+        val account = FakeData.legacyAccountDto.apply {
+            incomingServerSettings = incomingServerSettings.copy(type = Protocols.IMAP)
+            outgoingServerSettings = outgoingServerSettings.copy(type = Protocols.IMAP)
+        }
+
+        // Act
+        val result = router.create(account)
+
+        // Assert
+        assertThat(result).isNotSameAs(NoOpRemoteFolderCreator)
```
rafaeltonholo: To avoid compile error when updating assertk dependency, please use the suggested replacement instead.

```suggestion
        assertThat(result).isNotSameInstanceAs(NoOpRemoteFolderCreator)
```

Please, apply the same for the other parts of the code that uses `isNotSameAs(instance)` instead.

core/android/account/src/main/kotlin/net/thunderbird/core/android/account/LegacyAccountDtoManager.kt
```diff
@@ -11,7 +11,7 @@ import net.thunderbird.feature.mail.account.api.AccountManager
         "app.k9mail.legacy.account.LegacyAccount",
     ),
 )
-interface AccountManager : AccountManager<LegacyAccountDto> {
+interface LegacyAccountDtoManager : AccountManager<LegacyAccountDto> {
```
rafaeltonholo, line 13: I know that is not related with this PR's changes, but could you please update the `replaceWith` expression to use the correct replacement?

```kotlin
@Deprecated(
    message = "Use net.thunderbird.feature.mail.account.api.AccountManager<TAccount : BaseAccount> instead",
    replaceWith = ReplaceWith(
        expression = "AccountManager<LegacyAccountDto>",
        "net.thunderbird.feature.mail.account.api.AccountManager",
        "net.thunderbird.core.android.account.LegacyAccountDto",
    ),
)
```

legacy/core/src/main/java/com/fsck/k9/controller/MessagingControllerWrapper.kt
```diff
@@ -0,0 +1,217 @@
...
+import net.thunderbird.core.android.account.LegacyAccountDto
+import net.thunderbird.core.android.account.LegacyAccountDtoManager
+import net.thunderbird.feature.account.AccountId
+
+/**
+ * A wrapper around [MessagingController] that takes care of loading the account by [AccountId] and
+ * provides some convenience methods.
+ */
+@Suppress("TooManyFunctions")
+class MessagingControllerWrapper(
```
rafaeltonholo, line 16: suggestion: I recommend dividing this class into smaller, contextualized classes to facilitate better decoupling in the future. These can then be composed together using the constructor. This approach would also help eliminate the need to suppress `TooManyFunctions`.

I would also consider renaming `MessagingControllerWrapper` to `MessagingFacade`, as it acts like one. The constructor would look like this:

```kotlin
class MessagingFacade(
    private val capabilityService: CapabilityService,
    private val syncService: SyncService,
    private val searchService: SearchService,
    private val flagService: FlagService,
    private val refileService: RefileService, // Or MoveCopyService
    private val deletionService: MessageDeletionService,
    private val archiveService: MessageArchiveService,
) { ... }

interface CapabilityService {
    fun isMoveCapable(id: AccountId): Boolean
    fun isCopyCapable(id: AccountId): Boolean
    fun supportsExpunge(id: AccountId): Boolean
    // ...
}

interface SyncService {
    fun synchronizeMailbox(id: AccountId, folderId: Long, notify: Boolean, listener: MessagingListener?)
    // ...
}

interface SearchService {
    fun searchRemoteMessages(
        id: AccountId,
        folderId: Long,
        query: String?,
        requiredFlags: Set<Flag>?,
        forbiddenFlags: Set<Flag>?,
        listener: MessagingListener,
    ): Future<*>?
    // ...
}

// ... other interfaces.
```

Not required for this PR, but I think it would be beneficial to start thinking about this. We can also consider to make the split into another issue.

legacy/core/src/main/java/com/fsck/k9/controller/MessagingControllerWrapper.kt
```diff
@@ -0,0 +1,217 @@
...
+        val account = getAccountDtoOrThrow(id)
+        return messagingController.isPushCapable(account)
+    }
+
+    fun markAllMessagesRead(id: AccountId, folderId: Long) {
+        val account = getAccountDtoOrThrow(id)
+        messagingController.markAllMessagesRead(account, folderId)
+    }
+
+    fun checkAuthenticationProblem(id: AccountId) = {
```
rafaeltonholo: Was the intention to return a lambda or just a typo?

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
```diff
@@ -852,19 +855,23 @@ class MessageListFragment :
...
 
-            accountManager.saveAccount(account)
+            this.sortDateAscending = newSortAscendingMap[SortType.SORT_DATE] ?: SortType.SORT_DATE.isDefaultAscending
+
+            val updatedAccount = account.copy(
+                sortType = sortType,
+                sortAscending = newSortAscendingMap,
+            )
+            accountManager.saveAccount(updatedAccount)
+            this.account = updatedAccount
```
rafaeltonholo: Could we launch this within a coroutine? This will perform IO operation as it needs to write to the preferences, and I don't we are currently running it outside the main thread. If so, disconsider this comment.

```suggestion
            lifecycleScope.launch(ioDispatcher) {
                accountManager.saveAccount(updatedAccount)
                this.account = updatedAccount
            }
```

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
```diff
@@ -1317,13 +1324,23 @@ class MessageListFragment :
...
         }
     }
 
+    private fun setLastSelectedFolder(messages: List<MessageReference>, folderId: Long) {
+        val firstMessage = messages.firstOrNull() ?: return
+        val account = accountManager.getAccount(firstMessage.accountUuid) ?: return
+        accountManager.saveAccount(
+            account.copy(
+                lastSelectedFolderId = folderId,
+            ),
```
rafaeltonholo, line 1348: Shouldn't this be applied to `this.account` as well?

```suggestion
            account.copy(
                lastSelectedFolderId = folderId,
            ).also { this.account = it },
```

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
```diff
@@ -1409,26 +1426,46 @@ class MessageListFragment :
             .groupBy { it.folderId }
 
         for ((folderId, messagesInFolder) in folderMap) {
-            val account = accountManager.getAccount(messagesInFolder.first().accountUuid)
+            val account = accountManager.getAccount(messagesInFolder.first().accountUuid) ?: continue
```
rafaeltonholo: suggestion: I would consider adding a log warning that we couldn't find the account before the `continue`, so we could leave trace in case for an investigation.

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
```diff
@@ -1409,26 +1426,46 @@ class MessageListFragment :
...
                 } else {
-                    messagingController.copyMessages(account, folderId, messagesInFolder, destinationFolderId)
+                    messagingController.copyMessages(
+                        account.id,
+                        folderId,
+                        messagesInFolder,
+                        destinationFolderId,
+                    )
                 }
             }
```
rafaeltonholo: nit: You could use when with guards to remove the nesting if/else here:

```suggestion
            when (operation) {
                FolderOperation.MOVE if showingThreadedList -> messagingController.moveMessagesInThread(
                    account.id,
                    folderId,
                    messagesInFolder,
                    destinationFolderId,
                )
                FolderOperation.MOVE -> messagingController.moveMessages(
                    account.id,
                    folderId,
                    messagesInFolder,
                    destinationFolderId,
                )
                FolderOperation.COPY if showingThreadedList -> messagingController.copyMessagesInThread(
                    account.id,
                    folderId,
                    messagesInFolder,
                    destinationFolderId,
                )
                FolderOperation.COPY -> messagingController.copyMessages(
                    account.id,
                    folderId,
                    messagesInFolder,
                    destinationFolderId,
                )
            }

```

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
```diff
@@ -1688,7 +1729,7 @@ class MessageListFragment :
         messageListItems
             .map { it.account }
             .toSet()
-            .forEach(messagingController::checkAuthenticationProblem)
+            .forEach { account -> messagingController.checkAuthenticationProblem(account.id) }
```
rafaeltonholo, line 1747: The `messagingController.checkAuthenticationProblem(id)` is returning a lambda. If that was the intention, you need to trigger it here as well, otherwise this won't run.

```suggestion
            .forEach { account -> messagingController.checkAuthenticationProblem(account.id)() }
```

legacy/core/src/main/java/com/fsck/k9/controller/MessagingControllerWrapper.kt
rafaeltonholo: See also https://github.com/thunderbird/thunderbird-android/pull/9770#discussion_r2355175868

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
wmontwe: That was a typo.

legacy/core/src/main/java/com/fsck/k9/controller/MessagingControllerWrapper.kt
wmontwe: Good idea, but I think this is better done as separate task.

legacy/core/src/main/java/com/fsck/k9/controller/MessagingControllerWrapper.kt
wmontwe: That's a typo.

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
wmontwe: This is taken from the `MlfUtils` and the account wasn't updated there. Don't know what the side effects will be. From a first glance the `lastSelectedFolderId` is not relevant for the `MessageListFragment`.

rafaeltonholo: Oh, I see. I was under the impression that the `account` was also being added as a parameter in the `MlfUtils`. You can disregard this, though.

rafaeltonholo (approved):
LGTM, thanks for addressing all the comments!!

PR #9805, opened 2025-09-16
ci(dependabot): add workflow to automatically update dependency guard when dependabot creates a PR

rafaeltonholo:
Whenever `@dependabot` updates an Android version, the app will challenge the new version with the dependency guard, and the CI will fail.

This PR creates a workflow that, whenever `@dependabot` opens, reopens or updates a PR, this workflow will trigger and update the dependency guard if required.

### DIFF
--- a/.github/workflows/dependabot-dependency-guard-update.yml
+++ b/.github/workflows/dependabot-dependency-guard-update.yml
@@ -1,0 +1,50 @@
+name: dependabot PR - Update dependency guard
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+    branches:
+      - main
+    paths:
+      - '**/*.gradle'
+      - 'gradle/**'
+      - 'gradlew'
+      - '**/libs.versions.toml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event.pull_request.number || 'no-pr' }}
+  cancel-in-progress: true
+
+jobs:
+  update-dependency-guard:
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    steps:
+      - name: Checkout the repo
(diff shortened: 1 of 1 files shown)


wmontwe (dismissed):
LGTM

coreycb:
I don't think the git push will work because the default permissions for actions on this repo don't allow write. I would hope those settings apply to PR branches as well. I think we should be reducing the "permissions:" of the GITHUB_TOKEN to at least read, and instead use actions/create-github-app-token.

rafaeltonholo:
> I don't think the git push will work because the default permissions for actions on this repo don't allow write. I would hope those settings apply to PR branches as well. I think we should be reducing the "permissions:" of the GITHUB_TOKEN to at least read, and instead use actions/create-github-app-token.

Since this only pushes to the pull request's branch, I don't think it will require any special permissions. I reviewed the documentation again and realized that I had missed adding the `ref` parameter to the `action/checkout`. This is likely going to cause an issue, as it [checks out in detached HEAD mode](https://github.com/actions/checkout?tab=readme-ov-file#push-a-commit-to-a-pr-using-the-built-in-token).

Aside from that adjustment, I believe we should be able to commit and push without any issues. I don't think we need a custom token for this, as it still requires approval as part of the pull request process.

coreycb:
> > I don't think the git push will work because the default permissions for actions on this repo don't allow write. I would hope those settings apply to PR branches as well. I think we should be reducing the "permissions:" of the GITHUB_TOKEN to at least read, and instead use actions/create-github-app-token.
> 
> Since this only pushes to the pull request's branch, I don't think it will require any special permissions. I reviewed the documentation again and realized that I had missed adding the `ref` parameter to the `action/checkout`. This is likely going to cause an issue, as it [checks out in detached HEAD mode](https://github.com/actions/checkout?tab=readme-ov-file#push-a-commit-to-a-pr-using-the-built-in-token).
> 
> Aside from that adjustment, I believe we should be able to commit and push without any issues. I don't think we need a custom token for this, as it still requires approval as part of the pull request process.

I agree and the permissions should be reduced to read already based on the repo settings. LGTM

coreycb (approved):

PR #9814, opened 2025-09-17
fix(ci): dependabot workflow update dependency guard task

rafaeltonholo:
Removed unnecessary leading `:` from gradle task command


wmontwe (approved):
LGTM

PR #9820, opened 2025-09-18
Opt-in to Arm MTE for debug and daily

thgoebel:
This is an easy security improvement (at least for users with Google Pixel 8 and later, which have the hardware for MTE).

TFA doesn't have any explicit native code afaik, but there are some .so files in the APK (from Compose?), and also Android framework code that TFA calls might have native code. All of that runs in the app's process, so opting in to MTE will cover that. TFA parses attacker controlled data (emails), so enabling MTE is an easy defense-in-depth.

I've been running TFA on a Pixel 8a with GrapheneOS and MTE force-enabled for a few months now without any issues.
Still, I recommend you to test this again on an MTE-compatible device (Pixel 8 and later), just to be sure :)

For background on MTE, see:

- Android docs: https://source.android.com/docs/security/test/memory-safety/arm-mte
- My blog post :innocent: : https://thore.io/posts/2025/09/introduction-to-arm-memory-tagging-extensions/

### DIFF
--- a/app-thunderbird/src/daily/AndroidManifest.xml
+++ b/app-thunderbird/src/daily/AndroidManifest.xml
@@ -4,7 +4,9 @@
     xmlns:tools="http://schemas.android.com/tools"
     >
 
-    <application tools:ignore="MissingApplicationIcon">
+    <application
+        tools:ignore="MissingApplicationIcon"
+        android:memtagMode="async">
 
         <!-- This component is disabled by default (if possible). It will be enabled programmatically if necessary. -->
         <!-- IMPORTANT: The component name must be -->
--- a/app-thunderbird/src/debug/AndroidManifest.xml
+++ b/app-thunderbird/src/debug/AndroidManifest.xml
@@ -4,7 +4,9 @@
     xmlns:tools="http://schemas.android.com/tools"
     >
 
-    <application tools:ignore="MissingApplicationIcon">
+    <application
+        tools:ignore="MissingApplicationIcon"
+        android:memtagMode="async">
 
         <!-- This component is disabled by default (if possible). It will be enabled programmatically if necessary. -->
         <!-- IMPORTANT: The component name must be -->


asoucar:
Thanks for bringing this up. @wmontwe @rafaeltonholo I'd like to add this to a meeting or thread to discuss since security related work should probably get more eyes.

asoucar:
Thanks for bringing this to our attention. After some discussion with @wmontwe we don't currently have the testing bandwidth to safely enable this.  We believe this should be fairly extensively tested before we'd be comfortable including the MTE changes in the app. I'm going to leave this PR up so that once we feel we can sufficiently test that it is top of mind.

wmontwe:
@thgoebel We don't have test devices with GrapheneOS to verify these changes. But we could add this setting to the develop and daily version of Thunderbird. It would require to update the manifest files within `app-thunderbird/src/debug/` and `app-thunderbird/src/daily/` instead of the one within `app-common/src/main`.

If there are any volunteers in the community with GrapheneOS, they could help by testing our daily version for compatibility, once this patch was merged. If there is positive feedback we could enable it for all versions.

What do you think?

thgoebel:
Sounds good, I moved the manifest tag to debug and daily.

Note that with this change, you don't need GrapheneOS, you only need a Pixel 8 or later. This change works on any ROM, including stock.

GrapheneOS is only relevant because it allows users to force-enable MTE for apps where the developer has not opted in via the manifest. This is how I'm using all my apps (including Thunderbird).

In the past year I have seen basically no crashes, only in two apps (Proton Mail and Apple Music).
Fwiw, an MTE-induced crash is a memory bug that should be fixed anyway. MTE just makes those bugs visible, instead of silently corrupting the memory (and potentially crashing much later due to to that initial corrupution).

TLDR: By opting in via the manifest, any community member with a Pixel 8 or later can test MTE (on debug/daily), no matter which ROM they are using. Rolling it out to daily first (and later beta?) makes sense.

wmontwe:
> Sounds good, I moved the manifest tag to debug and daily.
> 
> Note that with this change, you don't need GrapheneOS, you only need a Pixel 8 or later. This change works on any ROM, including stock.
> 
> GrapheneOS is only relevant because it allows users to force-enable MTE for apps where the developer has not opted in via the manifest. This is how I'm using all my apps (including Thunderbird).
> 
> In the past year I have seen basically no crashes, only in two apps (Proton Mail and Apple Music). Fwiw, an MTE-induced crash is a memory bug that should be fixed anyway. MTE just makes those bugs visible, instead of silently corrupting the memory (and potentially crashing much later due to to that initial corrupution).
> 
> TLDR: By opting in via the manifest, any community member with a Pixel 8 or later can test MTE (on debug/daily), no matter which ROM they are using. Rolling it out to daily first (and later beta?) makes sense.

Thanks for the detailed explanation. Then I could use my Pixel 9 to test daily. When this results in a stable experience, we could enable MTE for beta and test with a larger audience.

I created a ticket to keep track of these changes: #10053

PR #9836, opened 2025-09-22
fix(message-list): message list skipping first item and load more footer also doesn't show up

### LINKED ISSUE #9819: Regression: MessageList doesn't show "load more messages" anymore
When navigating to a folder you have not downloaded before, there should be a list item "Load up to xxx more" and when clicked should trigger the download of the messages.

Expected:

Image

Actual:

Image

rafaeltonholo:
Fixes #9819.

This PR address two issues:

1. The "Load up to 100 more" footer not showing up
2. The Message List was skipping the first email from the list

Additionally, it is also removing a background colour from the `banner_global_compose_view` that was left over.


asoucar:
Found a bug in this. Pull to refresh then scroll to the bottom. Each refresh adds bottom rows
| 1 refresh: | 3 refreshes: |
| --------- | ------------ |
| Screenshot_20250922_145137 | Screenshot_20250922_145251 |

rafaeltonholo:
> Found a bug in this. Pull to refresh then scroll to the bottom. Each refresh adds bottom rows
> | 1 refresh: | 3 refreshes: |
> | ----- | ----- |
> | Screenshot_20250922_145137 | Screenshot_20250922_145251 |

Thanks for pointing that! I was able to reproduce and fix the issue!

PR #9837, opened 2025-09-22
chore(migration-v90): add extra where clause to ensure the outbox folder or local folder's won't change

rafaeltonholo:
This PR adds more clauses to the WHERE clause in the Migration to version 90 to ensure that the migration does not touch any local folders or the Outbox folder.


wmontwe (approved):
LGTM

PR #9854, opened 2025-09-25
feat(notifications): add capability to dismiss a notification using NotificationDismisser or NotificationManager

rafaeltonholo:
Part of #9626.

- Enable In-app notifications to be dismissed via `NotificationDismisser` or `NotificationManager`. 
- Fix some in-app notifications not displaying after publish being called
- `AuthenticationErrorNotification` not taking in consideration non-OAuth accounts.


feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/receiver/InAppNotificationEventBus.kt
```diff
@@ -22,7 +22,7 @@ internal interface InAppNotificationEventBus : InAppNotificationReceiver {
 }
 
 internal fun InAppNotificationEventBus(): InAppNotificationEventBus = object : InAppNotificationEventBus {
-    private val _events = MutableSharedFlow<InAppNotificationEvent>(replay = 1)
+    private val _events = MutableSharedFlow<InAppNotificationEvent>(replay = 32)
```
wmontwe, line 25: Is this number related to something?

feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/command/DismissInAppNotificationCommand.kt
```diff
@@ -0,0 +1,27 @@
...
+    notifier: NotificationNotifier<InAppNotification>,
+) : DismissNotificationCommand<InAppNotification>(
+    logTag = TAG,
+    logger = logger,
+    featureFlagProvider = featureFlagProvider,
+    notificationRegistry = notificationRegistry,
+    notification = notification,
+    notifier = notifier,
+) {
+    override val featureFlagKey: FeatureFlagKey = FeatureFlagKey.DisplayInAppNotifications
```
wmontwe, line 26: Ideally the key is not defined as part of the feature flag module, but feature specific. Example:

```kotlin
object NotificationFeatureFlags {
    val DisplayInAppNotifications: FeatureFlagKey =
        "display_in_app_notifications".toFeatureFlagKey()

    val UseNotificationSenderForSystemNotifications: FeatureFlagKey =
        "use_notification_sender_for_system_notifications".toFeatureFlagKey()
} 
```

This avoids coupling the feature flag module to feature specifics.

feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/sender/DefaultNotificationSender.kt
```diff
@@ -2,41 +2,72 @@ package net.thunderbird.feature.notification.impl.sender
...
+                DisplayInAppNotificationCommand(
+                    logger = logger,
+                    featureFlagProvider = featureFlagProvider,
+                    notificationRegistry = notificationRegistry,
+                    notification = notification,
+                    notifier = inAppNotificationNotifier,
+                ),
+            )
+        }
+    }
```
wmontwe, line 86: What should happen when this is called with an unknown notification type? I think this should log the unknown case.

wmontwe (dismissed):
LGTM

Added a comment for the feature flag, but that could also be a separate task.

feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/receiver/InAppNotificationEventBus.kt
rafaeltonholo: No, it is just an arbitrary number based on trial-and-error that provided the best results

feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/sender/DefaultNotificationSender.kt
rafaeltonholo: I agree with you. I will add a log when the `buildCommands` returns an empty list of commands.

feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/command/DismissInAppNotificationCommand.kt
rafaeltonholo: I believe we need to find a better solution for our feature flag system, but this is a bigger discussion

PR #9880, opened 2025-09-30
feat(in-app-notifications): create error notifications full screen dialog

### LINKED ISSUE #9862: In-App Error Notifications - Create Error Notifications Full screen dialog

rafaeltonholo:
Resolves #9862.

Screen recording:

https://github.com/user-attachments/assets/25dfa7e5-5a5f-40a6-bcba-ada06bea44da

### DIFF
--- a/feature/notification/api/src/androidMain/kotlin/net/thunderbird/feature/notification/api/ui/dialog/ErrorNotificationsDialog.kt
+++ b/feature/notification/api/src/androidMain/kotlin/net/thunderbird/feature/notification/api/ui/dialog/ErrorNotificationsDialog.kt
@@ -62,9 +62,9 @@ private fun ErrorNotificationsDialogContent(
                 },
             )
             LazyColumn(
-                verticalArrangement = Arrangement.spacedBy(MainTheme.spacings.half),
+                verticalArrangement = Arrangement.spacedBy(MainTheme.spacings.default),
                 contentPadding = PaddingValues(
-                    top = MainTheme.spacings.half,
+                    top = MainTheme.spacings.default,
                     start = MainTheme.spacings.double,
                     end = MainTheme.spacings.double,
                     bottom = MainTheme.spacings.double,


legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
```diff
@@ -2017,6 +2023,34 @@ class MessageListFragment :
         }
     }
 
+    override fun filterInAppNotificationEvents(event: InAppNotificationEvent): Boolean {
```
asoucar, line 2174: If adding these functions results in the "Too many functions" warning, do all of these functions need to be in this file? MessageListFragment has become a catch-all, and it may be better if functions that don't rely on anything directly in the messagelistfragment are placed in a more suitable location. At a minimum this filter could probably go in a helper or util file

rafaeltonholo: Adding these functions will not trigger the "Too many functions" warning; instead, it only exposes the existing issue. Currently, without the new functions, the `MessageListFragment` contains 120 functions, exceeding our threshold of 11.

Ideally, both filtering and click handling should be conducted within a `ViewModel`; however, we are not in an ideal situation.

For filtering, I need to use information that `MessageListFragment` maintains as state, specifically the `accountsUuid`. I could export that function to another class, but this would require passing a lambda as a parameter to retrieve the data whenever I need to filter an event, since the accounts may change.

The methods `onNotificationClick` and `onNotificationClicked` could be combined into a single function and exported to a new class. However, I still need to link this new class to the `MessageListFragment`, as I require a `Context` and a `ChildFragmentManager` to navigate. Additionally, I would have to pass the same instance of this new class to the `MessageListAdapter` through its constructor, ensuring that both components utilize the same logic.

So, I have two options to consider:

1. We can maintain the current God class structure and leave it as is for now, until we begin refactoring the Message List.
2. We can export it to a dedicated class that is intrinsically linked to the Message List. However, I would still need to override the `onNotificationActionClick`, since it is a listener function from the `ErrorNotificationsDialogFragmentActionListener`. I would also be required to pass that new class instance to the `MessageListAdapter`, as the filtering and the notification action also happen there.



<details>
<summary>Here is the diff if we consider the second option</summary>

Note: the suppress for "LargeClass" and "TooManyFunctions" will still be required.
```diff
Subject: [PATCH] chore(notifications): increase banner inline item spacing
---
Index: legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListAdapter.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListAdapter.kt b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListAdapter.kt
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListAdapter.kt	(revision 03a32c83896c4f12271af68f9ddfd92b39104dc5)
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListAdapter.kt	(date 1759489355354)
@@ -46,6 +46,7 @@
     private val relativeDateTimeFormatter: RelativeDateTimeFormatter,
     private val themeProvider: FeatureThemeProvider,
     private val featureFlagProvider: FeatureFlagProvider,
+    private val inAppNotificationsHandler: MessageListInAppNotificationsHandler,
 ) : RecyclerView.Adapter<MessageListViewHolder>() {
 
     val colors: MessageViewHolderColors = MessageViewHolderColors.resolveColors(theme)
@@ -218,8 +219,8 @@
             TYPE_IN_APP_NOTIFICATION_BANNER_INLINE_LIST if isInAppNotificationEnabled ->
                 BannerInlineListInAppNotificationViewHolder(
                     view = ComposeView(context = parent.context),
-                    eventFilter = listItemListener::filterInAppNotificationEvents,
-                    onNotificationActionClick = listItemListener::onNotificationActionClicked,
+                    eventFilter = inAppNotificationsHandler::filterInAppNotificationEvents,
+                    onNotificationActionClick = inAppNotificationsHandler::onNotificationActionClick,
                 )
 
             else -> error("Unsupported type: $viewType")
@@ -403,8 +404,6 @@
     fun onToggleMessageSelection(item: MessageListItem)
     fun onToggleMessageFlag(item: MessageListItem)
     fun onFooterClicked()
-    fun filterInAppNotificationEvents(event: InAppNotificationEvent): Boolean
-    fun onNotificationActionClicked(action: NotificationAction)
 }
 
 sealed interface MessageListViewItem {
Index: legacy/ui/legacy/src/test/java/com/fsck/k9/ui/messagelist/MessageListAdapterTest.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/legacy/ui/legacy/src/test/java/com/fsck/k9/ui/messagelist/MessageListAdapterTest.kt b/legacy/ui/legacy/src/test/java/com/fsck/k9/ui/messagelist/MessageListAdapterTest.kt
--- a/legacy/ui/legacy/src/test/java/com/fsck/k9/ui/messagelist/MessageListAdapterTest.kt	(revision 03a32c83896c4f12271af68f9ddfd92b39104dc5)
+++ b/legacy/ui/legacy/src/test/java/com/fsck/k9/ui/messagelist/MessageListAdapterTest.kt	(date 1759489355344)
@@ -438,6 +438,7 @@
             relativeDateTimeFormatter = RelativeDateTimeFormatter(context, TestClock()),
             themeProvider = FakeThemeProvider(),
             featureFlagProvider = FakeFeatureFlagProvider(),
+            inAppNotificationsHandler = mock(),
         )
     }
 
Index: legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt	(revision 03a32c83896c4f12271af68f9ddfd92b39104dc5)
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt	(date 1759490870753)
@@ -246,6 +246,25 @@
     val isShowAccountIndicator: Boolean
         get() = isUnifiedFolders || !isSingleAccountMode
 
+    private val inAppNotificationsHandler = MessageListInAppNotificationsHandler(
+        fetchAccountsUuid = { accountUuids.toSet() },
+        onUpdateIncomingServerSettingsActionClick = { action ->
+            FeatureLauncherActivity.launch(
+                context = requireContext(),
+                target = FeatureLauncherTarget.AccountEditIncomingSettings(action.accountUuid),
+            )
+        },
+        onUpdateOutgoingServerSettingsActionClick = { action ->
+            FeatureLauncherActivity.launch(
+                context = requireContext(),
+                target = FeatureLauncherTarget.AccountEditOutgoingSettings(action.accountUuid),
+            )
+        },
+        onOpenNotificationCentreActionClick = {
+            errorNotificationsDialogFragmentFactory.show(fragmentManager = childFragmentManager)
+        },
+    )
+
     override fun onAttach(context: Context) {
         super.onAttach(context)
 
@@ -357,6 +376,7 @@
             relativeDateTimeFormatter = RelativeDateTimeFormatter(requireContext(), clock),
             themeProvider = featureThemeProvider,
             featureFlagProvider = featureFlagProvider,
+            inAppNotificationsHandler = inAppNotificationsHandler,
         ).apply {
             activeMessage = this@MessageListFragment.activeMessage
         }
@@ -517,7 +537,7 @@
                                 DisplayInAppNotificationFlag.SnackbarNotifications,
                             ),
                             onSnackbarNotificationEvent = ::onSnackbarInAppNotificationEvent,
-                            eventFilter = ::filterInAppNotificationEvents,
+                            eventFilter = inAppNotificationsHandler::filterInAppNotificationEvents,
                             modifier = Modifier
                                 .animateContentSize()
                                 .onSizeChanged { size ->
@@ -2023,33 +2043,8 @@
         }
     }
 
-    override fun filterInAppNotificationEvents(event: InAppNotificationEvent): Boolean {
-        val accountUuid = event.notification.accountUuid
-        return accountUuid != null && accountUuid in accountUuids
-    }
-
-    override fun onNotificationActionClicked(action: NotificationAction) = onNotificationActionClick(action)
-
-    override fun onNotificationActionClick(action: NotificationAction) {
-        when (action) {
-            is NotificationAction.UpdateIncomingServerSettings ->
-                FeatureLauncherActivity.launch(
-                    context = requireContext(),
-                    target = FeatureLauncherTarget.AccountEditIncomingSettings(action.accountUuid),
-                )
-
-            is NotificationAction.UpdateOutgoingServerSettings ->
-                FeatureLauncherActivity.launch(
-                    context = requireContext(),
-                    target = FeatureLauncherTarget.AccountEditOutgoingSettings(action.accountUuid),
-                )
-
-            is NotificationAction.OpenNotificationCentre ->
-                errorNotificationsDialogFragmentFactory.show(fragmentManager = childFragmentManager)
-
-            else -> Unit
-        }
-    }
+    override fun onNotificationActionClick(action: NotificationAction) =
+        inAppNotificationsHandler.onNotificationActionClick(action)
 
     internal inner class MessageListActivityListener : SimpleMessagingListener() {
         private val lock = Any()
@@ -2458,3 +2453,25 @@
         }
     }
 }
+
+class MessageListInAppNotificationsHandler(
+    private val fetchAccountsUuid: () -> Set<String>,
+    private val onUpdateIncomingServerSettingsActionClick: (NotificationAction.UpdateIncomingServerSettings) -> Unit,
+    private val onUpdateOutgoingServerSettingsActionClick: (NotificationAction.UpdateOutgoingServerSettings) -> Unit,
+    private val onOpenNotificationCentreActionClick: () -> Unit,
+) {
+    fun filterInAppNotificationEvents(event: InAppNotificationEvent): Boolean {
+        val accountUuids = fetchAccountsUuid()
+        val accountUuid = event.notification.accountUuid
+        return accountUuid != null && accountUuid in accountUuids
+    }
+
+    fun onNotificationActionClick(action: NotificationAction) {
+        when (action) {
+            is NotificationAction.UpdateIncomingServerSettings -> onUpdateIncomingServerSettingsActionClick(action)
+            is NotificationAction.UpdateOutgoingServerSettings -> onUpdateOutgoingServerSettingsActionClick(action)
+            is NotificationAction.OpenNotificationCentre -> onOpenNotificationCentreActionClick()
+            else -> Unit
+        }
+    }
+}

```
</details>

I'm fine with whatever you decide. Let me know what you prefer.

asoucar (dismissed):

asoucar (approved):

PR #9885, opened 2025-10-01
Github Action No space left on device Error

shamim-emon:
- Fixes Error : No space left on device
### Implementation Highlights: 
 - Added Free disk space step right before Spotless/Detekt/Lint/Build.
 - It removes unused tools from the GitHub runner (dotnet, ghc, boost, old SDKs),  frees ~10 GB.
 - Shows disk usage before/after so you can confirm it’s working in the logs.

PR #9893, opened 2025-10-01
Fix GitHub actions disk space usage

wmontwe:
Resolves #9885


.github/workflows/markdown.yml
```diff
@@ -22,16 +22,10 @@ jobs:
...
-      - uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165 # v5.0.0
+      - name: Setup Gradle environment
+        uses: ./.github/actions/setup-gradle
         with:
-          distribution: 'temurin'
-          java-version: '21'
-
-      - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@748248ddd2a24f49513d8f472f81c3a07d4d50e1 # v4.4.4
+          cache-encryption-key: ${{ secrets.GRADLE_CACHE_ENCRYPTION_KEY }}
```
coreycb: I think pull_request only has access to the PR branch's secrets

.github/workflows/build-android.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Show disk usage
+        uses: ./.github/actions/disk-usage
+
+      - name: Cleanup disk
+        uses: ./.github/actions/disk-cleanup
+
+      - name: Setup Gradle environment
+        uses: ./.github/actions/setup-gradle
+        with:
+          cache-encryption-key: ${{ secrets.GRADLE_CACHE_ENCRYPTION_KEY }}
```
coreycb: Similar comment here

.github/workflows/markdown.yml
wmontwe: Yeah, I thought there is a setup that allows this by defining an environment. But doesn't work, will remove it and as long as the releases are not writing to the cache there should be no issue.

.github/workflows/build-android.yml
wmontwe: See above, I'll remove it.

PR #9900, opened 2025-10-02
feat(notifications): add in-app notification to message compose

### LINKED ISSUE #6276: Warn if Sent folder not found
### Checklist

### App version

6.202

With the default settings if Sent folder detection fails, sent emails aren't stored anywhere.

### Suggested solution

If “Upload sent messages” is checked but ”Sent folder“ is “Automatic (None)” show a warning message when the user tries to send an email.

_No response_

### LINKED ISSUE #9888: In-App Notifications - Add in-app notifications into message composer

rafaeltonholo:
Resolves #9888.
Resolves #6276.

- Enable Message Compose to display In-App notifications
- Introduce a warning within the Message Compose screen when no sent folder is detected
- Improve the in-app banner's animation


feature/notification/api/src/commonMain/kotlin/net/thunderbird/feature/notification/api/content/SentFolderNotFoundNotification.kt
asoucar, line 1: Why is this notification its own class? Creating a class like this makes it seem like there will be different types of SentFolderNotFound notifications, which is doesn't seem like a normal use case. Why is this not simply create a notification with the correct fields?

rafaeltonholo:
> More comments possibly pending, but wanted to get these available. I think it is a mistake to have adding the in-app notifications being added to the message compose tied in the PR to fixing the sent folder issue. I think these are two separate ideas that should have separate PRs

You are right, I will separate the fix from this PR.

feature/notification/api/src/commonMain/kotlin/net/thunderbird/feature/notification/api/content/SentFolderNotFoundNotification.kt
rafaeltonholo: This file isn't part of this PR anymore. I will move this discussion to the correct PR: #9912

PR #9912, opened 2025-10-03
feat(notifications): trigger sent folder not found in-app notification on message compose

### LINKED ISSUE #6276: Warn if Sent folder not found

rafaeltonholo:
Fixes #6276.

- Introduce a warning within the Message Compose screen when no sent folder is detected
- Add a confirmation dialog so users can choose to send the message without setting up a sent folder and warn them the message will be deleted locally

### DIFF
--- a/app-common/build.gradle.kts
+++ b/app-common/build.gradle.kts
@@ -37,6 +37,7 @@ dependencies {
     implementation(projects.feature.account.avatar.impl)
     implementation(projects.feature.account.setup)
     implementation(projects.feature.mail.account.api)
+    implementation(projects.feature.mail.message.composer)
     implementation(projects.feature.migration.provider)
     implementation(projects.feature.notification.api)
     implementation(projects.feature.notification.impl)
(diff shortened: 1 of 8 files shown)


rafaeltonholo:
> @asoucar: Why is this notification its own class? Creating a class like this makes it seem like there will be different types of SentFolderNotFound notifications, which is doesn't seem like a normal use case. Why is this not simply create a notification with the correct fields?
https://github.com/thunderbird/thunderbird-android/pull/9900#discussion_r2399804108

## Short answer: 
We intentionally model recurring domain events as concrete types. A dedicated `SentFolderNotFoundNotification` encodes the invariant fields (severity, icon, action set, and style) so every call site behaves consistently. It also makes it easier to change these invariants when needed because they’re centralized in one place.

## Long answer:
I will split the answer into a few topics.

> Why is this notification its own class?

Because the model is intentionally type-driven: each domain event encodes its invariants (severity, icon, actions, and style) once, in a named type. That prevents call-site drift and gives us a single place for localization and UX rules. It also lets us reconstruct the same notification deterministically (e.g., from `SentFolderNotFound` + `accountUuid`) to dismiss or re-emit without carrying an ID around.

> Creating a class like this makes it seem like there will be different types of `SentFolderNotFound` notifications, which is doesn't seem like a normal use case.

Not different ***types***, just different ***instances***. This event is account-scoped; two accounts with no Sent folder produce two independent notifications. The factory takes `accountUuid` for that reason.

Although the notification UI may look the same, they’re distinct because they’re tied to different accounts.

> Why is this not simply create a notification with the correct fields?

Ad-hoc field composition pushes UX decisions to every caller and is easy to get wrong (severity/actions/styling drift, inconsistent strings). Encoding those invariants in a dedicated type keeps behavior consistent, testable, and localizable. That's the core reason we avoid a generic "fill the fields" builder.

Let me know if that address your concerns.

PR #9924, opened 2025-10-07
chore(ci): android workflow improvments

wmontwe:
This joins the Android ci workflows and moves checks and cli related tasks to their own workflows.


.github/workflows/build-android.yml
```diff
@@ -45,19 +45,138 @@ jobs:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    steps:
+      - name: Check if CLI tools are changed
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
+        id: changes
+        with:
+          filters: |
+            cli_tools_changed:
+              - 'cli/**'
```
coreycb: Probably ok to use the 3rd party action and I see we were already using it, but another possibility might be just using git (with fetch-depth: 2):
```
    if git diff --name-only ${{ github.event.before }} ${{ github.sha }} | grep -q '^cli/'; then
      echo "cli_tools_changed=true" >> $GITHUB_OUTPUT
    else
      echo "cli_tools_changed=false" >> $GITHUB_OUTPUT
    fi
```
...
`if: steps.changes.outputs.cli_tools_changed == 'true'`

coreycb: Nevermind, I see this was removed in a later commit.

PR #9932, opened 2025-10-09
uplift(beta): crash when opening placeholder folder

rafaeltonholo:
This PR uplifts #9716 to beta, fixing the placeholder account crash issue.


platform34:
I see this was lifted to 13.0B5 release. Is it possible to also uplift #9743 "fix: remove Gmail prefix when updating folder list" to beta. It's the secondary part to this fix. The full resolution could then be tested in beta, instead of the partial fix.

PR #9960, opened 2025-10-16
feat(debug): add AuthDebugActions to help debug OAuth issues

wmontwe:
This introduces debug-only authentication actions for OAuth accounts in the message list UI. 

It adds new menu options (visible only in debug builds for OAuth accounts) that allow developers to simulate invalid or expired OAuth tokens and force authentication failures. 

These actions help with testing and debugging authentication flows.


wmontwe:
I did a last minute refactoring that broke the server side invalidation, need to fix it.

wmontwe:
The issue is fixed.

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
```diff
@@ -1111,6 +1128,137 @@ class MessageListFragment :
...
+                            Toast.LENGTH_SHORT,
+                        ).show()
+                    }
+                    else -> {
+                        Toast.makeText(
+                            requireContext(),
+                            R.string.debug_invalidate_access_token_unavailable,
+                            Toast.LENGTH_SHORT,
+                        ).show()
+                    }
```
rafaeltonholo: nit: The else clause here is redundant as the when was already exhausted

PR #9968, opened 2025-10-20
feat: add core file manager

wmontwe:
This was part of #9964 and extracted as standalone change.

This pull request introduces a core file module that provides a unified API for file operations across Android and JVM. It integrates the file operation with the file logger implementation.

The current implementation is limited to `copy` only, but could be extended when there is further need. Copy will be used by #9964

For better file handling, this introduces use of a KMP Uri implementation: https://github.com/eygraber/uri-kmp


wmontwe:
I resolved the conflict with main

PR #9970, opened 2025-10-20
chore(ci): Add database migration to uplift template

coreycb:


coreycb:
Considering https://github.com/thunderbird/thunderbird-android/pull/9793 I figured it would be good to ensure we don't introduce an uplift that skips a database migration # on accident.

dandarnell (approved):

PR #10008, opened 2025-10-25
feat(imap): add remaining utf8 bits

arnt:
This contains two changes that build on the earlier EAI work and completes the feature.

- MIME4J needed a parser improvement and new release; without that, Thunderbird would show "unknown sender" for some messages. With 0.8.13 it now shows the correct address. Sorry about the lateness; I overlooked the release of 0.8.13.
- Addresses such as example@zn--gr-zia.example are legal according to the syntax, because the syntax extends rather than narrows RFC5322. However, these addresses are incomprehensible to users, cause questions and even complaints, and are best rewritten to the human-readable form before reaching anyone's screen.

### DIFF
--- a/mail/common/src/main/java/com/fsck/k9/mail/helper/Rfc822Token.java
+++ b/mail/common/src/main/java/com/fsck/k9/mail/helper/Rfc822Token.java
@@ -205,6 +205,7 @@ public class Rfc822Token {
     }
 
     public static String withULabelDomain(final String address) {
+        if (address == null) return null;
         int at = address.lastIndexOf('@');
         if (at < 0 || at == address.length()-1)
             return address;
--- a/mail/common/src/main/java/com/fsck/k9/mail/helper/Rfc822Token.java
+++ b/mail/common/src/main/java/com/fsck/k9/mail/helper/Rfc822Token.java
@@ -204,7 +204,8 @@ public class Rfc822Token {
                 stringEquals(mComment, other.mComment));
     }
 
-    public static String withULabelDomain(final String address) {
+    @Nullable
+    public static String withULabelDomain(@Nullable final String address) {
         if (address == null) return null;
         int at = address.lastIndexOf('@');
         if (at < 0 || at == address.length()-1)
--- a/mail/common/src/main/java/com/fsck/k9/mail/helper/Rfc822Token.java
+++ b/mail/common/src/main/java/com/fsck/k9/mail/helper/Rfc822Token.java
@@ -17,6 +17,7 @@
 package com.fsck.k9.mail.helper;
 
 
+import java.net.IDN;
 import org.jetbrains.annotations.Nullable;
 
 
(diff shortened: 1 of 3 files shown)


wmontwe:
@arnt I updated the pr, we already use MIME4J 0.8.13. Thank you for the update!

mail/common/src/main/java/com/fsck/k9/mail/helper/Rfc822Token.java
```diff
@@ -202,4 +203,12 @@ public boolean equals(@Nullable Object o) {
                 stringEquals(mAddress, other.mAddress) &&
                 stringEquals(mComment, other.mComment));
     }
+
+    public static String withULabelDomain(final String address) {
+        int at = address.lastIndexOf('@');
```
wmontwe: This needs an additional null check because of `Rfc822Token(@Nullable String name, @Nullable String address, @Nullable String comment)`


```suggestion
    public static String withULabelDomain(final String address) {
        if (address == null) return null;
        int at = address.lastIndexOf('@');
```

wmontwe (changes requested):
Thank you! Just a small null safety check missing.

mail/common/src/main/java/com/fsck/k9/mail/helper/Rfc822Token.java
wmontwe: I fixed it.

mail/common/src/main/java/com/fsck/k9/mail/helper/Rfc822Token.java
```diff
@@ -202,4 +203,13 @@ public boolean equals(@Nullable Object o) {
                 stringEquals(mAddress, other.mAddress) &&
                 stringEquals(mComment, other.mComment));
     }
+
+    public static String withULabelDomain(final String address) {
```
rafaeltonholo: Please consider adding `@Nullable` to the method and to the parameter as well.

```suggestion
    @Nullable
    public static String withULabelDomain(final @Nullable String address) {
```

rafaeltonholo (dismissed):

arnt:
> @arnt I updated the pr, we already use MIME4J 0.8.13. Thank you for the update!

What!

I've looked, and it seems that I rebased on on my laptop and then worked on my normal desktop at home, where I was still on code from the end of July. That implies that this code isn't as thoroughly tested as I thought. Sorry.

I'm on the road this week without my test phones. I'm running 'gradle check' now. I'll test properly when I'm home next week.

jbott-tbird:
Hi @arnt are you able to make those changes?  Thank you.

arnt:
Oh no, I had this PR misfiled in my personal tracker. Of course I can. Sorry. I'll get to it tomorrow.

arnt:
It seems that I did the necessary changes in the same week, but… something. Maybe I didn't click the right buttons on github after pushing? Not sure. Anyway, Address.java has transsubstantiated itself into Address.kt since then, so a little bit of redoing was necessary.

But I saw something strange when I tested the final thing on a real device with a real server. It looked like a server bug, but I haven't the time to debug now (today's a vacation day and my wife just came home). I'll look on Monday and push then.

wmontwe:
@arnt Thanks for looking into this. Please take your time.

wmontwe:
I found an issue with the latest mime4j version and fixed it here: #11510

wmontwe:
The utf-8 issue has been resolved.

PR #10016, opened 2025-10-27
refactor(in-app-notifications): simplify receiving in-app notifications logic

### LINKED ISSUE #9991: Simplify In-app Notifications show/dismiss logic
The current logic for showing and dismissing notifications should be simplified to improve maintainability and consistency when displaying notifications while navigating across the app.

rafaeltonholo:
Resolves #9991.

Depends on #10015.

The initial logic for showing and hiding in-app notifications was built using an Event Bus. In this setup, the class or UI component expecting an in-app notification event would listen for it.

While this method functions effectively, it presents some challenges, such as race conditions and the risk of missing notifications; if a class that needs to receive a notification isn't listening for events at the time the notification is triggered, it will not receive it, as the Event Bus operates on a "fire and forget" principle.

This PR aims to simplify that logic by transitioning from the Event Bus approach to a state-based approach. Now, the class or UI component expecting an in-app notification will listen to a stream of currently displayed InAppNotifications, preventing these problems.

### DIFF
--- a/feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/command/DismissNotificationCommand.kt
+++ b/feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/command/DismissNotificationCommand.kt
@@ -54,7 +54,7 @@ sealed class DismissNotificationCommand<TNotification : Notification>(
 
             notification in notificationRegistry -> {
                 val id = checkNotNull(notificationRegistry[notification]) {
-                    "Unexcepted state when trying to dismiss a notification. " +
+                    "Unexpected state when trying to dismiss a notification. " +
                         "The required notification was not found in registry." +
                         "This might have been caused by a concurrent modification of the registry." +
                         "Please report this issue." +


legacy/core/src/main/java/com/fsck/k9/controller/MessagingController.java
```diff
@@ -2605,32 +2605,44 @@ public void checkAuthenticationProblem(LegacyAccountDto account) {
...
                 serverSettings.authenticationType == AuthType.XOAUTH2 && account.getOAuthState() == null;
     }
 
-    private void clearAuthenticationErrorNotification(LegacyAccountDto account, boolean incoming) {
+    private ServerSettings getServerSettings(LegacyAccountDto account, boolean incoming) {
+        return incoming ? account.getIncomingServerSettings() : account.getOutgoingServerSettings();
+    }
+
+    private void clearAuthenticationErrorNotification(
+        LegacyAccountDto account, boolean incoming, boolean clearOnlyForOAuthAccounts) {
```
wmontwe: This could be folded to the next line:


```suggestion
        LegacyAccountDto account, boolean incoming, boolean clearOnlyForOAuthAccounts
    ) {
```

feature/notification/impl/src/commonTest/kotlin/net/thunderbird/feature/notification/impl/DefaultNotificationRegistryTest.kt
```diff
@@ -1,20 +1,32 @@
...
 import net.thunderbird.feature.notification.api.NotificationId
 import net.thunderbird.feature.notification.testing.fake.FakeNotification
 
 @Suppress("MaxLineLength")
 class DefaultNotificationRegistryTest {
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @BeforeTest
+    fun setup() {
+        Dispatchers.setMain(StandardTestDispatcher())
+    }
```
wmontwe: You could import `implementation(projects.core.testing)` and then replace this with:


```suggestion
    @get:Rule
    val mainDispatcherRule = MainDispatcherRule(StandardTestDispatcher())
```

feature/notification/impl/src/commonTest/kotlin/net/thunderbird/feature/notification/impl/DefaultNotificationRegistryTest.kt
```diff
@@ -51,6 +63,7 @@ class DefaultNotificationRegistryTest {
             .isEqualTo(notification)
     }
 
+    @OptIn(ExperimentalCoroutinesApi::class)
```
wmontwe: With the above change, this is not needed anymore.

feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/DefaultNotificationRegistry.kt
```diff
@@ -3,35 +3,48 @@ package net.thunderbird.feature.notification.impl
...
 import net.thunderbird.feature.notification.api.NotificationId
 import net.thunderbird.feature.notification.api.NotificationRegistry
 import net.thunderbird.feature.notification.api.content.Notification
 
 @OptIn(ExperimentalAtomicApi::class)
-class DefaultNotificationRegistry : NotificationRegistry {
+class DefaultNotificationRegistry(
+    dispatcher: CoroutineDispatcher = Dispatchers.Main,
+) : NotificationRegistry {
+    private val scope: CoroutineScope = CoroutineScope(dispatcher)
```
wmontwe: I think this should use a `SupervisorJob` to avoid leaking the scope.


```suggestion
    private val scope: CoroutineScope = CoroutineScope(SupervisorJob() + dispatcher)
```

feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/DefaultNotificationRegistry.kt
```diff
@@ -3,35 +3,48 @@ package net.thunderbird.feature.notification.impl
...
     // during registration.
-    private val _registrar = mutableMapOf<Notification, NotificationId>()
+    private val internalRegistrar = MutableStateFlow<Map<Notification, NotificationId>>(emptyMap())
     private val rawId = AtomicInt(value = 0)
 
-    override val registrar: Map<NotificationId, Notification> get() = _registrar
-        .entries
-        .associate { (notification, notificationId) -> notificationId to notification }
+    internal val registrar: StateFlow<Map<NotificationId, Notification>> = internalRegistrar
+        .map { current -> current.map { it.value to it.key }.toMap() }
```
wmontwe: This creates a new list of pairs and a new map on every emission. I think the optimization above creates this situation. You could add a registry state that keeps both maps and exposes them where needed.

```kotlin
internal data class RegistrarState(
    val byNotification: Map<Notification, NotificationId>,
    val byId: Map<NotificationId, Notification>
)
```

and update `register`, `unregister` to maintain both maps.

OR if emissions/updates are rare this could simplify map inversion


```suggestion
        .map { current -> current.entries.associate { it.value to it.key } }
```

feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/command/DismissNotificationCommand.kt
wmontwe: Small typo: Unexcepted -> Unexpected

feature/notification/impl/src/commonTest/kotlin/net/thunderbird/feature/notification/impl/DefaultNotificationRegistryTest.kt
rafaeltonholo: The issue with this approach is that `@Rule` comes from JUnit. It currently works, as we have no non-JVM targets, but it will fail once we add a new target. I'm not sure if we have a full KMP approach for `@Rule`.

However, I don't see a problem using it for now

feature/notification/impl/src/commonTest/kotlin/net/thunderbird/feature/notification/impl/DefaultNotificationRegistryTest.kt
rafaeltonholo: This isn't necessary at all. It might be a leftover

feature/notification/impl/src/commonMain/kotlin/net/thunderbird/feature/notification/impl/DefaultNotificationRegistry.kt
rafaeltonholo: Thanks for the feedback! I have moved to the data class approach.

feature/notification/impl/src/commonTest/kotlin/net/thunderbird/feature/notification/impl/DefaultNotificationRegistryTest.kt
wmontwe: Yes, I think we could inject the dispatcher into the test subject.

Then we could do:

```kotlin

@Test
fun `test with dispatcher overwrite`() = runTest {
    val dispatcher = StandardTestDispatcher(testScheduler)
    val registry = DefaultNotificationRegistry(dispatcher)

    ...
}

```

PR #10031, opened 2025-10-30
fix: biometric prompt is not shown

### LINKED ISSUE #10013: Displaying the password in plain text does not prompt for authentication
### Checklist

### App

Thunderbird for Android

### App version

13.0 and 14.0b2

Google Play

### Android version

LineageOS 16 (Android 9), Stock Android 10, OneUI 8 (Android 16)

### Device model

Gigaset GS290, Samsung Galaxy(s)

### Steps to reproduce

Go to "Settings"
Select an (IMAP)-account
Select "Fetching Mail"
Select "Incoming Server"
Unmask the password field

### Expected behavior

A prompt for authentication with the device PIN or biometric authentication should be displayed.
(k9-mail did this in earlier versions.)

### Actual behavior

The password is displayed in plain text without further prompting.

### Logs

_No response_

wmontwe:
Resolves #10013 

This refactors the password input components to use a flexible authentication interface, replacing the previous hardcoded biometric authentication logic. The changes introduce a new `Authenticator` abstraction, allowing for easier extension and testing. Updated the UI components and added tests accordingly. The authenticated reveal logic is now tested and working.

### DIFF
--- a/feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/ui/common/ProtectedPasswordInput.kt
+++ b/feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/ui/common/ProtectedPasswordInput.kt
@@ -1,6 +1,6 @@
 package app.k9mail.feature.account.server.settings.ui.common
 
-import androidx.activity.compose.LocalActivity
+import androidx.annotation.StringRes
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.runtime.Composable
@@ -15,15 +15,16 @@ import app.k9mail.core.ui.compose.designsystem.molecule.input.InputLayout
 import app.k9mail.core.ui.compose.designsystem.molecule.input.inputContentPadding
 import app.k9mail.feature.account.server.settings.R
 import kotlinx.coroutines.delay
+import net.thunderbird.feature.account.server.settings.ui.common.AuthenticationError
 import net.thunderbird.feature.account.server.settings.ui.common.Authenticator
-import net.thunderbird.feature.account.server.settings.ui.common.BiometricAuthenticator
+import net.thunderbird.feature.account.server.settings.ui.common.rememberBiometricAuthenticator
 import app.k9mail.core.ui.compose.designsystem.R as RDesign
 
 private const val SHOW_WARNING_DURATION = 5000L
 
 /**
  * Variant of [PasswordInput] that only allows the password to be unmasked after the user has authenticated using
- * [Authenticator] that defaults to [BiometricAuthenticator].
+ * [Authenticator] that defaults to [rememberBiometricAuthenticator].
  */
 @Composable
 fun ProtectedPasswordInput(
@@ -33,14 +34,18 @@ fun ProtectedPasswordInput(
     isRequired: Boolean = false,
     errorMessage: String? = null,
     contentPadding: PaddingValues = inputContentPadding(),
-    authenticator: Authenticator? = null,
+    authenticator: Authenticator = rememberBiometricAuthenticator(
+        title = stringResource(R.string.account_server_settings_password_authentication_title),
+        subtitle = stringResource(R.string.account_server_settings_password_authentication_subtitle),
+    ),
 ) {
-    var biometricWarning by remember { mutableStateOf<String?>(value = null) }
+    var authenticationError by remember { mutableStateOf<AuthenticationError?>(value = null) }
+    val authenticationWarning = authenticationError?.let { stringResource(it.mapToStringRes()) }
 
-    LaunchedEffect(key1 = biometricWarning) {
-        if (biometricWarning != null) {
+    LaunchedEffect(key1 = authenticationError) {
+        if (authenticationError != null) {
             delay(SHOW_WARNING_DURATION)
-            biometricWarning = null
+            authenticationError = null
         }
     }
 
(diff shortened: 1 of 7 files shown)


feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/ui/common/ProtectedPasswordInput.kt
```diff
@@ -56,13 +55,21 @@ fun BiometricPasswordInput(
...
-        TextFieldOutlinedPasswordBiometric(
+        val resolvedAuthenticator: Authenticator = authenticator ?: run {
+            val activity = LocalActivity.current as androidx.fragment.app.FragmentActivity
+            BiometricAuthenticator(
+                activity = activity,
+                title = title,
+                subtitle = subtitle,
+                needScreenLockMessage = needScreenLockMessage,
+            )
+        }
```
rafaeltonholo: The `BiometricAuthenticator` will be recreated every time the component recomposes, in case the `authenticator` passed via parameter is `null`. Meaning it will create a new instance on every character typed.

Please, wrap it into a remember block:

```suggestion
        val activity = LocalActivity.current as androidx.fragment.app.FragmentActivity
        val resolvedAuthenticator: Authenticator = remember(authenticator, title, subtitle, needScreenLockMessage) {
            authenticator ?: BiometricAuthenticator(
                activity = activity,
                title = title,
                subtitle = subtitle,
                needScreenLockMessage = needScreenLockMessage,
            )
        }
```

feature/account/server/settings/src/main/kotlin/net/thunderbird/feature/account/server/settings/ui/common/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            override fun onAuthenticationError(errorCode: Int, errString: CharSequence) {
+                val message: String = when (errorCode) {
+                    BiometricPrompt.ERROR_HW_NOT_PRESENT,
+                    BiometricPrompt.ERROR_NO_DEVICE_CREDENTIAL,
+                    BiometricPrompt.ERROR_NO_BIOMETRICS,
+                    -> needScreenLockMessage
+
+                    else -> if (errString.isNotEmpty()) errString.toString() else "Authentication failed"
```
rafaeltonholo: The "Authentication failed" should be localized.

feature/account/server/settings/src/main/kotlin/net/thunderbird/feature/account/server/settings/ui/common/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,71 @@
...
+            )
+            .setTitle(title)
+            .setSubtitle(subtitle)
+            .build()
+
+        val executor = ContextCompat.getMainExecutor(activity)
+        try {
+            BiometricPrompt(activity, executor, authenticationCallback).authenticate(promptInfo)
+        } catch (e: Exception) {
+            val message: String = e.message ?: "Unable to start biometric prompt"
```
rafaeltonholo: "Unable to start biometric prompt" should be localized

feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/ui/common/ProtectedTextFieldOutlinedPassword.kt
```diff
@@ -0,0 +1,86 @@
...
+        activity.setSecure(isPasswordVisible)
+
+        onDispose {
+            activity.setSecure(false)
+        }
+    }
+}
+
+private fun Activity.setSecure(secure: Boolean) {
+    window.setFlags(if (secure) WindowManager.LayoutParams.FLAG_SECURE else 0, WindowManager.LayoutParams.FLAG_SECURE)
```
rafaeltonholo, line 86: suggestion: Although this is nice to have, I don't think that this input should be responsible for setting the Activity in the secure mode.

Instead, I would have a `onPasswordVisibleChange` callback, and the screen that consumes this component would change the Activity flag. 

Additionally, removing the `FLAG_SECURE` flag when disposing of this component opens the possibility of retrieving the password via a screenshot. 

See video below:

https://github.com/user-attachments/assets/6ecc5d8e-14ca-4bbf-8410-b182a59f99ce

wmontwe:
I agree that the implementation is not great and it doesn't work reliably. It basically still follows the original request to keep "[existing feature that only reveals a saved password when the user has authenticated](https://github.com/thundernest/k-9/blob/ce9704b275dca2d98c772b9bd191e7ceee70c2a0/app/ui/base/src/main/java/com/fsck/k9/ui/base/extensions/TextInputLayoutExtensions.kt)" from: #7202

This just patched the implementation to a state that it works, but it's not meant to be watertight.

feature/account/server/settings/src/main/kotlin/app/k9mail/feature/account/server/settings/ui/common/ProtectedPasswordInput.kt
wmontwe: Right, I wanted to inject the authenticator and somehow left this piece in there.

rafaeltonholo:
> I agree that the implementation is not great and it doesn't work reliably. It basically still follows the original request to keep "[existing feature that only reveals a saved password when the user has authenticated](https://github.com/thundernest/k-9/blob/ce9704b275dca2d98c772b9bd191e7ceee70c2a0/app/ui/base/src/main/java/com/fsck/k9/ui/base/extensions/TextInputLayoutExtensions.kt)" from: #7202
> 
> This just patched the implementation to a state that it works, but it's not meant to be watertight.

The comments I provided, aside from the one concerning FLAG_SECURE, are not meant to make the implementation watertight but rather to mitigate potential future issues (localization and performance). 

Furthermore, after reviewing the comment on #7202, I feel that the current implementation does not fulfill its intended purpose:

> That's to prevent the unmasked password from showing up in screenshots.

As demonstrated in the screen recording, I was able to take a screenshot that revealed the unmasked password.

However, the FLAG_SECURE issue could be addressed in another task, as it was a suggestion comment.

wmontwe:
I think there might be a misunderstanding — I never meant that localization or performance shouldn’t be addressed (why would I?). I completely agree that both are important considerations, and I assumed it was obvious that these would be fixed, even if I didn’t say so explicitly. But if it helps, I can make that clearer next time — for example, by explicitly commenting or adding a 👍 to acknowledge.

My point was simply that the previous implementation didn’t actually achieve its intended purpose, so the priority was to get it working first without rearchitecting the entire functionality. The only change beyond that was improving testability, since the original logic wasn’t functioning as intended. Once the feature works reliably, we can refine the FLAG_SECURE handling and other improvements in a follow-up.

rafaeltonholo:
> I think there might be a misunderstanding — I never meant that localization or performance shouldn’t be addressed (why would I?). I completely agree that both are important considerations, and I assumed it was obvious that these would be fixed, even if I didn’t say so explicitly. But if it helps, I can make that clearer next time — for example, by explicitly commenting or adding a 👍 to acknowledge.
> 
> My point was simply that the previous implementation didn’t actually achieve its intended purpose, so the priority was to get it working first without rearchitecting the entire functionality. The only change beyond that was improving testability, since the original logic wasn’t functioning as intended. Once the feature works reliably, we can refine the FLAG_SECURE handling and other improvements in a follow-up.

I'm sorry, I'm a bit slow today, and I might have misunderstood your first comment! Thanks for making it clear! I agree with you

PR #10034, opened 2025-10-31
Remove default flags

Goooler:
https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes#default-changes


Goooler:
Green!

Issue #9978, opened 2025-10-21
Revert unexpected default value changes when migrated from preferences to use the new PreferenceManager


VladLucaci:
# The issue

The preference "General Settings > Display > Message lists > Colorize contact pictures" got its default value unexpectedly changed from `true` to `false`. We should revert that change.

Additionally, we should double-check the other default values that were migrated to make sure no other unexpected default value change was made.

---
# Original post

### Checklist

- [x] I have used the search function to see if someone else has already submitted the same bug report.
- [x] I will describe the problem with as much detail as possible.

### App

Thunderbird for Android

### App version

13.0 and 14.0b1

### Where did you get the app from?

Google Play

### Android version

15 and 16

### Device model

Samsung Galaxy Fold 4

### Steps to reproduce

1. Launch TB.
2. Login with valid credentials.
3. Reach the inbox.

### Expected behavior

All the Sender icons are coloured.

### Actual behavior

All the Sender icons are black and white.

This issue is NOT present on K9 13.0 and Thunderbird 12.1.
This issue is NOT present when application is updated to 13.0 and 14.0b1 respectively.

### Logs

![Image](https://github.com/user-attachments/assets/f4f5587b-9e26-4ad1-86e5-4f977fa5b350)

rafaeltonholo:
Hi @VladLucaci, 

It feels like the "Colorize contact pictures" preference is disabled on your Thunderbird app. Could you please double-check that?

You can find this preference by:
1. Navigate to Settings
2. Navigate to General Settings
3. Navigate to Display
4. Scroll down to the "Message List" section
5. Check the "Colorize contact pictures" option

VladLucaci:
You were right, that option was indeed disabled.

However , I did not interact with that option at all. I simply installed a fresh instance of TB , went through the login process and noticed the icons are not coloured.

With the same steps, on K9 13.0 or even TB 12.1 the "Colorize contact pictures" is enabled by default. So maybe this is the issue actually? That's why I did not encounter this issue while updating the application, because I had the option turned on by default in 12.1. (option becomes deselected only on fresh installs)

rafaeltonholo:
> You were right, that option was indeed disabled.
> 
> However , I did not interact with that option at all. I simply installed a fresh instance of TB , went through the login process and noticed the icons are not coloured.
> 
[...]

PR #10035, opened 2025-11-01
Restore Original Default Value for "messagelist_colorize_missingcontact_picture" after migration to new PreferenceManager

shamim-emon:
- Sees #9978 
 - Some default values of Preferences were mistakenly changed during migration to new PreferenceManager. This PR restores  original default for "messagelist_colorize_missingcontact_picture"


wmontwe (approved):
LGTM

Issue #9031, opened 2025-04-02
Account Settings - Account Monogram customization


wmontwe:
The account monogram is currently derived from the account name. Users should be able to adjust the monogram to their liking.

Acceptance criteria:
- It should be composed of min. 1 and max. 3 characters

kewisch:
What sort of data could they enter that breaks the account?

wmontwe:
> What sort of data could they enter that breaks the account?

I changed this to customizable monogram, which makes more sense.

PR #10048, opened 2025-11-04
refactor: change account color handling

wmontwe:
Part of #9031

This refactors the account color picking logic to use the new `AccountProfileRepository` and an immutable account color list and improves dependency injection setup. The color picking logic now ensures colors are distributed as evenly as possible among accounts. Rendering the `default_account_colors` obsolete.

It also introduces documentation for the `Outcome` type.


wmontwe:
Close in favor of #10104

PR #10069, opened 2025-11-06
chore(feature-flag): enable dropdown drawer for release

wmontwe:
Resolves #10068

PR #10072, opened 2025-11-06
uplift(beta): enable feature-flag dropdown drawer for release

wmontwe:
This uplifts #10069 to beta.


coreycb:
I think we can land this on beta and not release another beta before the merge of beta->release, if that makes sense.

PR #10076, opened 2025-11-07
fix(message-reader): complex html email rendering incorrectly

### LINKED ISSUE #3172: Wrap long lines when displaying message (text size gets too small)
K9 doesn't warp links, which makes some e-mails very wide:
screenshot_20180208-131048
***
compared to:
screenshot_20180208-131057

rafaeltonholo:
Fixes #3172.

This PR will address the majority of emails with complex HTML and CSS being rendered incorrectly. Large images and tables will still zoom out the WebView.

- Introduce the `:feature:mail:message:reader:api` and `:feature:mail:message:reader:impl` modules
- Adds a new feature flag "use_new_message_viewer_css_styles"
- Moved CSS styles from `DisplayHtml` to `CssStyleProvider` interface implementations, enabling old CSS fallback when the feature flag is off
- Add new CSS rules, supporting more complex email rendering for mobile

For future improvements: 
> We need to figure out a way to verify if we can resize large images and tables without using JavaScript. Otherwise, we will need to decide on either enabling JS, but limiting its usage via HTML preprocessor, or leaving it zooming out the WebView.

### DIFF
--- a/cli/html-cleaner-cli/build.gradle.kts
+++ b/cli/html-cleaner-cli/build.gradle.kts
@@ -9,8 +9,6 @@ application {
 }
 
 dependencies {
-    implementation(projects.core.featureflag)
-    implementation(projects.feature.mail.message.reader.api)
     implementation(projects.library.htmlCleaner)
 
     implementation(libs.clikt)
--- a/cli/html-cleaner-cli/src/main/kotlin/app/k9mail/cli/html/cleaner/Main.kt
+++ b/cli/html-cleaner-cli/src/main/kotlin/app/k9mail/cli/html/cleaner/Main.kt
@@ -12,8 +12,6 @@ import com.github.ajalt.clikt.parameters.types.enum
 import com.github.ajalt.clikt.parameters.types.file
 import com.github.ajalt.clikt.parameters.types.inputStream
 import java.io.File
-import net.thunderbird.core.featureflag.FeatureFlagResult
-import net.thunderbird.feature.mail.message.reader.api.css.CssClassNameProvider
 import okio.buffer
 import okio.sink
 import okio.source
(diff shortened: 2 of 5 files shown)


app-k9mail/src/main/kotlin/app/k9mail/feature/FeatureModule.kt
```diff
@@ -18,4 +21,11 @@ val featureModule = module {
     includes(featureMessageListModule)
 
     single<FundingSettings> { K9FundingSettings() }
+
+    single<CssClassNameProvider> {
+        DefaultCssClassNameProvider(
+            featureFlagProvider = get(),
+            defaultNamespaceClassName = BuildConfig.APPLICATION_ID.replace(".", "-"),
```
wmontwe, line 28: 👍

app-thunderbird/src/daily/kotlin/net/thunderbird/android/featureflag/TbFeatureFlagFactory.kt
```diff
@@ -25,6 +26,7 @@ class TbFeatureFlagFactory : FeatureFlagFactory {
             FeatureFlag(MessageListFeatureFlags.UseComposeForMessageListItems, enabled = false),
             FeatureFlag(MessageViewFeatureFlags.ActionExportEml, enabled = true),
             FeatureFlag(AccountSettingsFeatureFlags.EnableAvatarCustomization, enabled = false),
+            FeatureFlag(MessageViewerFeatureFlags.UseNewMessageViewerCssStyles, enabled = false),
```
wmontwe: You could enable features that are already functionally complete for daily builds as well.

legacy/ui/base/src/main/java/com/fsck/k9/ui/base/KoinModule.kt
```diff
@@ -13,7 +14,7 @@ val uiBaseModule = module {
             generalSettingsManager = get(),
             appCoroutineScope = get(named("AppCoroutineScope")),
         )
-    }
+    } bind net.thunderbird.core.ui.theme.api.ThemeManager::class
```
wmontwe: Could be an import declaration.

app-thunderbird/src/daily/kotlin/net/thunderbird/android/featureflag/TbFeatureFlagFactory.kt
rafaeltonholo: Thanks for the reminder. I've enabled it for daily builds

Djfe:
very interesting, I will report back once this hits a release and give feedback here
Thanks for working on this!! 🙏

PR #10100, opened 2025-11-12
feat(account-setting): add account monogram customization part 1

wmontwe:
Part 1 of #9031

Depends on #10048


app-common/src/main/kotlin/net/thunderbird/app/common/account/AppCommonAccountModule.kt
```diff
@@ -43,10 +48,16 @@ internal val appCommonAccountModule = module {
...
+            ThemeCommonR.array.account_colors,
+        ).toList().toImmutableList()
+    }
+
     factory {
         AccountColorPicker(
-            accountManager = get(),
-            resources = get(),
+            repository = get(),
+            accountColors = get(named("AccountColors")),
```
rafaeltonholo: suggestion: use a const to `"AccountColors"`, so it makes easier to maintain in the feature.

```suggestion
            accountColors = get(named(AccountColorPicker.ACCOUNT_COLORS_QUALIFIER)),
```

and then:
```kotlin
class AccountColorPicker(/* ... */) {
    companion object {
        const val ACCOUNT_COLORS_QUALIFIER = "AccountColors"
    }
    // ... rest of the code
}

```

wmontwe:
Close in favor of #10104

PR #10101, opened 2025-11-12
feat(account-setting): add account monogram customization part 2

wmontwe:
Part 2 of #9031

Depends on #10100


wmontwe:
Close in favor of #10104

PR #10104, opened 2025-11-13
feat(account-setting): add account monogram customization part 3

### LINKED ISSUE #9031: Account Settings - Account Monogram customization

wmontwe:
Resolves #9031

### DIFF
--- a/feature/account/core/build.gradle.kts
+++ b/feature/account/core/build.gradle.kts
@@ -2,11 +2,12 @@ plugins {
     id(ThunderbirdPlugins.Library.kmp)
 }
 
-android {
-    namespace = "net.thunderbird.feature.account.core"
-}
-
 kotlin {
+    androidLibrary {
+        namespace = "net.thunderbird.feature.account.core"
+        withHostTest {}
+    }
+
     sourceSets {
         commonMain.dependencies {
             api(projects.feature.account.api)


app-common/src/main/kotlin/net/thunderbird/app/common/account/AccountColorPicker.kt
```diff
@@ -1,27 +1,26 @@
...
-import net.thunderbird.core.android.account.LegacyAccountDtoManager
+import kotlinx.collections.immutable.ImmutableList
+import kotlinx.coroutines.flow.first
+import net.thunderbird.feature.account.profile.AccountProfileRepository
 
 internal class AccountColorPicker(
-    private val accountManager: LegacyAccountDtoManager,
-    private val resources: Resources,
+    private val repository: AccountProfileRepository,
+    private val accountColors: ImmutableList<Int>,
```
rafaeltonholo, line 9: Why this needs to be an `ImmutableList` instead of just `List`?

app-common/src/main/kotlin/net/thunderbird/app/common/account/AppCommonAccountModule.kt
```diff
@@ -43,10 +48,16 @@ internal val appCommonAccountModule = module {
...
+            ThemeCommonR.array.account_colors,
+        ).toList().toImmutableList()
+    }
+
     factory {
         AccountColorPicker(
-            accountManager = get(),
-            resources = get(),
+            repository = get(),
+            accountColors = get(named("AccountColors")),
```
rafaeltonholo, line 60: suggestion: use a const to `"AccountColors"`, so it makes easier to maintain in the feature.

```suggestion
            accountColors = get(named(AccountColorPicker.ACCOUNT_COLORS_QUALIFIER)),
```

and then:
```kotlin
class AccountColorPicker(/* ... */) {
    companion object {
        const val ACCOUNT_COLORS_QUALIFIER = "AccountColors"
    }
    // ... rest of the code
}

```

core/validation/src/commonMain/kotlin/net/thunderbird/core/validation/input/IntegerInputField.kt
```diff
@@ -0,0 +1,76 @@
+package net.thunderbird.core.validation.input
+
+import net.thunderbird.core.outcome.Outcome
+import net.thunderbird.core.validation.ValidationError
+import net.thunderbird.core.validation.ValidationOutcome
+
+class IntegerInputField(
```
rafaeltonholo, line 7: suggestion: This feels like a manual data class... Why not declare it as a data class and use `.copy` on each `update` method?

That would avoid us manually overriding `equals` and `hashCode` method.

core/ui/setting/impl-dialog/src/main/kotlin/net/thunderbird/core/ui/setting/dialog/ui/components/dialog/value/TextDialogView.kt
```diff
@@ -43,20 +44,34 @@ internal fun TextDialogView(
...
     LaunchedEffect(Unit) {
         delay(EDIT_TEXT_FOCUS_DELAY)
         focusRequester.requestFocus()
     }
 
+    LaunchedEffect(textFieldValue.text) {
+        delay(VALIDATION_DEBOUNCE_DELAY)
+        val transformedText = setting.transform(textFieldValue.text)
+        errorMessage = setting.validate(transformedText)
+    }
```
rafaeltonholo, line 57: Unless you cancel the execution of the code after the `delay`, this won't behave as a debouce, but will just delay the execution of each change.

I would use a `derivedStateOf` instead.

core/ui/setting/impl-dialog/src/main/kotlin/net/thunderbird/core/ui/setting/dialog/ui/components/dialog/value/TextDialogView.kt
```diff
@@ -43,20 +44,34 @@ internal fun TextDialogView(
             ),
         )
     }
+    var errorMessage by remember { mutableStateOf<String?>(null) }
```
rafaeltonholo, line 47: I guess using a `derivedStateOf` here instead will fit better with what you want.

```suggestion
    var errorMessage by remember {
        derivedStateOf {
            val transformedText = setting.transform(textFieldValue.text)
            setting.validate(transformedText)
        }
    }
```

feature/account/settings/impl/src/main/kotlin/net/thunderbird/feature/account/settings/impl/domain/usecase/GetGeneralSettings.kt
```diff
@@ -1,141 +1,28 @@
...
-import net.thunderbird.feature.account.settings.AccountSettingsFeatureFlags
-import net.thunderbird.feature.account.settings.impl.domain.AccountSettingsDomainContract.ResourceProvider
-import net.thunderbird.feature.account.settings.impl.domain.AccountSettingsDomainContract.SettingsError
+import net.thunderbird.feature.account.settings.impl.domain.AccountSettingsDomainContract.AccountSettingError
 import net.thunderbird.feature.account.settings.impl.domain.AccountSettingsDomainContract.UseCase
-import net.thunderbird.feature.account.settings.impl.domain.AccountSettingsOutcome
-import net.thunderbird.feature.account.settings.impl.domain.entity.GeneralPreference
-import net.thunderbird.feature.account.settings.impl.domain.entity.generateId
 
 internal class GetGeneralSettings(
```
rafaeltonholo: With the present change, I guess we should rename this to `GetAccountProfileSettings` or `GetAccountProfile`

feature/account/settings/impl/src/main/kotlin/net/thunderbird/feature/account/settings/impl/domain/usecase/ValidateAccountName.kt
```diff
@@ -0,0 +1,26 @@
...
+            name.length > MAX_NAME_LENGTH -> Outcome.failure(ValidateAccountNameError.TooLongName)
+            else -> Outcome.success(Unit)
+        }
+    }
+
+    private companion object {
+        /**
+         * Maximum length for an account name.
+         *
+         * See RFC 5321, 4.5.3.1.3.
```
rafaeltonholo:
```suggestion
         * See [RFC 5321, 4.5.3.1.3](https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.3).
```

feature/account/settings/impl/src/main/kotlin/net/thunderbird/feature/account/settings/impl/ui/general/GeneralSettingId.kt
```diff
@@ -0,0 +1,9 @@
+package net.thunderbird.feature.account.settings.impl.ui.general
+
+internal object GeneralSettingId {
```
rafaeltonholo, line 3: Shouldn't this be named `AccountSettingsId`?

feature/account/settings/impl/src/main/kotlin/net/thunderbird/feature/account/settings/impl/ui/general/GeneralSettingsBuilder.kt
```diff
@@ -0,0 +1,179 @@
...
+import net.thunderbird.feature.account.settings.AccountSettingsFeatureFlags
+import net.thunderbird.feature.account.settings.impl.domain.AccountSettingsDomainContract
+import net.thunderbird.feature.account.settings.impl.domain.AccountSettingsDomainContract.ValidateAccountNameError
+import net.thunderbird.feature.account.settings.impl.domain.AccountSettingsDomainContract.ValidateMonogramError
+import net.thunderbird.feature.account.settings.impl.ui.general.GeneralSettingsContract.State
+
+/**
+ * Builds the General Settings from [State].
+ */
+internal class GeneralSettingsBuilder(
```
rafaeltonholo, line 23: Nice!

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/common/DisplayAccountUtils.kt
```diff
@@ -29,10 +29,23 @@ internal fun getDisplayAccountName(account: DisplayAccount): String {
...
         }
+    }
+}
+
+@Composable
+internal fun getDisplayAccountAvatar(account: DisplayAccount): Avatar {
+    return when (account) {
+        is UnifiedDisplayAccount -> {
+            val name = stringResource(R.string.navigation_drawer_dropdown_unified_account_title)
+            Avatar.Monogram(name.first().uppercase())
```
rafaeltonholo, line 44: Shouldn't we be using an icon instead of a monogram, when it is the Unified Account?

app-common/src/main/kotlin/net/thunderbird/app/common/account/AppCommonAccountModule.kt
wmontwe: This should move to the app theme and be a typed value. For the sake of this PR I left it as simple named accessor. Added a ticket: #10108

core/validation/src/commonMain/kotlin/net/thunderbird/core/validation/input/IntegerInputField.kt
wmontwe: I can't remember what was the original reason to not use a data class, too long ago. I added a follow-up ticket: #10107

feature/account/settings/impl/src/main/kotlin/net/thunderbird/feature/account/settings/impl/ui/general/GeneralSettingId.kt
wmontwe: It's just the general account settings part and there will be more once we migrate other settings.

feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/common/DisplayAccountUtils.kt
wmontwe: The icons are added later, this just ensures that a proper avatar is available.

core/ui/setting/impl-dialog/src/main/kotlin/net/thunderbird/core/ui/setting/dialog/ui/components/dialog/value/TextDialogView.kt
wmontwe: I'll have a look.

core/ui/setting/impl-dialog/src/main/kotlin/net/thunderbird/core/ui/setting/dialog/ui/components/dialog/value/TextDialogView.kt
wmontwe: I'll have a look.

wmontwe: This should works as debounce, the `LaunchedEffect` cancels all prior executions as soon as the key `textFieldValue.text` changes. To avoid edge cases I could launch the validation as a coroutine and keep a reference to the job. Then cancel the job in `onTextChange`, prior to launching a new coroutine.

core/ui/setting/impl-dialog/src/main/kotlin/net/thunderbird/core/ui/setting/dialog/ui/components/dialog/value/TextDialogView.kt
wmontwe: I tested this, but realised that the text transform is redundant. The text is already transformed in `onTextChange` and then stored as `textFieldValue` -> removed it.

app-common/src/main/kotlin/net/thunderbird/app/common/account/AccountColorPicker.kt
wmontwe: It should signal that this list is not changeable and when passed to a Composable it is considered as _**stable**_. A standard `List`, by contrast, is always treated as **unstable**, forcing recomposition even if its contents are identical, because the compiler cannot guarantee that the underlying list has not been mutated.

rafaeltonholo: I couldn't find any Composable that is using the `AccountColorPicker`. If the main reason is to signal that the list must not change, I'm ok with it.

wmontwe:
The mdbook failure is fixed in #10121, will update this PR once merged.

PR #10121, opened 2025-11-20
Fix: pin mdbook dependencies

### LINKED ISSUE #10118: Mdbook should pin dependencies
Our docs use mdbook for rendering, but the dependency versions are not fixed and could lead to breaking changes in unrelated context. 

We should pin the dependencies.

wmontwe:
Resolves #10118 

Also updates mdbook to 0.5.0 and removes outdated plugins.

### DIFF
--- a/docs/book.toml
+++ b/docs/book.toml
@@ -9,7 +9,7 @@ src = "."
 edition = "2018"
 
 [build]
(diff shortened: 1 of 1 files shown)


rafaeltonholo (approved):
LGTM

PR #10135, opened 2025-11-24
feat(account-settings): add icon customization

### LINKED ISSUE #9369: Account Settings - Account Icons support
Add support for account icons:

Image

wmontwe:
Resolves #9369

To test this you need to enable `AccountSettingsFeatureFlags.EnableAvatarCustomization`.

grafik

The unified account defaults to the group icon now.

### DIFF
--- a/core/ui/setting/impl-dialog/src/debug/kotlin/net/thunderbird/core/ui/setting/dialog/ui/components/common/AutoHeightLazyVerticalGridPreview.kt
+++ b/core/ui/setting/impl-dialog/src/debug/kotlin/net/thunderbird/core/ui/setting/dialog/ui/components/common/AutoHeightLazyVerticalGridPreview.kt
@@ -1,0 +1,131 @@
+package net.thunderbird.core.ui.setting.dialog.ui.components.common
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import app.k9mail.core.ui.compose.designsystem.PreviewWithTheme
+import app.k9mail.core.ui.compose.designsystem.atom.text.TextBodyLarge
+import kotlinx.collections.immutable.persistentListOf
+
+@Composable
+@Preview(showBackground = true)
+internal fun AutoHeightLazyVerticalWithOnePreview() {
+    PreviewWithTheme {
+        AutoHeightLazyVerticalGrid(
+            items = persistentListOf(1),
+            itemSize = 64.dp,
+            itemContent = { ItemContent(it) },
+        )
+    }
+}
+
+@Composable
+@Preview(showBackground = true)
+internal fun AutoHeightLazyVerticalWithOneShortPreview() {
+    PreviewWithTheme {
+        AutoHeightLazyVerticalGrid(
+            items = persistentListOf(1, 2, 3, 4),
+            itemSize = 64.dp,
+            itemContent = { ItemContent(it) },
+        )
+    }
+}
+
+@Composable
+@Preview(showBackground = true)
+internal fun AutoHeightLazyVerticalGridOverflowByOnePreview() {
+    PreviewWithTheme {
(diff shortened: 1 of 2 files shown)


core/ui/setting/impl-dialog/src/main/kotlin/net/thunderbird/core/ui/setting/dialog/ui/components/list/value/IconListItem.kt
```diff
@@ -0,0 +1,48 @@
...
+            IconView(
+                icon = icon,
+                color = setting.color,
+                isSelected = icon == setting.value,
+                onClick = {
+                    onSettingValueChange(setting.copy(value = icon))
+                },
+                modifier = Modifier.size(MainTheme.sizes.iconAvatar),
+            )
+        }
```
rafaeltonholo, line 46: Is there a specific reason for not using `FlowRow` instead of creating a new custom layout?

It seems that `FlowRow` is exactly the component you need to achieve your goal.

Also, I've noticed that the height calculation in `AutoHeightLazyVerticalGrid` is incorrect, causing the rocket icon to be hidden.

```suggestion
        FlowRow(
            verticalArrangement = Arrangement.spacedBy(MainTheme.spacings.default),
            horizontalArrangement = Arrangement.spacedBy(MainTheme.spacings.default),
        ) {
            setting.icons.forEach { icon ->
                key(icon.id) {
                    IconView(
                        icon = icon,
                        color = setting.color,
                        isSelected = icon == setting.value,
                        onClick = {
                            onSettingValueChange(setting.copy(value = icon))
                        },
                        modifier = Modifier.size(MainTheme.sizes.iconAvatar),
                    )
                }
            }
        }
```

feature/account/avatar/impl/src/main/kotlin/net/thunderbird/feature/account/avatar/ui/CompatAccountAvatar.kt
```diff
@@ -18,6 +18,7 @@ fun rememberCompatAvatar(
 ): Avatar = remember(avatar, name) {
     when (avatar) {
         is Avatar.Monogram -> avatar
+        is Avatar.Icon -> avatar
```
rafaeltonholo, line 21: nit: You could use instead:
```diff
-        is Avatar.Monogram -> avatar
-        is Avatar.Icon -> avatar        
+        is Avatar.Monogram, is Avatar.Icon -> avatar
```

feature/account/avatar/api/src/commonMain/kotlin/net/thunderbird/feature/account/avatar/AvatarIconCatalog.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * The default icon to use when an unknown name is requested.
+     */
+    val defaultIcon: TIcon
+
+    /**
+     * The stable string name of the default icon.
+     */
+    val defaultName: String
```
rafaeltonholo: Is it required to use strings for the names? If not, I suggest considering using enums instead.

feature/account/avatar/impl/src/main/kotlin/net/thunderbird/feature/account/avatar/ui/CompatAccountAvatar.kt
wmontwe: This is just there for a short time, it will be removed in #10162

core/ui/setting/impl-dialog/src/main/kotlin/net/thunderbird/core/ui/setting/dialog/ui/components/list/value/IconListItem.kt
wmontwe: I gave it another try with `FlowRow` but the icons are not nicely layed out. I'm going to fix the `AutoHeightLazyVerticalGrid` instead.

wmontwe: I fixed the `AutoHeightLazyVerticalGrid` to properly calculate now.

feature/account/avatar/api/src/commonMain/kotlin/net/thunderbird/feature/account/avatar/AvatarIconCatalog.kt
wmontwe: I introduced a `AvatarIcon` and a new enum to host the icons.

PR #10153, opened 2025-11-25
refactor: rename message list to main activity

### LINKED ISSUE #10085: Rename MessageList to MainActivity
This is just a suggestion.

Currently, whenever we refer to ""MessageList,"" I automatically link it to what is actually managed by the MessageListFragment.

In reality, ""MessageList"" refers to what I would call either HomeActivity or MainActivity. This component not only handles the list of messages but will also be responsible for managing the message view and other features that are or not related to the messages at all.

I believe renaming it to ""HomeScreenActivity"" would clarify that it encompasses more than just the list of messages in the app; it signifies something more significant, avoiding confusing.

wmontwe:
Resolves #10085


wmontwe:
@dani-zilla I had to update the badging to include the new `MainActivity`

dani-zilla (approved):

Issue #9497, opened 2025-07-24
Fix MessageList state handling


wmontwe:
The current message list state handling is causing frustration, resulting in frequent errors and inconsistencies across various app navigation points (e.g., drawer, widget, notification entry). 

To address this, we need to introduce a centralized state management approach via a ViewModel-based state machine, which will serve as the single source of truth for all components. 

This change also aims to provide unified account support and replace existing local state management.

rafaeltonholo:
This task is too big to be handled by a single task. We will continue working on this in the #10412.

PR #10170, opened 2025-11-28
feat: introduce state machine class

rafaeltonholo:
Part of #9497.

This PR introduces a `StateMachine` class, which allows us to define a set of states with transitions triggered by events. This will help in the refactor of the Message List, improving the state handling.


wmontwe (approved):
Looks good to me from a high-level perspective. I still need to exercise an implementation, but that shouldn’t block these changes.

PR #10171, opened 2025-11-30
Fix: Importing Settings Causes Loss of The Expected Imported Values

### LINKED ISSUE #10149: When exporting and importing "General settings" some settings are not preserved as they should be
### Checklist

### App

K-9 Mail

### App version

14

Google Play

### Android version

no

### Device model

SM-S926B/DS

### Steps to reproduce

1. In the Settings menu, select "Display" and uncheck "Show contact names".
2. In the "Settings" menu press "Export settings" and save the file.
3. Uninstall K9 and reinstall it.
4. Open K9 and import the settings from the file exported at 1.
5. The setting "Show contact names" is checked.

### Expected behavior

The setting "Show contact names" should not be checked.

### Actual behavior

The setting "Show contact names" is checked.

### Logs

_No response_

shamim-emon:
- Fixes #10149 
  ###  🪲 Cause of the Bug
 - After importing settings into the preference storage, all **PreferenceManager** instances continued using outdated preference data until the app was restarted.
 - This happened because
   - **preferenceStorages** in **PreferenceManagers** were retriving only old values and missing the imported values.
   - **PreferenceManager** instances did   not observe or react to updates in the underlying preference storage after the import operation.

 ### ✅ Solution
 - Injected **StoragePersister** into each **PreferenceManager** so that they always fetch the latest preference storage instance, ensuring updated values are returned immediately.
 - Made **PreferenceManager** instances notifiable and updated their configuration flows to refresh themselves when notified.
 - After importing settings, the system now triggers notifications to all **PreferenceManager** instances, prompting them to reload the latest configuration from the updated **preference storage**.

### DIFF
--- a/core/preference/impl/src/commonMain/kotlin/net/thunderbird/core/preference/display/visualSettings/DefaultDisplayVisualSettingsPreferenceManager.kt
+++ b/core/preference/impl/src/commonMain/kotlin/net/thunderbird/core/preference/display/visualSettings/DefaultDisplayVisualSettingsPreferenceManager.kt
@@ -15,6 +15,8 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import net.thunderbird.core.logging.Logger
+import net.thunderbird.core.preference.PreferenceChangeBroker
+import net.thunderbird.core.preference.PreferenceChangeSubscriber
 import net.thunderbird.core.preference.display.visualSettings.message.list.MessageListPreferencesManager
 import net.thunderbird.core.preference.storage.Storage
 import net.thunderbird.core.preference.storage.StorageEditor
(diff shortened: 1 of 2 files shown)


legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
```diff
@@ -57,6 +57,7 @@ internal class DefaultGeneralSettingsManager(
     private val debugLogConfigurator: DebugLogConfigurator,
     private val backgroundDispatcher: CoroutineDispatcher = Dispatchers.IO,
     private val platformConfigProvider: PlatformConfigProvider,
+    private val preferenceChangePublisher: PreferenceChangePublisher,
```
wmontwe: There is already a `changePublisher` above -> line 46

shamim-emon: updated!

PR #10198, opened 2025-12-04
docs(adr): add ADR for api internal split

### LINKED ISSUE #10197: docs(adr): refine api/impl split into api/internal
While the api/impl split is already a great step into better project organization it doesn't reflect our initial intention to split between code that is used and shared accross features and code that implements a specific feature but should not be visible to other parts of the app.

We propose clarifying and formalizing the module and package structure across Thunderbird for Android by:

* Renaming all `impl` modules to `internal` to clearly mark them as private implementation details.
* Keeping `api` modules as the only public contract surface between features and core areas.
* Establishing consistent package naming:
  * `api` modules use direct feature/core packages
  * `internal` modules mirror them under `.internal` package suffix.
* Enforcing dependency rules so only `api` modules may be referenced across areas; `internal` modules are only used by composition modules (`:app-common`, `:app-k9mail`, `:app-thunderbird`).
* Adding build checks to prevent accidental dependencies on `internal` modules.

## Tasks

- [x] Update documentation
- [ ] Build automatic checks

wmontwe:
Resolves #10197

### DIFF
--- a/docs/architecture/adr/0009-api-internal-split.md
+++ b/docs/architecture/adr/0009-api-internal-split.md
@@ -1,0 +1,138 @@
+# Feature/Core API/Internal split and dependency rules
+
+- Issue: [#10197](https://github.com/thunderbird/thunderbird-android/issues/10197)
+- Pull Request: [#10198](https://github.com/thunderbird/thunderbird-android/pull/10198)
+
+## Status
+
+- **proposed**
+
+## Context
+
+Thunderbird for Android uses a modular architecture. Many feature and core areas are split into `api` (public contracts)
+and `impl` (implementation) modules, e.g. `:feature:account:settings:api` and `:feature:account:settings:impl`.
+
+Over time, the main friction was ambiguity in naming and ownership:
+- Inconsistent use of `api` in package names made it unclear whether a type was part of a public contract.
+- Lack of a clear convention for naming feature-internal packages (`impl`, `internal`, or none) led to confusion.
+- Team wasn’t aligned on what belongs in `api` (stable contracts) vs. what should stay internal to the feature.
+
+To improve clarity and discoverability, we rename `impl` modules to `internal` and formalize module, package, and
+content rules for both API and internal code — for both feature and core modules.
+
+## Decision
+
+- `:feature:*:api` modules define the public contracts exposed to other features.
+- `:feature:*:impl` modules are renamed to `:feature:*:internal`, marking them as private implementation details.
+- `:core:*:api` modules define public contracts exposed to feature and other core modules.
+- `:core:*:impl` modules are renamed to `:core:*:internal`, marking them as private implementation details.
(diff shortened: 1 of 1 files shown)


docs/architecture/adr/0009-api-internal-split.md
```diff
@@ -0,0 +1,138 @@
...
+- Binding of contracts to implementations happens in central composition modules (application assembly): `:app-common`
+  and the app-specific modules `:app-k9mail` and `:app-thunderbird`.
+
+### What goes into API vs. Internal
+
+Put only stable, intentionally shared contracts in `api`:
+- Public interfaces and abstractions other features depend on (e.g., repositories, use cases, service interfaces).
+- Data contracts exchanged across features (DTOs/value objects).
+- Navigation contracts and events that other features can trigger/observe.
+- DI entry points/interfaces needed by composition modules (`:app-common`, app modules).
```
rafaeltonholo, line 40: There are a few cases where we might need to expose UI implementations on the `api` module (e.g.: `:feature:notification:api`, `:core:ui:compose:designsystem:api`).

Unless we start using `fun interfaces` implementations and koin injection inside the Composables which uses these shared ui implementations, I believe we should include this exception here.

rafaeltonholo: Another option could be having an `api-ui` or `ui-api` module for these cases.

docs/architecture/adr/0009-api-internal-split.md
```diff
@@ -0,0 +1,138 @@
...
+> - Avoid adding `.api` to package names for new code — the module already is the API.
+> - Prefer small focused API packages with narrow sets of contracts; keep type stability in mind when promoting code to API.
+> - API packages should remain variant-agnostic in almost all cases; concrete variants live under `.internal.<variant>`.
+
+### Dependency rules (Gradle)
+
+- Feature-to-feature dependencies must target `:feature:*:api` only.
+- Feature-to-core dependencies must target `:core:*:api` only.
+- Core-to-core dependencies must target `:core:*:api` only.
+- Core modules must not depend on `:feature:*` modules.
```
rafaeltonholo, line 101: > [!NOTE]
> We might need to add to the `Migration plan` the creation of the following modules:
> - `:core:account:api`
> - `:core:account:storage:api`
> - `:core:notification:api`
> 
> That will be required to make `:core:android:account` compliant with that rule

wmontwe: The `:core:android:account` is there for interop reasons, it needs to be further refactored. Ideally it would be a legacy module.

docs/architecture/adr/0009-api-internal-split.md
wmontwe: Some modules don't fit into the api/internal split, there could be exceptions. I will update the docs on this.

PR #10206, opened 2025-12-04
Bump version number post-merge

dandarnell:


dandarnell:
Bypassing review requirement since Stephen has approved these changes and he's not in the review group yet.

PR #10215, opened 2025-12-08
Fixes #10214 The sidebar menu is not fully displayed when there is a …

### LINKED ISSUE #10214: The sidebar menu is not fully displayed when there is a bottom navigation bar.
### Checklist

### App

Thunderbird for Android

### App version

17.0a1, 16.0b1

Other

### Android version

12

### Device model

MI U8

### Steps to reproduce

1. Open the app and tap the menu icon of top left
2. The "Setting" menu is not fully displayed 

### Expected behavior

The "Setting" menu should be fully displayed .

### Logs

_No response_

xuenhua:
Fixes #10214 The sidebar menu is not fully displayed when there is a bottom navigation bar.

### DIFF
--- a/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/setting/AccountSettingList.kt
+++ b/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/setting/AccountSettingList.kt
@@ -1,6 +1,7 @@
 package net.thunderbird.feature.navigation.drawer.dropdown.ui.setting
 
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.grid.GridItemSpan
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/setting/AccountSettingList.kt
```diff
@@ -28,7 +30,8 @@ internal fun AccountSettingList(
     SettingList(
         modifier = modifier
             .padding(vertical = MainTheme.spacings.default)
-            .fillMaxWidth(),
+            .fillMaxWidth()
+            .height(defaultThemeSizes.bottomBarHeightWithFab*3),
```
dani-zilla, line 34: If there are issues with the viewable area, it may be helpful to instead define an exclusion area for edge-to-edge apps. https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge https://developer.android.com/develop/ui/compose/system/insets
Multiplying the default bottom bar height by a constant value may work for some or even most devices, but might not look right on all displays. Window insets could help create a better viewing area for devices that have issues.

wmontwe:
@xuenhua Thanks for your fix, this helped to find the original issue. It is resolved in #10236

PR #10236, opened 2025-12-09
Fix: possible edge to edge deadlock

### LINKED ISSUE #10235: Edge-to-edge deadlock with hidden content
When entering the permission screen in the account setup the `skip` button could be hidden by the bottom nav bar. As the back button is disabled on that screen the user could not go back or continue.

While debugging the issue I found other places where the edge-to-edge setup wasn't applied, need to be fixed too.

wmontwe:
Resolves #10235 #10214

I fixed a couple of other places where the experience wasn't great too.


legacy/ui/legacy/src/main/res/layout/activity_push_info.xml
```diff
@@ -1,21 +1,34 @@
...
+        style="?attr/toolbarStyle"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:minHeight="?attr/actionBarSize"
+        tools:navigationIcon="@drawable/ic_arrow_back"
+        >
+
+        <include layout="@layout/toolbar" />
+
+    </com.google.android.material.appbar.MaterialToolbar>
```
rafaeltonholo: Causing crash. Missing the `app_bar_layout` id.
```suggestion
    <com.google.android.material.appbar.AppBarLayout
        android:id="@+id/app_bar_layout"
        style="?attr/toolbarStyle"
        android:layout_width="match_parent"
        android:layout_height="wrap_content"
        android:minHeight="?attr/actionBarSize"
        tools:navigationIcon="@drawable/ic_arrow_back"
        >

        <include layout="@layout/toolbar" />

    </com.google.android.material.appbar.AppBarLayout>
```

legacy/ui/legacy/src/main/res/layout/activity_push_info.xml
```diff
@@ -1,21 +1,34 @@
...
 
     <androidx.fragment.app.FragmentContainerView
         android:id="@+id/fragment_container"
```
rafaeltonholo: Causing crash. Need to use `content_container` instead.
```suggestion
        android:id="@+id/content_container"
```

legacy/ui/legacy/src/main/res/layout/activity_push_info.xml
wmontwe: Copy paste issue 😞

legacy/ui/legacy/src/main/res/layout/activity_push_info.xml
wmontwe: Copy paste issue 😞

legacy/ui/legacy/src/main/res/layout/activity_push_info.xml
rafaeltonholo: it happens haha

PR #10254, opened 2025-12-11
refactor: migrate account setting key with migration logic

### LINKED ISSUE #10145: Account Setting Key Migration
As [suggested ](https://github.com/thunderbird/thunderbird-android/pull/10077#discussion_r2528575680) during the review of PR #10077, the preference key used for the “Expand all folders by default” setting should be renamed to better match its corresponding title.

However, this preference is already in use by existing users, which means renaming the key outright would cause their saved value to be lost. To avoid breaking existing behavior, we need to introduce a migration strategy.

**Tasks**:
- Define a new, clearer key name that aligns with the setting title.
- Implement a migration step that:
   - Reads the value stored under the old key.
   - Writes it into the new key.
   - Removes (or safely ignores) the old key afterward.
- Update all references to use the new key.
- Verify preference loading and default values behave as expected after migration.

shamim-emon:
- Fixes #10145
### Description
This PR migrates an existing account setting preference key to a new name and ensures a smooth transition for existing users by adding migration logic. This prevents loss of user preferences when the key name changes.
### Changes
 - Preference Key Rename
   -  Old key: "account_setup_auto_expand_folder"
   -  New key: "auto_select_folder"
- Migration Implementation
  - Added the migration in  `StorageMigrationTo29` to automatically migrate values from the old key to the new key for existing users. 
- Test
  - Added unit test to verify the migration logic works correctly.

### DIFF
--- a/legacy/storage/src/test/java/com/fsck/k9/preferences/migration/StorageMigrationTo29Test.kt
+++ b/legacy/storage/src/test/java/com/fsck/k9/preferences/migration/StorageMigrationTo29Test.kt
@@ -1,0 +1,45 @@
+package com.fsck.k9.preferences.migration
+
+import assertk.assertThat
+import assertk.assertions.doesNotContainKey
+import assertk.assertions.isEqualTo
+import assertk.assertions.key
+import com.fsck.k9.preferences.createPreferencesDatabase
+import kotlin.test.Test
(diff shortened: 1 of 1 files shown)


wmontwe (approved):
Thank you!

PR #10259, opened 2025-12-13
fix: display Settings reset/mix up when toggling "Colorize contact pictures"

### LINKED ISSUE #9630: Display Settings reset/mix up when toggling "Colorize contact pictures"
### Checklist

### App

Thunderbird for Android

### App version

12.0b7

Other

### Android version

15

### Device model

Samsung Galaxy S23

### Steps to reproduce

1. Open the Thunderbird app
2. Go to **Settings** → **Display Settings**
3. Unselect Colorize contact pictures
4. Go back to the inbox
5. Return to Settings → **Display Settings**

Image
Image

### Expected behavior

- Colorize contact pictures should remain _unselected_
- Show contact names should stay _selected_ (unchanged)

### Actual behavior

- **Colorize contact pictures** has been re-enabled
- **Show contact names** has been _automatically unselected_, even though it was never changed by the user

Image

### Logs

_No response_

shamim-emon:
- Fixes #9630 
### 🐛 Root Cause of the Bug
 - All Message List–related display settings were not being loaded from PreferenceStorage; instead, they were retrieved from the default values defined in their corresponding Settings class.
 -  Additionally, whenever Message List–related display settings were being saved, they were not persisted because the corresponding storage editors were not committing the changes.
### 💡 Solution
 - Integrate the message list preferences so they combine correctly with existing general and display settings.
 -  Commit changes to the preference store when the corresponding settings are saved.

### DIFF
--- a/legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
+++ b/legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
@@ -14,6 +14,7 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import net.thunderbird.core.common.appConfig.PlatformConfigProvider
+import net.thunderbird.core.logging.Logger
 import net.thunderbird.core.logging.config.DebugLogConfigurator
 import net.thunderbird.core.preference.GeneralSettings
 import net.thunderbird.core.preference.GeneralSettingsManager
@@ -24,6 +25,7 @@ import net.thunderbird.core.preference.display.coreSettings.DisplayCoreSettingsP
 import net.thunderbird.core.preference.display.inboxSettings.DisplayInboxSettingsPreferenceManager
 import net.thunderbird.core.preference.display.miscSettings.DisplayMiscSettingsPreferenceManager
 import net.thunderbird.core.preference.display.visualSettings.DisplayVisualSettingsPreferenceManager
+import net.thunderbird.core.preference.display.visualSettings.message.list.MessageListPreferencesManager
 import net.thunderbird.core.preference.interaction.InteractionSettingsPreferenceManager
 import net.thunderbird.core.preference.network.NetworkSettingsPreferenceManager
 import net.thunderbird.core.preference.notification.NotificationPreferenceManager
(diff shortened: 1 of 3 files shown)


legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
```diff
@@ -93,6 +96,16 @@ internal class DefaultGeneralSettingsManager(
                 display = generalSettings.display.copy(visualSettings = visualSettings),
             )
         }
+        .combine(messageListPreferencesManager.getConfigFlow()) { generalSettings, messageListSettings ->
+            logger.debug { "messageListSettings: $messageListSettings" }
```
dani-zilla, line 101: Worth noting that logger is currently not recommended:
```
"Use a net.thunderbird.core.logging.Logger instance via dependency injection instead. " +
        "This class will be removed in a future release.",
```

dani-zilla (changes requested):
I think we could drop the logging, but I did want to ask if you were able to reproduce this in a newer version of the app. I could not reproduce it on v14 from Google Play or v17 from a fresh debug build, not with the existing views or compose either. Do you have any additional info on how this fix could be tested between the latest beta and this version? 

Thanks!

legacy/core/src/main/java/com/fsck/k9/preferences/DefaultGeneralSettingsManager.kt
shamim-emon: Updated, now injecting  net.thunderbird.core.logging.Logger instance instead as per suggestion.

shamim-emon:
> I think we could drop the logging, but I did want to ask if you were able to reproduce this in a newer version of the app. I could not reproduce it on v14 from Google Play or v17 from a fresh debug build, not with the existing views or compose either. Do you have any additional info on how this fix could be tested between the latest beta and this version?
> 
> Thanks!

@dani-zilla  I am able to reproduce this on latest debug build.  In-case you don't see the issue, toggle  Colorize contact pictures value and then restart the app, you will see this value along with any value from MessageList settings is not persisted.

dani-zilla (approved):
Was able to reproduce it with the force quit method. Thanks for that! Bug seems to be fixed in build and looks good to me. Thanks!

PR #10301, opened 2025-12-27
feat: configure notification actions

harshad1:
This PR addresses #3530

* Added a drag UI to select available actions and their order
* The Existing delete option has been updated to only be for summary notifications

image

image

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/notificationactions/NotificationActionsSettingsScreen.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/notificationactions/NotificationActionsSettingsScreen.kt
@@ -250,8 +250,7 @@ private fun NotificationActionRow(
             text = contentLabel,
             modifier = Modifier
                 .weight(1f)
-                .padding(start = MainTheme.spacings.triple, end = MainTheme.spacings.default)
-                .clearAndSetSemantics { },
+                .padding(start = MainTheme.spacings.triple, end = MainTheme.spacings.default),
         )
         Row(
             modifier = Modifier.padding(end = MainTheme.spacings.default),
@@ -259,13 +258,13 @@ private fun NotificationActionRow(
         ) {
             ArrowButton(
                 imageVector = Icons.Outlined.ExpandLess,
-                contentDescription = "$contentLabel. $itemStateDescription. ${buttonLabels.moveUpLabel}",
+                contentDescription = "$itemStateDescription. ${buttonLabels.moveUpLabel}",
                 enabled = rowState.canMoveUp,
                 onClick = buttonLabels.onMoveUp,
             )
             ArrowButton(
                 imageVector = Icons.Outlined.ExpandMore,
-                contentDescription = "$contentLabel. $itemStateDescription. ${buttonLabels.moveDownLabel}",
+                contentDescription = "$itemStateDescription. ${buttonLabels.moveDownLabel}",
                 enabled = rowState.canMoveDown,
                 onClick = buttonLabels.onMoveDown,
             )


harshad1:
In the future we can have additional configurations for

1. Summary messages (only some actions make sense - delete, archive, spam, mark read)
2. Other surfaces (android wear) where the number of actions can be > 3

wilcooo:
Great work! I hope to see this included in the app. I've some opinions about the details here.

I would remove the ability to drag the separation line. It might be slightly confusing since the line itself isn't an element on the notification.

Maybe even replace the grab icons with + and - depending on whether the option is included at the moment, to make it even more clear which options will be included. (clicking plus/minus moves it to the other side of the line).

The main objective here is including/excluding options and should be visually obvious. Reordering is secondary and can be explained in text (as you have already).

To make using the + and - buttons more consistent, you can remember the complete order of all 5 items. So that removing one and adding it again returns it to the same spot. e.g. the order can be EBACD, but only B and D are shown so what's visible in the UI is BD/EAC, and dragging will then be the only way to change the underlying order.

ckardaris:
@wilcooo I will have to disagree.
> I would remove the ability to drag the separation line. It might be slightly confusing since the line itself isn't an element on the notification.
> 
> Maybe even replace the grab icons with + and - depending on whether the option is included at the moment, to make it even more clear which options will be included. (clicking plus/minus moves it to the other side of the line).

While first reading your suggestion I was thinking that your suggestions make sense and would probably improve the UX. Less moving parts, so it would be clean. **But** only if the order of elements is fixed and non-configurable by the user. If re-ordering is allowed, then that would mean that there would be a "hidden" functionality when holding and sliding a line, which maybe not be very intuitive unless clearly explained in the top description.

In my opinion, the current UI with the "dimmed" color for the hidden options and normal color for the shown is intuitive enough. Taking your suggestions into consideration maybe we could have the separator not movable at all on its own and instead it could span the whole width of the screen with no margin left/right. This would mean that we would not be able to enable/disable multiple elements with one slide, but I don't consider this use case very useful anyway.

image1

Having said that, @harshad1 have you considered how your solution behaves with regards to accessibility?

harshad1:
Some options:

1. Keep the current UI, but remove the dragability of the line
2. Move to a checkbox centric approach - remove the line, add a checkbox per item and enforce a maximum of n (= 3) checked items.

Option 2 is similar to this UI I implemented in Markor (an open source text editor I contribute to) 
Screenshot_20251228-144913

@ckardaris I have not considered accessibility. Thanks for bringing that up. I'll investigate.

harshad1:
I have made a few changes:

1. The bar behaves similarly but doesn't have a drag handle (should be clear that this isn't an item)
2. Added Accessibility helpers
3. Added star as an option
4. cleaned up the code (single source of truth)

The bar remains draggable. This makes it easier to control the number of actions, makes the code simpler etc

IMO this is now quite clear and usable

screen-20251229-174052-1767058832656

harshad1:
Is there anything I can do to help move this forward?

This is my first Thunderbird-android contribution and I am unfamiliar with the SOP here.

ty!

dani-zilla:
Thanks for putting this together! Just giving you an update, we’re having the UX team go over the new design. They’ll potentially have some feedback, so you should hold off on any changes until they review this. 

My first consideration is that the new UI should be built in [compose for any new screens](https://github.com/thunderbird/thunderbird-android/blob/main/docs/architecture/ui-architecture.md). Just hold off on starting it until we have design feedback.

I do want to raise concerns over the accessibility of a drag & drop interface. Drag & drop _looks_ intuitive, and creates a visually pleasing interface (which I think this one is), but users with vision and mobility impairments find it difficult to interact with drag & drop. TalkBack would have to be set up to read what’s above and below the line on any changes to the ordering or inclusion. Likely with a debouncer so we wait to make sure they’re not still editing quickly between read-outs. Also we’d want to consider using [accessibility actions](https://developer.android.com/guide/topics/ui/accessibility/principles#accessibility-actions) to make the drag and drop feature work as smoothly for someone using TalkBack or pointing aids as it would for the rest of our users.

My recommendation might be to go with something more akin to [this Microsoft example](https://medium.com/microsoft-mobile-engineering/accessible-reordering-for-touch-devices-e7f7a7ef404). By using arrows to re-order items, we can make it easier to make individual changes to ordering one at a time, and also drop the requirement for a consistent touch to drag and reorder items. We could also consider making the reorder buttons only available above the line, as they’re the only ones that it would matter for reordering, by using the previously suggested checkbox design. If an item is checked off, it appears above the line, with the other checkboxes disabled after 3 are selected. 

However, I strongly recommend waiting until UX gets back to us with what they think. I’d hate for you to have to make more changes again. Just wanted to plant some ideas on what I was thinking here along with the update. Thanks again for this!

harshad1:
@dani-zilla thanks!

For reference I did add the following

 - Each action row and the cutoff divider exposes TalkBack actions “Move up” and “Move down” via ViewCompat.replaceAccessibilityAction, so reorder works without drag.
 - The cutoff row has a contentDescription, and the drag handle has its own description.
 - Actions trigger the same move logic as drag, respecting the max cutoff position.

Thanks for the example. I'll look into it when UX gets back.

laurelterlesky:
Hi all, Design Team chiming in here. I don't have anything more to add on top of what's been said already. Let's get this out there and then see if we need to further iterate. Thanks for the contributions!

harshad1:
@dani-zilla is there anything I need to do to enable workflows? Ty

dani-zilla:
Enabled the workflow. Looks like there are a few issues with Spotless, Detekt, and tests. 

We also aren't introducing new features with XML views, opting instead for Jetpack Compose for views. Now that we know the design is good, I think it's safe to tackle that.

harshad1:
@dani-zilla It appears that workflows need to be approved again.

I have moved the UI to compose and fixed the CI issues (I believe)

core/common/src/commonMain/kotlin/net/thunderbird/core/common/notification/NotificationActionTokens.kt
```diff
@@ -0,0 +1,15 @@
...
+ */
+object NotificationActionTokens {
+    const val REPLY = "reply"
+    const val MARK_AS_READ = "mark_as_read"
+    const val DELETE = "delete"
+    const val STAR = "star"
+    const val ARCHIVE = "archive"
+    const val SPAM = "spam"
+
+    const val DEFAULT_ORDER = "reply,mark_as_read,delete,star,archive,spam"
```
dani-zilla: These should use the constants defined above so the string doesn't have to be changed if the keys change or are added to

dani-zilla: Would this potentially be better as a List rather than a comma-separated string? My guess is that it's to make it easier to write to preferences as a string. Though it might be easier to only do the parsing on read/write, rather than when passing the data around and making changes.

legacy/core/src/main/java/com/fsck/k9/notification/SingleMessageNotificationDataCreator.kt
```diff
@@ -76,13 +83,81 @@ internal class SingleMessageNotificationDataCreator(
...
+        }
+
+        return filled
+    }
+
+    private fun parseActionsOrder(raw: String): List<NotificationAction> {
+        val tokens = raw
+            .split(',')
+            .map { it.trim() }
+            .filter { it.isNotEmpty() }
```
dani-zilla: Noted this above with the default order constant, but what is the purpose of doing this as a string instead of keeping it as a list?

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/notificationactions/MessageNotificationAction.kt
```diff
@@ -0,0 +1,64 @@
...
+        iconRes = Icons.Outlined.Report,
+    ),
+    ;
+
+    companion object {
+        fun fromToken(token: String): MessageNotificationAction? {
+            return entries.firstOrNull { it.token == token }
+        }
+
+        fun defaultOrder(): List<MessageNotificationAction> = listOf(
```
dani-zilla: This does seem to mean we define the default order in two places. Seeing as this class already depends on the NotificationActionTokens, maybe it could just parse the string or expose the list here?

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/notificationactions/NotificationActionsAdapter.kt
```diff
@@ -0,0 +1,448 @@
...
+    val outValue = TypedValue()
+    val resolved = context.theme.resolveAttribute(android.R.attr.listPreferredItemHeightSmall, outValue, true)
+    val px = if (resolved) {
+        if (outValue.resourceId != 0) {
+            context.resources.getDimension(outValue.resourceId)
+        } else {
+            TypedValue.complexToDimension(outValue.data, context.resources.displayMetrics)
+        }
+    } else {
+        48f * context.resources.displayMetrics.density
```
dani-zilla: Not a fan of setting constants in the code like this. I'd say either define it as a constant or we could set the minimum list item height a different way

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/notificationactions/NotificationActionsAdapter.kt
```diff
@@ -0,0 +1,448 @@
...
+        )
+        TextBodyLarge(
+            text = stringResource(action.labelRes),
+            modifier = Modifier
+                .weight(1f)
+                .padding(start = MainTheme.spacings.triple, end = MainTheme.spacings.double),
+        )
+        Box(
+            modifier = Modifier
+                .width(56.dp)
```
dani-zilla: Constants for the items should be either made as constant somewhere or not used and let the relation to the other elements control the size. 

You could consider breaking this out into its own composable and using it that way. We do use an atomic structure, where we break down screens into atoms, the smallest reusable parts, and build up from there. It's described [here](https://github.com/thunderbird/thunderbird-android/blob/main/docs/architecture/design-system.md). This could allow the constants to live in their atomic components, and will shrink the implementation of those components in a screen as well.

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/notificationactions/NotificationActionsAdapter.kt
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    companion object {
+        private const val VIEW_TYPE_ACTION = 0
+        private const val VIEW_TYPE_CUTOFF = 1
+    }
+}
+
+@Composable
+private fun NotificationActionRow(
```
dani-zilla: I definitely would prefer to see the components split out and templated for easier re-use.

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/notificationactions/NotificationActionsAdapter.kt
```diff
@@ -0,0 +1,448 @@
...
+internal sealed interface NotificationListItem {
+    data class Action(
+        val action: MessageNotificationAction,
+        val isDimmed: Boolean,
+    ) : NotificationListItem
+
+    object Cutoff : NotificationListItem
+}
+
+internal class NotificationActionsAdapter(
```
dani-zilla: I was hoping to see the entire view move to compose. Replacing inflated views and recycler views with lazy columns, that sort of thing. That would reduce the complexity and file size as well.

harshad1:
just FYI for any watching  - I'm still working on this and do plan to address comments and push things through as time permits

core/common/src/commonMain/kotlin/net/thunderbird/core/common/notification/NotificationActionTokens.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+object NotificationActionTokens {
+    const val REPLY = "reply"
+    const val MARK_AS_READ = "mark_as_read"
+    const val DELETE = "delete"
+    const val STAR = "star"
+    const val ARCHIVE = "archive"
+    const val SPAM = "spam"
+
+    val DEFAULT_ORDER: List<String> = listOf(REPLY, MARK_AS_READ, DELETE, STAR, ARCHIVE, SPAM)
```
harshad1, line 14: Single source of truth for default order

harshad1:
- Order tokens now come from NotificationActionTokens, and parsing is centralized.
  - messageActionsOrder is now a list in the preference model; we only serialize at the storage boundary.
  - messageActionsOrder is a List<String> in NotificationPreference. Parsing happens only in DefaultNotificationPreferenceManager
- Min height uses MainTheme.sizes.iconAvatar inside composables.


- Tried full Compose; drag was janky and I couldn't get it to animate / move smoothly (but I am not familiar with compose)
  - Reverted to RecyclerView + ItemTouchHelper

dani-zilla:
>     * Tried full Compose; drag was janky and I couldn't get it to animate / move smoothly (but I am not familiar with compose)
>       
>       * Reverted to RecyclerView + ItemTouchHelper

Thanks for giving it a shot. So, we don’t want to merge anything using old xml views. Obviously doing small modifications to existing views is fine, and we wouldn’t currently ask anyone submitting changes to a view to update it to Jetpack Compose unless it was a large change, however, for all new views, we’re using Compose with an ["atomic" design system](https://github.com/thunderbird/thunderbird-android/blob/main/docs/architecture/adr/0002-ui-wrap-material-components-in-atomic-design-system.md). More info on that is [here](https://github.com/thunderbird/thunderbird-android/blob/main/docs/architecture/design-system.md). 

I think we could look at two options here. The first is to do it in compose, jankiness and all, and work to smooth it out before merging. We can figure out what’s causing the stuttering behavior and see if we can fix it. Most likely just a mutable data object or some other item causing frequent recomposing. I can’t tell what the issue may have been because it doesn’t look like it was committed, but we can look into it if you want to give it another go. 

The second option is to change the way we're displaying this list of options entirely, if the drag & drop itself is the issue. We could go back to the older idea of using arrow buttons. This is a more static event, with each item only being reordered by one place at a time, and could likely reduce the jank. It might also work better for people using TalkBack.

harshad1:
After some more effort I think I have the UI ported to compose. I also replaced the drag handle with buttons.

screen-20260215-183002-1771208978636

legacy/core/src/main/java/com/fsck/k9/notification/NotificationActionIntents.kt
```diff
@@ -0,0 +1,119 @@
...
+internal const val ACTION_DELETE = "ACTION_DELETE"
+internal const val ACTION_ARCHIVE = "ACTION_ARCHIVE"
+internal const val ACTION_SPAM = "ACTION_SPAM"
+internal const val ACTION_STAR = "ACTION_STAR"
+internal const val ACTION_DISMISS = "ACTION_DISMISS"
+internal const val EXTRA_ACCOUNT_UUID = "accountUuid"
+internal const val EXTRA_MESSAGE_REFERENCE = "messageReference"
+internal const val EXTRA_MESSAGE_REFERENCES = "messageReferences"
+
+object NotificationActionIntents {
```
harshad1, line 19: Moved several functions here from NotificationActionService.kt as having them there exceeded the maximum functions lint rule

harshad1:
@dani-zilla Could you run the workflows again? Can they just be enabled so I can run them and not have to ask you?

This is ready for review again

harshad1:
I fixed the failing tests

dani-zilla:
https://github.com/user-attachments/assets/85c2a991-ed37-46f6-aa0e-c47abcddf467

Still going through a larger review/discussion about this with the team, but I did want to quickly call out an accessibility issue I noticed. 

The instructions for drag order are read after going through the list all the way once. Also, when going through the list the first time, the only way to access the controls is to activate them with up/down swipes (if the user has set this action to select controls, specifically, as I did in this recording). It's only on the second pass through the list that the user can select the buttons. 

I think there's some tweaking to do here concerning tab ordering. I wanted to go through it more on my own, but time is a bit limited and I haven't gotten the chance, so I thought I'd share it to let you see what I noticed and get a chance to work on it before I wrap up this review. Hope this video of what I noticed helps. Thanks!!

harshad1:
Good catch. I will address this this weekend. Still not quite used to compose :)

harshad1:
Fixed conflicts. Please review again.

dani-zilla:
https://github.com/user-attachments/assets/9f141b18-f623-4898-bc5a-43ded9f95e53

This seems to now make it much more difficult to select the up/down arrows. The user can't tab to them and instead must select the entire line, triple tap, select actions, then select the action they want to perform, then double tap to perform the action. That's unfortunately quite a lot of work required to do something like reordering, which they'd presumably have to do multiple times for each item

harshad1:
I reworked the accessibility so that it works only on the buttons. IMO this is cleaner and much easier to navigate and use.

dani-zilla:
> I reworked the accessibility so that it works only on the buttons. IMO this is cleaner and much easier to navigate and use.

It might be. Although it is unconventional. Might want to check in with the design team. My thought is that it should read each item, then the controls. Tapping on a control, if it changes its state, in or out of the list, would then read out. 

For example, if Spam is out of the notification, it could read:

"Spam, not shown in the notifications" <Swipe over to up arrow> "Move Spam Up" <Tapping on the control> "Spam, shown in the notification"

This would require reading each line's label separate from the controls and speaking after a control is selected if its state in the notification has changed. And trust me, I know how tricky accessibility can be, so I'm guessing this is something you tried already and couldn't get working. 

I'm not sure how to do this without making it too verbose. I may play around with it myself a little and look into the WCAG guidelines. Usually we separate the announcement of what an item is from the control, making it easier to swipe through based on list items. Someone may not want to swipe across the controls for delete, for example, they might hear delete and want to skip to the next line item without addressing the controls. This might require a bit of research on our part, but we are looking into it, just wanted to make sure we gave you an update. And thanks again for taking a crack at it!

dani-zilla:
I decided to take a quick look at this today and I think this works. I can look more Monday, but had a bit of time and decided to take a look. 

```
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/notificationactions/NotificationActionsSettingsScreen.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/notificationactions/NotificationActionsSettingsScreen.kt
@@ -30,6 +30,7 @@ import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.SemanticsProperties
 import androidx.compose.ui.semantics.clearAndSetSemantics
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.zIndex
@@ -250,8 +251,7 @@ private fun NotificationActionRow(
             text = contentLabel,
             modifier = Modifier
                 .weight(1f)
-                .padding(start = MainTheme.spacings.triple, end = MainTheme.spacings.default)
-                .clearAndSetSemantics { },
+                .padding(start = MainTheme.spacings.triple, end = MainTheme.spacings.default),
         )
         Row(
             modifier = Modifier.padding(end = MainTheme.spacings.default),
@@ -259,13 +259,13 @@ private fun NotificationActionRow(
         ) {
             ArrowButton(
                 imageVector = Icons.Outlined.ExpandLess,
-                contentDescription = "$contentLabel. $itemStateDescription. ${buttonLabels.moveUpLabel}",
+                contentDescription = "$itemStateDescription. ${buttonLabels.moveUpLabel}",
                 enabled = rowState.canMoveUp,
                 onClick = buttonLabels.onMoveUp,
             )
             ArrowButton(
                 imageVector = Icons.Outlined.ExpandMore,
-                contentDescription = "$contentLabel. $itemStateDescription. ${buttonLabels.moveDownLabel}",
+                contentDescription = "$itemStateDescription. ${buttonLabels.moveDownLabel}",
                 enabled = rowState.canMoveDown,
                 onClick = buttonLabels.onMoveDown,
             )
```

The idea is to not clear the accessibility role from the label for the row and remove the content label from the arrow buttons.


As for the spotless check, it's silly, but it's about reordering the inputs to this:
```
...
import androidx.compose.ui.draw.alpha
import androidx.compose.ui.graphics.graphicsLayer
import androidx.compose.ui.graphics.vector.ImageVector
import androidx.compose.ui.input.pointer.pointerInput
...
```

You can run `./gradlew spotlessApply` to do it automatically.

harshad1:
The way this works now is that swiping with talkback on jumps between the buttons only 

When it selects a button you hear
- the name of the currently selected action
- whether the action is enabled or disabled 
- whether the button is enabled or disabled

Activating the button announces a state change if it occured

So we have almost exactly this:
> "Spam, not shown in the notifications" "Move Spam Up" "Spam, shown in the notification"

Would it be possible to merge this in as 'good enough' and iterate on accessibility after, if needed.

dani-zilla:
> The way this works now is that swiping with talkback on jumps between the buttons only
> 
> When it selects a button you hear
> 
>     * the name of the currently selected action
> 
>     * whether the action is enabled or disabled
> 
>     * whether the button is enabled or disabled
> 
> 
> Activating the button announces a state change if it occured
> 
> So we have almost exactly this:
> 
> > "Spam, not shown in the notifications" "Move Spam Up" "Spam, shown in the notification"
> 
> Would it be possible to merge this in as 'good enough' and iterate on accessibility after, if needed.

The changes I suggested amount to very small changes and make the labels work. The reason we want them read out is because people will use talkback differently, and not everyone who is using it will be completely blind. You won't want to skip parts of the screen they may want to read. Besides, talkback users can also select the action performed when swiping up or down. This is usually headings or controls for most users. They can already jump through the controls this way if they want to. But a user would expect the label to be read separately from the control. Optimally, at some point, we could look into making those labels easier to jump between as well, but reading them all out and, on the first item, telling the user how many items are in the list could be very helpful. I played with this a bit more this morning and think the change to re-enable the labels might be that good starting ground to merge and improve upon later, if we can.

harshad1:
Thanks for the explanation. I have added your suggested change.

dani-zilla:
Thanks for getting this across the finish line! I know it was a lot, and a large project, but it looks great. Thanks again!

harshad1:
@dani-zilla Thank you for the feedback and reviews and (especially) your patience!

PR #10326, opened 2026-01-07
docs: add preference migration guide

### LINKED ISSUE #10325: Document the preference migration
In #10254 a new migration was introduced, meaning the existing preference migration system is not discoverable. As we can't fix this quickly, we need to document how to migrate app preferences.

wmontwe:
Resolves #10325

### DIFF
--- a/docs/developer/preference-migration-guide.md
+++ b/docs/developer/preference-migration-guide.md
@@ -1,0 +1,177 @@
+# Preference Migration Guide
+
+This document explains how to update and migrate preferences in Thunderbird for Android.
+
+## Overview
+
+Thunderbird for Android uses a dedicated SQLite database to store its preferences as key-value pairs.
+Both the database file and the primary table are named `preferences_storage`. This is managed by the
+`K9StoragePersister`.
+
+- **Database Name:** `preferences_storage`
+- **Storage Table:** `preferences_storage`
+- **Columns:** `primkey` (TEXT), `value` (TEXT)
+- **Persister:** `legacy/storage/src/main/java/com/fsck/k9/preferences/K9StoragePersister.java`
+- **Migration Manager:** `legacy/storage/src/main/java/com/fsck/k9/preferences/migration/StorageMigrations.kt`
+
+## When to use a Preference Migration
+
+You must implement a preference migration whenever you:
+
+- Rename a preference key.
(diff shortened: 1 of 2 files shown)


docs/developer/preference-migration-guide.md
```diff
@@ -0,0 +1,126 @@
...
+## Adding a New Preference
+
+If you are adding a completely new preference, you don't necessarily need a migration unless you need to populate it 
+with a default value that depends on other existing settings.
+
+To add a new preference:
+
+1. Define the key in the relevant component.
+2. Use the `GeneralSettingsManager` or a specific preference manager to read and write the value.
+3. If using the legacy `Storage` system directly, use `StorageEditor` to save the value.
```
rafaeltonholo, line 35: I also would include:

4. Bump up the preferences' version.

docs/developer/preference-migration-guide.md
```diff
@@ -0,0 +1,126 @@
...
+    }
+}
+```
+
+## Best Practices
+
+- **Idempotency:** Migrations should be safe to run multiple times, although the framework typically ensures they run only once.
+- **No Application Logic:** Do not use `Preferences` or other high-level application classes inside migrations, as their behavior might change over time. Use `StorageMigrationHelper` for direct database access.
+- **Self-contained:** Keep the migration logic simple and contained within its class.
+- **Testing:** Always write unit tests for your migrations. You can find examples in `legacy/storage/src/test/java/com/fsck/k9/preferences/migration/`.
```
rafaeltonholo: I would also include:

- Avoid doing network calls during the migration, as they may fail due to external factors. In case it is really necessary, properly handle network failures to avoid data wipeout.

rafaeltonholo (commented):
I would also include a section talking about writing unit tests for the Migration, and set it as a requirement instead of just mentioning it as a Best practice.

docs/developer/preference-migration-guide.md
wmontwe: Only necessary when a default value needs to be set that is derived from other settings.

I update the docs.

rafaeltonholo (dismissed):
👍

wmontwe:
Sorry updated again, forgot to push the test section.

PR #10329, opened 2026-01-07
 feat(message-list): create the Message List state machine handling all states and transitions

rafaeltonholo:
Part of #9497 

- Adds the new `MessageListFragment` and it's MVi contract
- Adds a new `StateMachineViewModel` for screens that will use a state machine to handle their state
- Create the Message List state machine with its states and transitions

### DIFF
--- a/feature/mail/message/list/internal/src/test/kotlin/net/thunderbird/feature/mail/message/list/internal/ui/state/machine/MessageListStateMachineTest.kt
+++ b/feature/mail/message/list/internal/src/test/kotlin/net/thunderbird/feature/mail/message/list/internal/ui/state/machine/MessageListStateMachineTest.kt
@@ -1,0 +1,719 @@
+package net.thunderbird.feature.mail.message.list.internal.ui.state.machine
+
+import androidx.compose.ui.graphics.Color
+import app.cash.turbine.test
+import assertk.all
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import assertk.assertions.isInstanceOf
+import assertk.assertions.isNull
+import assertk.assertions.isTrue
+import assertk.assertions.prop
+import dev.mokkery.spy
+import dev.mokkery.verify
+import dev.mokkery.verify.VerifyMode
+import kotlin.test.Test
+import kotlinx.collections.immutable.ImmutableList
+import kotlinx.collections.immutable.ImmutableSet
+import kotlinx.collections.immutable.persistentListOf
+import kotlinx.collections.immutable.persistentSetOf
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.TestScope
+import kotlinx.coroutines.test.advanceUntilIdle
+import kotlinx.coroutines.test.runTest
+import net.thunderbird.core.common.action.SwipeAction
+import net.thunderbird.core.common.action.SwipeActions
+import net.thunderbird.core.preference.display.visualSettings.message.list.UiDensity
+import net.thunderbird.feature.account.AccountId
+import net.thunderbird.feature.account.AccountIdFactory
+import net.thunderbird.feature.mail.message.list.preferences.ActionRequiringUserConfirmation
+import net.thunderbird.feature.mail.message.list.preferences.MessageListDateTimeFormat
+import net.thunderbird.feature.mail.message.list.preferences.MessageListPreferences
+import net.thunderbird.feature.mail.message.list.ui.event.MessageItemEvent
+import net.thunderbird.feature.mail.message.list.ui.event.MessageListEvent
+import net.thunderbird.feature.mail.message.list.ui.event.MessageListSearchEvent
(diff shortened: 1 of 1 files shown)


core/ui/compose/common/src/main/kotlin/net/thunderbird/core/ui/compose/common/mvi/BaseStateMachineViewModel.kt
rafaeltonholo, line 1: Missing unit test

feature/mail/message/list/internal/src/main/kotlin/net/thunderbird/feature/mail/message/list/internal/ui/state/machine/MessageListStateMachine.kt
rafaeltonholo, line 1: Missing unit test. The tests should ensure the transition between states when an event happens.

core/common/src/commonMain/kotlin/net/thunderbird/core/common/state/sideeffect/StateSideEffectHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+
+import kotlinx.coroutines.CoroutineScope
+import net.thunderbird.core.logging.Logger
+
+/**
+ * A base class for handling side effects that arise from state transitions in the message list UI.
+ *
+ * Each implementation of this class is responsible for a specific side effect, such as loading more messages,
+ * marking messages as read, or showing notifications. The handler decides whether to act based on the
+ * incoming event and the resulting state change.
```
wmontwe: This still references the message list UI, but it's now located within common and should be agnostic.

core/ui/compose/common/src/main/kotlin/net/thunderbird/core/ui/compose/common/mvi/StateMachineViewModel.kt
```diff
@@ -0,0 +1,114 @@
...
+ *  UI state of a screen.
+ * @param TEvent The type that represents user actions that can be handled by the ViewModel.
+ *  For example, a button click.
+ * @param TUiSideEffect The type that represents one-time side effects that can occur
+ *  in response to state changes. For example, a navigation event or showing a toast message.
+ * @param logger The [Logger] instance for logging events and state changes.
+ * @param sideEffectHandlersFactories A list of factories for creating [StateSideEffectHandler]s.
+ *  These handlers can be used to trigger side effects in response to state transitions.
+ */
+abstract class StateMachineViewModel<TState : Any, TEvent : Any, TUiSideEffect>(
```
wmontwe: This should be called `BaseStateMachineViewModel`

feature/mail/message/list/internal/src/main/kotlin/net/thunderbird/feature/mail/message/list/internal/ui/state/machine/SetupGlobalState.kt
```diff
@@ -0,0 +1,52 @@
...
+    state<MessageListState> {
+        transition<MessageListEvent.UpdatePreferences> { state, event ->
+            when (state) {
+                is MessageListState.LoadedMessages -> state.copy(preferences = event.preferences)
+                is MessageListState.LoadingMessages -> state.copy(preferences = event.preferences)
+                is MessageListState.SearchingMessages -> state.copy(preferences = event.preferences)
+                is MessageListState.SelectingMessages -> state.copy(preferences = event.preferences)
+                is MessageListState.WarmingUp -> state.copy(preferences = event.preferences)
+            }
+        }
```
wmontwe, line 21: Optional: I see that this is necessary to use `copy` on the right type. Maybe you could introduce something like:

```kotlin
fun MessageListState.withPreferences(
    preferences: Preferences
): MessageListState = when (this) {
    is LoadedMessages -> copy(preferences = preferences)
    is LoadingMessages -> copy(preferences = preferences)
    is SearchingMessages -> copy(preferences = preferences)
    is SelectingMessages -> copy(preferences = preferences)
    is WarmingUp -> copy(preferences = preferences)
}
```

and:

```suggestion
        transition<MessageListEvent.UpdatePreferences> { state, event ->
            state.withPreferences(event.preferences)
        }
```

rafaeltonholo:
> Looks good to me.
> 
> Observation:
> 
> A significant amount of data is carried over unchanged between state transitions, which leads to a large amount of repeated code whose sole purpose is to identify the correct state type.
> 
> Going beyond the `withPreferences` approach, this could be addressed by introducing a shared `MessageListContent` (or `Metadata` / `Snapshot`) that groups the common data. Each `MessageListState` would then expose a dedicated
> 
> ```kotlin
> fun updateMetadata(
>     transform: (MessageListMetadata) -> MessageListMetadata
> ): MessageListState
> ```
> 
> allowing shared data to be updated across all states without duplicating state-specific branching logic.

Thanks for the input. I've applied both suggestions, please take a look again

PR #10334, opened 2026-01-08
feat(message-list): add state side effect handler for loading swipe actions

rafaeltonholo:
Part of #9497.
Depends on #10329


wmontwe (approved):
LGTM

PR #10343, opened 2026-01-10
fix: don't include CSS when copying link text

### LINKED ISSUE #10115: LINK_TEXT_COPY does copy some style items
### Checklist

### App

Thunderbird for Android

### App version

13.0

F-Droid

### Android version

Not relevant - if you can't reproduce on your device - there is no problem

### Device model

Not relevant

### Steps to reproduce

1. Open E-Mail with Hyperlink inside
per example:

[test_copy_link_text.html](https://github.com/user-attachments/files/23592009/test_copy_link_text.html)

3. Press long on the Hyperlink
4. select "Copy link text to clipboard"
5. Paste in Browser-Address-Input or Text-Editor - anything without format
6. Do you see the problem?

### Expected behavior

Get the text `this_is_the_problem_text` which I see in the E-Mail

### Actual behavior

I get this:
```
.linkStyledText, .linkStyledText span, span.linkStyledText { color: #0968F6!important; }
@media(prefers-color-scheme: dark) {
.linkStyledText, .linkStyledText span, span.linkStyledText { color: #4D93FC!important; }
}
this_is_the_problem_text
```

### More

I think you have to change something in the file
`legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/LinkTextHandler.kt`

farukgdev:
Fixes #10115 

### Problem
If malformed HTML nests a `<style>` tag inside an `<a>`, long-press the link and choose "Copy link text to clipboard". This copies CSS from the `<style>` tag along with the visible link text.

Minimal example:
```html
<a href="https://example.com/">
  <style>.x{color:#00f}</style>
  <span>Link text</span>
</a>
```

### Changes
1) HTML sanitization
- In `BodyCleaner`, move `<style>` elements found inside `<a>` out of the anchor and insert them immediately before the `<a>` (CSS is preserved).
- This is done during sanitization so the WebView receives a normalized DOM and the UI doesn't need additional parsing/heuristics at long-press time.
- The sanitizer's allowlist already strips elements like `<script>`; `<style>` is explicitly allowed, so handling `<style>` inside `<a>` is sufficient for this bug.
- This is a no-op for valid HTML where `<style>` isn't nested inside `<a>`.

2) Trimming title
- `requestFocusNodeHref()` can return leading/trailing whitespace in the `title` for this kind of malformed markup. Trim before copying so the clipboard content matches what users see.

### Testing
- Added unit test: `should move style tags out of anchors`

Manual test:
1. Send an HTML email containing the minimal example above (as message body).
2. Open the email in the Android app, long-press the link, choose "Copy link text to clipboard".
3. Paste into a plain text field. Only `Link text` is pasted (no CSS, no leading/trailing whitespace).

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/LinkTextHandler.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/LinkTextHandler.kt
@@ -12,8 +12,8 @@ internal class LinkTextHandler(
 ) : Handler() {
 
     override fun handleMessage(message: Message) {
-        val bundle = message.data
-        val linkText = bundle.getString("title") ?: return
+        val linkText = message.data.getString("title")?.trim().orEmpty()
+        if (linkText.isEmpty()) return
 
         val label = context.getString(R.string.webview_contextmenu_link_text_clipboard_label)
         clipboardManager.setText(label, linkText)


rafaeltonholo (approved):
LGTM

PR #10354, opened 2026-01-13
refactor: address quality issues

wmontwe:
Part of #10209

This PR introduces improvements to the `legacy:ui` and core modules, focusing on migrating deprecated Android APIs to their modern counterparts, refining the user experience in split-screen and landscape modes, and addressing several long-standing build issues.

#### 1. Modernization of Menu Handling (`MenuHost` / `MenuProvider`)
Migrated all fragments in the `legacy:ui` module from the deprecated `setHasOptionsMenu(true)` and `onCreateOptionsMenu` pattern to the modern `MenuHost` and `MenuProvider` API.
*   **Fragments Updated**: `FolderSettingsFragment`, `ManageFoldersFragment`, `AbstractMessageListFragment`, `MessageViewFragment`, and `AccountSettingsFragment`.
*   **Menu Logic Optimization**: Systematically split the shared, massive message list menu into context-specific resources:
    *   `message_list_option_menu.xml`: Strictly list actions (Search, Sort, etc.).
    *   `message_view_option_menu.xml`: Specifically for individual message actions.
*   **Visibility Control**: Refined logic to ensure that only relevant menu items are shown in various display modes (Single Pane, Split View, and Landscape), preventing menu clutter when transitioning between list and message views.

#### 2. Migrated Back Navigation to `OnBackPressedDispatcher`
Replaced the deprecated `onBackPressed()` overrides in primary activities with the modern `OnBackPressedCallback` system.
*   **`MainActivity`**: Refactored complex back-stack logic to correctly handle drawer closing, search collapsing, and app exit. Improved delegation to the system back handler when app-specific conditions are not met to prevent navigation loops.
*   **`MessageCompose`**: Migrated draft saving/discarding prompts to the new callback system, ensuring consistent lifecycle behavior.

#### 3. Search UI Enhancements
*   **State Preservation**: Added explicit state saving for the `SearchView` in `AbstractMessageListFragment`. Search queries and expanded/collapsed states are now correctly preserved across screen rotations.
*   **Delegation**: Updated `onSearchRequested()` in `MainActivity` to properly delegate search expansion to the active fragment, ensuring hardware search buttons and system requests trigger the correct UI.
*   **Contextual Visibility**: Fixed issues where search was incorrectly visible or hidden in split-screen landscape modes, ensuring a clean toolbar focused on the user's current task.

#### 4. Build Cleanup and Simple Fixes
*   **Typo Correction**: Renamed `rootCauseMassage` to `rootCauseMessage` in `ThrowableExtensions.kt` and updated all call sites.
*   **Deprecation Cleanup**: Replaced deprecated `ExceptionHelper.getRootCauseMessage()` calls with the modern `ThrowableExtensions` property across `ImapSync`, `Pop3Sync`, `SendFailedNotificationController`, and `MessagingController`.

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MainActivity.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MainActivity.kt
@@ -41,8 +41,8 @@ import com.fsck.k9.ui.BuildConfig
 import com.fsck.k9.ui.R
 import com.fsck.k9.ui.base.BaseActivity
 import com.fsck.k9.ui.managefolders.ManageFoldersActivity
-import com.fsck.k9.ui.messagelist.AbstractMessageListFragment
-import com.fsck.k9.ui.messagelist.AbstractMessageListFragment.MessageListFragmentListener
+import com.fsck.k9.ui.messagelist.BaseMessageListFragment
+import com.fsck.k9.ui.messagelist.BaseMessageListFragment.MessageListFragmentListener
 import com.fsck.k9.ui.messagelist.DefaultFolderProvider
 import com.fsck.k9.ui.messageview.MessageViewContainerFragment
 import com.fsck.k9.ui.messageview.MessageViewContainerFragment.MessageViewContainerListener
@@ -84,7 +84,7 @@ private const val TAG = "MainActivity"
  * "View Message" notification.
  *
  * `MainActivity` manages the overall layout, including the navigation drawer and the main content area,
- * which currently displays either a [AbstractMessageListFragment] or a [MessageViewContainerFragment]. It orchestrates
+ * which currently displays either a [BaseMessageListFragment] or a [MessageViewContainerFragment]. It orchestrates
  * the interactions between these fragments and handles the back stack. The responsibilities for managing the
  * action bar, search functionality, and single-pane/split-view layout logic are currently handled here but
  * are intended to be refactored into more dedicated components over time.
@@ -116,8 +116,8 @@ open class MainActivity :
     private var openFolderTransaction: FragmentTransaction? = null
     private var progressBar: ProgressBar? = null
     private var messageViewPlaceHolder: PlaceholderFragment? = null
-    private val messageListFragmentFactory: AbstractMessageListFragment.Factory by inject()
-    private var messageListFragment: AbstractMessageListFragment? = null
+    private val messageListFragmentFactory: BaseMessageListFragment.Factory by inject()
+    private var messageListFragment: BaseMessageListFragment? = null
     private var messageViewContainerFragment: MessageViewContainerFragment? = null
     private var account: LegacyAccountDto? = null
     private var search: LocalMessageSearch? = null
@@ -271,7 +271,7 @@ open class MainActivity :
         val fragmentManager = supportFragmentManager
         messageListFragment = fragmentManager.findFragmentById(
             R.id.message_list_container,
-        ) as? AbstractMessageListFragment
+        ) as? BaseMessageListFragment
         messageViewContainerFragment =
             fragmentManager.findFragmentByTag(FRAGMENT_TAG_MESSAGE_VIEW_CONTAINER) as? MessageViewContainerFragment
 
@@ -1121,7 +1121,7 @@ open class MainActivity :
         }
     }
 
-    private fun addMessageListFragment(fragment: AbstractMessageListFragment) {
+    private fun addMessageListFragment(fragment: BaseMessageListFragment) {
         messageListFragment?.isActive = false
 
         supportFragmentManager.commit {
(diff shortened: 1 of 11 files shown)


legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MainActivity.kt
```diff
@@ -781,26 +778,30 @@ open class MainActivity :
...
-                    val defaultFolderId = defaultFolderProvider.getDefaultFolder(account!!)
-                    val currentFolder = if (singleFolderMode) search!!.folderIds[0] else null
-                    if (currentFolder == null || defaultFolderId != currentFolder) {
-                        openFolderImmediately(defaultFolderId)
-                    } else {
-                        super.onBackPressed()
-                    }
+                    callback.isEnabled = false
+                    onBackPressedDispatcher.onBackPressed()
+                    callback.isEnabled = true
```
rafaeltonholo: This is repeating at least 3 times. Could we extract to a method?

core/common/src/commonMain/kotlin/net/thunderbird/core/common/collections/Queue.kt
```diff
@@ -49,7 +49,14 @@ interface Queue<T> : Collection<T> {
...
 
-abstract class AbstractQueue<T> : AbstractMutableCollection<T>(), Queue<T> {
+/**
+ * Base class for [Queue] implementations.
+ *
+ * Provides default implementations for some of the [Queue] and [MutableCollection] operations.
+ *
+ * @param T The type of elements held in this collection.
+ */
+abstract class BaseQueue<T> : AbstractMutableCollection<T>(), Queue<T> {
```
rafaeltonholo, line 59: Is this renaming change necessary?

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/BaseMessageListFragment.kt
```diff
@@ -1086,37 +1157,29 @@ abstract class AbstractMessageListFragment :
             menu.findItem(R.id.expunge).isVisible = false
         }
 
-        menu.findItem(R.id.search).isVisible = !isManualSearch
```
rafaeltonholo, line 1089: I want to confirm whether it was intentional to remove this line. If it is removed, users will see the search icon during the "Search Results", which differs from the current behaviour.

Keeping this menu enabled after performing a search allows users to enter another query. However, it also creates a navigation stack, which means users will have to press the back button multiple times to return to the message list.

rafaeltonholo (dismissed):
LGTM, I've just left a few comments, but nothing major!

Thanks for this cleanup!!

legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MainActivity.kt
wmontwe: This logic is still not easy to read. I'll extract the repeated calls. And in a later PR rewrite the whole handling.

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/BaseMessageListFragment.kt
wmontwe: I tried to match the existing behavior and this seemed not necessary. How could I replicate it?

core/common/src/commonMain/kotlin/net/thunderbird/core/common/collections/Queue.kt
wmontwe: Yes. We use `BaseXX` to denote abstract classes. Having both `BaseXX` and `AbstractXX` introduces unnecessary confusion. Using `AbstractXX` would also lead to declarations like `abstract class AbstractXX`, which is redundant and doesn’t read well. Additionally, the `AbstractXX` naming convention is primarily associated with older Java codebases and is not idiomatic in modern Java/Kotlin projects.

rafaeltonholo: Could you please rename `AbstractMessageListFragment` to `BaseMessageListFragment` too?

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/BaseMessageListFragment.kt
rafaeltonholo: To replicate the current behaviour, just revert that line deletion. It will hide the search button when you enter the Search view.

core/common/src/commonMain/kotlin/net/thunderbird/core/common/collections/Queue.kt
wmontwe: Sure

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/BaseMessageListFragment.kt
wmontwe: Found and fixed it.

PR #10388, opened 2026-01-20
feat: add new main activity

### LINKED ISSUE #10387: Add new MainActivity
The current `MainActivity` is very large. It needs to be split into message list related code and a general purpose startup logic activity. 

Therefore this should:

- Rename `MainActivity` to `MessageListActivity`
- Add new `MainActivity`
- Move startup logic into `MainActivity`

wmontwe:
Resolves #10387

This pull request introduces a new `MainActivity` for the `app-common` module to handle app startup and database upgrades, and refactors the launcher flow accordingly. It also removes references to the legacy `MainActivity` throughout the codebase, replacing them with `MessageListActivity` to streamline navigation and clarify responsibilities.

Renames:
- `MainActivity` -> `MessageListActivity`
- `Search` -> `MessageSearchActivity`
- `UpgradeDatabases` - > `UpgradeDatabaseActivity`

Additionally converted from Java to Kotlin:
- `Search`
- `UpgradeDatabases`

Also fixed the failing `LoadPreferencesSideEffectTest`

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageHomeActivity.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageHomeActivity.kt
@@ -90,7 +90,7 @@ private const val TAG = "MainActivity"
  * are intended to be refactored into more dedicated components over time.
  */
 @Suppress("TooManyFunctions", "LargeClass")
-open class MessageListActivity :
+open class MessageHomeActivity :
     BaseActivity(),
     MessageListFragmentListener,
     MessageViewFragmentListener,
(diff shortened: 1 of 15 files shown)


legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageHomeActivity.kt
rafaeltonholo, line 1: I recommend using a different name instead of `MessageListActivity`. Keeping this name can lead to the same confusion that prompted the original decision to rename it to `MainActivity`. Since this activity encompasses features from both the Message List and Message Reader and functions as the home screen, I suggest renaming it to `HomeScreenActivity`.

app-common/src/main/AndroidManifest.xml
```diff
@@ -9,7 +9,29 @@
...
+            >
+
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.LAUNCHER" />
+                <category android:name="android.intent.category.APP_EMAIL" />
+
+                <!-- TODO: Remove once minSdkVersion has been changed to 24+ -->
```
rafaeltonholo, line 27: nit: It would be nice having a ticket to track this

legacy/ui/legacy/src/main/java/com/fsck/k9/activity/UpgradeDatabaseActivity.kt
rafaeltonholo, line 1: I suggest avoiding using the bang operator and properly handling the null cases. Bang operators are a common root cause for null pointers.

legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageHomeActivity.kt
wmontwe: What about `MessageHomeActivity`?

PR #10406, opened 2026-01-21
feat(message-list): Date/Time Format Option for Message List

dani-zilla:
Adds functionality described in #[10202](https://github.com/thunderbird/thunderbird-android/issues/10202)

Adds date time format options for message list, including contextual and full. 

Full: Displays the entire date and time, localized to user's phone and options. 
Contextual: the default, follows these rules depending on when the email came in:
- Today: show time (ex: 2:34 PM or 14:34)
- Yesterday: "Yesterday"
- Between yesterday and 1 week ago, exclusive: Day of Week
- Greater than 7 days ago: Month/Day (ex: "Jan 1st")
- Greater than 1 year old: Full Date (ex: "1/23/24, 1:23 PM")

Option currently works on both XML view and upcoming Jetpack Compose views, however the compose views (bottom two screenshots below) require refactoring of the message list item to keep more consistent column widths (out of scope for this change). 

dateFormatOptions

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/BaseMessageListFragment.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/BaseMessageListFragment.kt
@@ -319,12 +319,6 @@ abstract class BaseMessageListFragment :
             return
         }
 
-        viewModel.getMessageListLiveData().observe(this) { messageListInfo: MessageListInfo ->
-            setMessageList(messageListInfo)
-        }
-
-        adapter = createMessageListAdapter()
-
         generalSettingsManager.getSettingsFlow()
             /**
              * Skips the first emitted item from the settings flow,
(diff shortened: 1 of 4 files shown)


dani-zilla:
Seems the latest changes from main causes a crash in `BaseMessageListFragment::createMessageListAdapter()` when the user changes a view setting _other_ than the date/time format or other changes to the message list items. Investigating how these changes aren't working together.

rafaeltonholo:
> Seems the latest changes from main causes a crash in `BaseMessageListFragment::createMessageListAdapter()` when the user changes a view setting _other_ than the date/time format or other changes to the message list items. Investigating how these changes aren't working together.

The crash is unrelated to your changes. Turn off the `enable_message_list_new_state` feature flag, and test your changes again. If the app doesn't crash, you are good to go.

dani-zilla:
> > Seems the latest changes from main causes a crash in `BaseMessageListFragment::createMessageListAdapter()` when the user changes a view setting _other_ than the date/time format or other changes to the message list items. Investigating how these changes aren't working together.
> 
> The crash is unrelated to your changes. Turn off the `enable_message_list_new_state` feature flag, and test your changes again. If the app doesn't crash, you are good to go.

I'll test with that as well, though I was able to fix the crash. Separately, our changes were fine, but when combined, it was refreshing the view twice, and once did so without initializing the message list appearance lateinit variable. I think this has it working as expected, but I do also want to test it with the new message list state

dani-zilla:
Can confirm the fix works in both the new message state and old state. Date format updates on change as expected either way, and other changes to display preferences do not cause a crash. It all seems to work as expected in a quick smoke test.

Issue #10424, opened 2026-01-27
Bump AGP to 9.1


wmontwe:
Update the AGP version to 9.0, now 9.1 (once IntelliJ IDEA is compatible).

You need to:
-  remove the `id("org.jetbrains.kotlin.android")` declaration from `thunderbird.library.android.gradle.kts` and `thunderbird.app.android.gradle.kts` for full compatibility.
- replace all `proguard-android.txt` by `proguard-android-optimize.txt` in `app-k9mail` and `app-thunderbird` `build.gradle.kts` files.

wmontwe:
We're going to ignore IntelliJ IDEA compatibility as this blocks other important dependency updates like Kotlin 2.3.0.

New AGP target version is 9.1.0

Issue #10423, opened 2026-01-27
Refactor build-plugin for AGP 9.0 compatibility


wmontwe:
The badging and version info scripts are not compatible. Also some of the android app and library definitions use removed attributes and AGP APIs that don't exist anymore.

PR #10425, opened 2026-01-27
refactor: build plugin for AGP 9 compatibility

wmontwe:
Depends on: #10438 

Resolvess #10423

This pull request introduces a new build logic plugin module and adds new Gradle plugins for app badging and versioning, and improves Gradle build configuration for AGP 9.0 compatibility.

The remaining work is documented in #10424 

Added a flag `-Pci=true` to compensate for the now mandatory optimization. It is used for the PR runs to reduce build times.

### DIFF
--- a/build-plugin/src/main/kotlin/thunderbird.library.android.compose.gradle.kts
+++ b/build-plugin/src/main/kotlin/thunderbird.library.android.compose.gradle.kts
@@ -9,14 +9,9 @@ plugins {
     id("thunderbird.quality.spotless")
 }
 
-android {
-    buildFeatures {
-        compose = true
-    }
-}
-
 androidComponents {
     beforeVariants(selector().withBuildType("release")) { variantBuilder ->
+        @Suppress("UnstableApiUsage")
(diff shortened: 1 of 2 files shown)


build-plugin/src/main/kotlin/thunderbird.library.android.compose.gradle.kts
```diff
@@ -10,7 +10,9 @@ plugins {
 }
 
 android {
-    configureSharedComposeConfig(libs)
+    buildFeatures {
+        compose = true
```
Goooler: This flas in unnecessary, the Compose plugin enables it automatically.

wmontwe: Yes, this is now not needed anymore. I'll remove that bit. Thanks!

wmontwe:
For performance reasons I removed the switch to optimized ProGuard rules in our build configs. This is now part of #10424

wmontwe:
@dani-zilla @coreycb I updated the branch and it's ready for review.

PR #10444, opened 2026-01-28
refactor(message-list): integrate sort criteria with new message list

### LINKED ISSUE #10390: Message List Refactor - Fetch and Update sort type using new Message List State
This task should take care of:
1. Override the logic to fetch the Sort Type from the Legacy MessageListFragment(BaseMessageListFragment) and delegate to the MessageListViewModel and its State Machine
2. Enable the new MessageListFragment to update the sort type via the new ViewModel

rafaeltonholo:
Resolves #10390.

## Description
- Introduce `SortCriteria` type, which correctly handles the primary and secondary sort type clauses
- Introduce a new Sort Type menu to the message list, improving the user's awareness of what sort type is currently selected. (I'll ask for a design review for this, and layout changes will be applied in a new PR)
- Introduce a `ClickableSurface` atom.
- Enforce trailing commas in the project settings.

## Know issue
When the message list loads for the first time, we are not yet using the new state to load the message items. As a result, the SortType and the MessageList loading process are not synchronized. This leads to the app performing a sort operation after the items have loaded, which displays an animation as the items are sorted. This issue should be resolved once we start loading the message items using the new `MessageListState`.

### DIFF
--- a/feature/mail/message/list/api/src/main/kotlin/net/thunderbird/feature/mail/message/list/domain/model/SortCriteria.kt
+++ b/feature/mail/message/list/api/src/main/kotlin/net/thunderbird/feature/mail/message/list/domain/model/SortCriteria.kt
@@ -41,6 +41,8 @@ data class SortCriteria(
         }
     }
 
+    val isSecondarySortRequired: Boolean = primary !in SecondaryNotRequiredForSortTypes
+
     companion object {
         val DateSortTypeOnly = setOf(SortType.DateAsc, SortType.DateDesc)
 
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/MessageListFragment.kt
@@ -11,20 +11,22 @@ import android.widget.PopupWindow
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.Crossfade
-import androidx.compose.animation.expandHorizontally
+import androidx.compose.animation.expandVertically
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
-import androidx.compose.animation.shrinkHorizontally
+import androidx.compose.animation.shrinkVertically
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.IntrinsicSize
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
-import androidx.compose.foundation.layout.wrapContentWidth
+import androidx.compose.foundation.layout.widthIn
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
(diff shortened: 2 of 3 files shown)


feature/mail/message/list/api/src/main/kotlin/net/thunderbird/feature/mail/message/list/domain/model/SortCriteria.kt
```diff
@@ -0,0 +1,53 @@
...
+@Stable
+data class SortCriteria(
+    val primary: SortType,
+    val secondary: SortType? = null,
+) {
+    init {
+        // using require here would imply to either not have a proper error message or a wrong one.
+        @Suppress("UseRequire")
+        if (primary in SecondaryNotRequiredForSortTypes && secondary != null) {
+            throw IllegalArgumentException("Secondary sorting criterion must be null for $primary")
```
dani-zilla, line 32: This appears to crash on launch for me:
`java.lang.IllegalArgumentException: Secondary sorting criterion must be null for ArrivalDesc`

rafaeltonholo: Good catch! I've missed `SORT_ARRIVAL` when loading the default sort types. Now it should be fixed.

wmontwe (approved):
LGTM

dani-zilla:
Screenshot 2026-02-04 at 2 20 42 PM
I noticed the UI seems a bit off here. Going to landscape doesn't give any more space, but I have those two items on the right of the menu that go off the screen. 

It seems to be only for options that have a secondary sort, like the attachments selected in this screenshot and the unread/read ones as well.

rafaeltonholo:
> Screenshot 2026-02-04 at 2 20 42 PM I noticed the UI seems a bit off here. Going to landscape doesn't give any more space, but I have those two items on the right of the menu that go off the screen.
> 
> It seems to be only for options that have a secondary sort, like the attachments selected in this screenshot and the unread/read ones as well.

Yeah, the second part only shows up when the sort criteria requires a secondary sort type.

Ouch, that is worse than I thought. Let's wait until the design review to get a better design solution to this, though

dani-zilla:
So the two-column sort window display issue had me thinking about a cleaner way to do that. Because the sort could end up taking up the whole screen or more depending on the font size, display size, and other factors. But what if we could display it in a single column. I had two ideas for this. 

So, first, I noticed that in `SortCriteria.kt` that the secondary option is _always_ date, new/old. We could always display the secondary options, especially since date new/old is the default sort anyway. So perhaps we could have it display at the top of the sort window no matter what, with a divider before the options that could set it as a secondary sort option, with the user only able to select one item from either side of the divider. Like this:

```
Date (oldest first) 
Date (newest first) ☑
----------------------
Arrival (oldest first)
Arrival (newest first)
...
...
Unread first ☑
```

The other idea I had was conditionally displaying date new/old under the items where that could become a secondary sort. This would make the sort view look pretty much as it does now, but, instead of adding a second column next to the first one when date can be a secondary sort option, it adds it under the item, like so:

Unselected state:
```
Starred first
Unstarred first
```
Selected state:
```
Starred first ☑
    Date (oldest first) 
    Date (newest first) ☑
Unstarred first
```

I do lean towards the first option, since it’s always a primary or secondary option, and then we don’t have to dynamically change the list for any reason other than showing what’s selected. 

If the write-up and text descriptions aren’t clear, I could always whip up a mock-up, but I do agree, at this point, we may want to wait to see what design says

dani-zilla:
> So the two-column sort window display issue had me thinking about a cleaner way to do that. Because the sort could end up taking up the whole screen or more depending on the font size, display size, and other factors. But what if we could display it in a single column. I had two ideas for this.
> 
> So, first, I noticed that in `SortCriteria.kt` that the secondary option is _always_ date, new/old. We could always display the secondary options, especially since date new/old is the default sort anyway. So perhaps we could have it display at the top of the sort window no matter what, with a divider before the options that could set it as a secondary sort option, with the user only able to select one item from either side of the divider. Like this:
> 
> ```
> Date (oldest first) 
> Date (newest first) ☑
> ----------------------
> Arrival (oldest first)
> Arrival (newest first)
> ...
> ...
> Unread first ☑
> ```
> 
I just thought, too, since one of the secondary options is always selected when it's sort by date, we could go with radio buttons here to further distinguish them

rafaeltonholo:
> > ```
> > Date (oldest first) 
> > Date (newest first) ☑
> > ----------------------
> > Arrival (oldest first)
> > Arrival (newest first)
> > ...
> > ...
> > Unread first ☑
> > ```
> 

My only concern about this approach is that it could give the impression that the Messages will be first ordered by date, then by unread, which "breaks" the sort users want and could cause confusion.

I would prefer the second option because it explicitly tells the user that it will be first sorted by "unread first" and then sorted by date, which is the correct sort criteria we are going to use.

We might need to use another component, maybe a bottom sheet, to display the options, as a pop-up here proved to be a bad choice.

We will have the design review today. Let's bring all these options to the table and see what they think about it.

dani-zilla:
> I would prefer the second option because it explicitly tells the user that it will be first sorted by "unread first" and then sorted by date, which is the correct sort criteria we are going to use. 

That is a good point, the first option does work better if it was something that was sorting within a filter, less for a secondary sort. The second option better communicates that it's a sort within a sort, I think you're right, it's the better way to go

rafaeltonholo:
> > I would prefer the second option because it explicitly tells the user that it will be first sorted by "unread first" and then sorted by date, which is the correct sort criteria we are going to use.
> 
> That is a good point, the first option does work better if it was something that was sorting within a filter, less for a secondary sort. The second option better communicates that it's a sort within a sort, I think you're right, it's the better way to go

I've implemented the second option for now. Once we have the design review next week, we can revisit the layout again in another PR.

PR #10452, opened 2026-01-30
chore(ci): tweak merge script to have less conflicts and output to be more readable

dandarnell:
This patch does two things:

Tells git to default to a "theirs" strategy when there are merge conflicts (i.e. in a main -> beta merge conflict, take the file from main).

Highlight "CONFLICT" in red, so it can be easily spotted in `git merge` output.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,4 +6,11 @@
 
 app-k9mail/build.gradle.kts merge=merge_gradle
 app-thunderbird/build.gradle.kts merge=merge_gradle
+
+app-thunderbird/src/release/res/raw/changelog_master.xml merge=ours
+app-thunderbird/src/beta/res/raw/changelog_master.xml merge=ours
 app-k9mail/src/main/res/raw/changelog_master.xml merge=ours
+
+app-metadata/com.fsck.k9/*/changelogs/** merge=ours
+app-metadata/net.thunderbird.android.beta/*/changelogs/** merge=ours
+app-metadata/net.thunderbird.android/*/changelogs/** merge=ours
--- a/scripts/ci/merges/do_merge.sh
+++ b/scripts/ci/merges/do_merge.sh
@@ -43,8 +43,9 @@ git checkout ${into_branch}
 git pull
 git config merge.ours.driver true
 git config merge.merge_gradle.driver "python3 scripts/ci/merges/merge_gradle.py %A %B"
+git config merge.merge_changelog.driver "scripts/ci/merges/merge_changelog.sh %A %O %B"
 set +e
-git merge "origin/${from_branch}"
+git merge -Xtheirs "origin/${from_branch}" 2>&1 | grep --color=always -E '\bCONFLICT\b|$'
 ret=$?
 set +x
 
--- a/scripts/ci/merges/merge_changelog.sh
+++ b/scripts/ci/merges/merge_changelog.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+# This merge driver exists to override the -Xtheirs CLI option used when merging
+# two trees together (e.g. main -> beta).
+
+A="$1"  # File A (ours)
+O="$2"  # Common ancestor
+B="$3"  # File B (theirs)
+
+git merge-file -- "$A" "$O" "$B"


scripts/ci/merges/do_merge.sh
```diff
@@ -44,7 +44,7 @@ git pull
 git config merge.ours.driver true
 git config merge.merge_gradle.driver "python3 scripts/ci/merges/merge_gradle.py %A %B"
 set +e
-git merge "origin/${from_branch}"
+git merge -Xtheirs "origin/${from_branch}" 2>&1 | grep --color=always -E '\bCONFLICT\b|$'
```
coreycb, line 48: Instead of always taking theirs on a conflict, I think we may need to be more surgical, at least for some files.

There are currently some merge drivers in use, maybe we add to them instead?
https://github.com/thunderbird/thunderbird-android/commit/64965cff7dd8dc7d4fd6e4cdb3a93cf57fd17e26

Minimally it looks like we might be able to do:
path/to/dir/**        merge=theirs  # take theirs if conflicts
path/to/dir/keep/**   merge=text  # should fail if conflicts? standard behavior

coreycb (changes requested):

.gitattributes
```diff
@@ -6,4 +6,11 @@
...
 app-thunderbird/build.gradle.kts merge=merge_gradle
-app-k9mail/src/main/res/raw/changelog_master.xml merge=ours
+
+app-k9mail/src/main/res/raw/changelog_master.xml merge=merge_changelog
+app-thunderbird/src/debug/res/raw/changelog_master.xml merge=merge_changelog
+app-thunderbird/src/daily/res/raw/changelog_master.xml merge=merge_changelog
+scripts/ci/templates/changelog_master.xml merge=merge_changelog
+
+app-metadata/com.fsck.k9/*/changelogs/** merge=merge_changelog
+app-metadata/net.thunderbird.android.beta/*/changelogs/** merge=merge_changelog
```
coreycb: You'll want to add this one too:
`app-metadata/net.thunderbird.android/*/changelogs/** merge=merge_changelog`

.gitattributes
```diff
@@ -6,4 +6,11 @@
 
 app-k9mail/build.gradle.kts merge=merge_gradle
 app-thunderbird/build.gradle.kts merge=merge_gradle
-app-k9mail/src/main/res/raw/changelog_master.xml merge=ours
+
+app-k9mail/src/main/res/raw/changelog_master.xml merge=merge_changelog
```
coreycb: I think we'll want to add this one too:
`./app-thunderbird/src/release/res/raw/changelog_master.xml`

It only exists in the release branch but to be safe we don't want to overwrite it.

.gitattributes
```diff
@@ -6,4 +6,11 @@
...
+app-thunderbird/src/debug/res/raw/changelog_master.xml merge=merge_changelog
+app-thunderbird/src/daily/res/raw/changelog_master.xml merge=merge_changelog
+scripts/ci/templates/changelog_master.xml merge=merge_changelog
```
coreycb: These 3 can can be removed. It should be fine to overwrite what's in beta/release with what is in main:
```
app-thunderbird/src/debug/res/raw/changelog_master.xml merge=merge_changelog
app-thunderbird/src/daily/res/raw/changelog_master.xml merge=merge_changelog
scripts/ci/templates/changelog_master.xml merge=merge_changelog
```

.gitattributes
```diff
@@ -6,4 +6,10 @@
...
 app-k9mail/build.gradle.kts merge=merge_gradle
 app-thunderbird/build.gradle.kts merge=merge_gradle
-app-k9mail/src/main/res/raw/changelog_master.xml merge=ours
+
+app-k9mail/src/main/res/raw/changelog_master.xml merge=merge_changelog
+app-thunderbird/src/release/res/raw/changelog_master.xml
+
+app-metadata/com.fsck.k9/*/changelogs/** merge=merge_changelog
+app-metadata/net.thunderbird.android.beta/*/changelogs/** merge=merge_changelog
+app-metadata/net.thunderbird.android/*/changelogs/** merge=merge_changelog
```
coreycb: Again, sorry I missed this. We don't want to merge anything in the above 3 changelogs directories. These directories contain release notes for releases from the current branch. (They are supposed to, at least, but that's probably not the case anymore due to previous merges.) The file names are based on the versionCode, and there's potential that the versionCode could overlap between beta and release.

So what we really want to do here is exclude any changes to these directories in the merge. Or at least use merge=ours.

.gitattributes
```diff
@@ -6,4 +6,10 @@
 
 app-k9mail/build.gradle.kts merge=merge_gradle
 app-thunderbird/build.gradle.kts merge=merge_gradle
-app-k9mail/src/main/res/raw/changelog_master.xml merge=ours
+
+app-k9mail/src/main/res/raw/changelog_master.xml merge=merge_changelog
+app-thunderbird/src/release/res/raw/changelog_master.xml
```
coreycb: I'm sorry, I missed something here and I think I sent you down the wrong path. For the following 3 files, we want to keep ours in all cases.

app-thunderbird/src/release/res/raw/changelog_master.xml  merge=ours
app-thunderbird/src/beta/res/raw/changelog_master.xml merge=ours
app-k9mail/src/main/res/raw/changelog_master.xml   merge=ours

coreycb (changes requested):

.gitattributes
dandarnell: So would this mean that the custom merge_changelog merge driver is not necessary after all?

PR #10454, opened 2026-01-30
chore(docs) update to merge documentation

coreycb:
The milestone guidance was incorrect and feature flags don't need to be checked during merges. They need
to be checked during code reviews and uplifts.


stephenking314 (approved):
LGTM

PR #10473, opened 2026-02-04
refactor: convert Address and AddressTest class from java to kotlin

### LINKED ISSUE #9276: Convert Address class from Kotlin to Java
We should migrate the `com.fsck.k9.mail.Address` class to Kotlin to align with the ongoing modernization efforts and Kotlin-first development approach across the codebase.

mohitsatr:
Resolves #9276 

Changes:
1. Convert both classes from Java to Kotlin.
2. Make static methods `toString()`, `pack()`, to return an empty string (instead of null) when addresses are not provided to avoid nullable type (String -> String?) conflicts in other places.
3. Add tests for the above.
4. Refactor `AddressTest` to use Assertk and make tests follow the AAA pattern

### DIFF
--- a/mail/common/src/test/java/com/fsck/k9/mail/AddressTest.kt
+++ b/mail/common/src/test/java/com/fsck/k9/mail/AddressTest.kt
@@ -1,24 +1,24 @@
-package com.fsck.k9.mail;
+package com.fsck.k9.mail
 
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import assertk.assertions.isFalse
+import assertk.assertions.isNotEqualTo
+import assertk.assertions.isNull
+import assertk.assertions.isTrue
+import com.fsck.k9.mail.Address.Companion.parse
+import com.fsck.k9.mail.Address.Companion.quoteString
+import net.thunderbird.core.logging.legacy.Log
+import net.thunderbird.core.logging.testing.TestLogger
+import org.junit.Before
+import org.junit.Test
 
-import net.thunderbird.core.logging.legacy.Log;
-import net.thunderbird.core.logging.testing.TestLogger;
-import org.junit.Before;
-import org.junit.Test;
-
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-
-
-public class AddressTest {
+class AddressTest {
 
     @Before
-    public void setUp() {
-        Log.logger = new TestLogger();
+    fun setUp() {
+        Log.logger = TestLogger()
     }
-
 
     /**
      * test the possibility to parse "From:" fields with no email.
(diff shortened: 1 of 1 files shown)


mail/common/src/test/java/com/fsck/k9/mail/AddressTest.java
```diff
@@ -1,164 +0,0 @@
...
-    }
-
-
-    /**
-     * test the possibility to parse "From:" fields with no email.
-     * for example: From: News for Vector Limited - Google Finance
-     * http://code.google.com/p/k9mail/issues/detail?id=3814
-     */
-    @Test
-    public void parse_withMissingEmail__shouldSetPersonal() {
```
mohitsatr, line 29: Is this name correct? Shouldn't it be `shouldNotSetPersonal` ?

mail/common/src/test/java/com/fsck/k9/mail/AddressTest.kt
```diff
@@ -0,0 +1,277 @@
...
+
+        // Act
+        val withoutAddress = Address(nameOnly)
+        val withAddress = Address(address = address, personal = nameOnly)
+        val hostname = withAddress.hostname
+
+        // Assert
+        assertThat(withoutAddress.hostname).isEqualTo("")
+        assertThat(withAddress.hostname).isEqualTo(domain)
+    }
```
rafaeltonholo, line 56: The hostname should return null instead of an empty string.
```suggestion
    @Test
    @Throws(Exception::class)
    fun `hostname when address without is provided without domain should return null`() {
        // Arrange
        val nameOnly = "Max"
        val domain = "domain.com"
        val address = "$nameOnly@$domain"

        // Act
        val withoutAddress = Address(nameOnly)
        val withAddress = Address(address = address, personal = nameOnly)

        // Assert
        assertThat(withoutAddress.hostname).isNull()
        assertThat(withAddress.hostname).isEqualTo(domain)
    }
```

mail/common/src/test/java/com/fsck/k9/mail/AddressTest.kt
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     * test the possibility to parse "From:" fields with no email.
+     * for example: From: News for Vector Limited - Google Finance
+     * http://code.google.com/p/k9mail/issues/detail?id=3814
+     */
+    @Test
+    @Throws(Exception::class)
+    fun `parse with Missing Email should not set personal`() {
```
rafaeltonholo:
```suggestion
    fun `parse with Missing Email should return empty addresses`() {
```

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,307 @@
...
+
+    constructor(address: Address) : this(address.mAddress!!, address.personal, false)
+
+    constructor(address: String, personal: String?) : this(address, personal, true)
+
+    constructor(address: String) : this(address, null, true)
+
+    @Suppress("MemberNameEqualsClassName")
+    val address: String
+        get() = mAddress!!
```
rafaeltonholo: There is no need to have a private nullable field and all of these constructors. `mAddress` is marked with `@NotNull`, so we can assume it is `String` instead of `String?`.

For the multiple constructors, you should use the `@JvmOverloads` annotation.

```suggestion
class Address @JvmOverloads constructor(
    address: String,
    personal: String? = null,
    parse: Boolean = false,
) : Serializable {
    var address: String = address
        private set
    var personal: String? = personal
        private set

    constructor(address: Address) : this(address.address, address.personal)

    init {
        if (parse) {
            val tokens = Rfc822Tokenizer.tokenize(address)
            if (tokens.isNotEmpty()) {
                val token = tokens[0]
                this.address = requireNotNull(token.address) {
                    "token.getAddress()"
                }
                val name = token.name
                this.personal = if (!name.isNullOrEmpty()) {
                    /*
                     * Don't use the "personal" argument if "address" is of the form:
                     * James Bond <james.bond@mi6.uk>
                     *
                     * See issue 2920
                     */
                    name
                } else {
                    personal?.trim()
                }
            } else {
                Log.e("Invalid address: %s", address)
            }
        }
    }
```

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,307 @@
...
+                return ""
+            }
+
+            val hostIdx = mAddress!!.lastIndexOf("@")
+            if (hostIdx == -1) {
+                return ""
+            }
+
+            return mAddress!!.substring(hostIdx + 1)
+        }
```
rafaeltonholo: Ideally, we should not change the behaviour of a property when converting it from Java to Kotlin. In Java, the `getHostname()` function returns `null` when we can't find the `@` char inside the address.

That means, ideally, this property should be nullable, but I can see why you picked to use Non-null String.

I would suggest using the Nullable string and applying the fix in the `HostNameUtils.kt`.
```suggestion
    val hostname: String?
        get() {
            val hostIdx = address.lastIndexOf("@")
            return if (hostIdx == -1) null else address.substring(hostIdx + 1)
        }
```
The diff for the `HostNameUtils` fix:
```diff
Index: core/common/src/commonMain/kotlin/net/thunderbird/core/common/net/HostNameUtils.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/core/common/src/commonMain/kotlin/net/thunderbird/core/common/net/HostNameUtils.kt b/core/common/src/commonMain/kotlin/net/thunderbird/core/common/net/HostNameUtils.kt
--- a/core/common/src/commonMain/kotlin/net/thunderbird/core/common/net/HostNameUtils.kt	(revision 9d2d6368284c924d981c7ed933bb3ff39693ea1d)
+++ b/core/common/src/commonMain/kotlin/net/thunderbird/core/common/net/HostNameUtils.kt	(date 1770744078000)
@@ -16,7 +16,7 @@
      *
      * @return Unobscured host name if `hostName` is valid.
      */
-    fun isLegalHostNameOrIP(hostName: String): String? {
+    fun isLegalHostNameOrIP(hostName: String?): String? {
         /*
          RFC 1123:
          Whenever a user inputs the identity of an Internet host, it SHOULD
@@ -26,7 +26,7 @@
          looking it up in the Domain Name System.
          */
 
-        return isLegalIPAddress(hostName) ?: isLegalHostName(hostName)
+        return hostName?.let { isLegalIPAddress(it) ?: isLegalHostName(it) }
     }
 
     /**
@@ -34,8 +34,8 @@
      *
      * @return Unobscured canonicalized IPv4 or IPv6 address if it is valid, otherwise `null`.
      */
-    fun isLegalIPAddress(hostName: String): String? {
-        return isLegalIPv4Address(hostName) ?: isLegalIPv6Address(hostName)
+    fun isLegalIPAddress(hostName: String?): String? {
+        return hostName?.let { isLegalIPv4Address(it) ?: isLegalIPv6Address(it) }
     }
 
     /**

```

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,307 @@
...
+    override fun hashCode(): Int {
+        var hash = 0
+        if (mAddress != null) {
+            hash += mAddress.hashCode()
+        }
+        if (this.personal != null) {
+            hash += 3 * personal.hashCode()
+        }
+        return hash
+    }
```
rafaeltonholo, line 74: Use the generated `equals` and `hashCode` instead.
```suggestion
    override fun equals(other: Any?): Boolean {
        if (this === other) return true
        if (javaClass != other?.javaClass) return false

        other as Address

        if (address != other.address) return false
        if (personal != other.personal) return false

        return true
    }

    override fun hashCode(): Int {
        var result = address.hashCode()
        result = 31 * result + (personal?.hashCode() ?: 0)
        return result
    }
```

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,307 @@
...
+        } else {
+            quoteAtoms(this.personal!!) + " <" + mAddress + ">"
+        }
+
+    fun toEncodedString(): String =
+        if (isEmpty(this.personal)) {
+            mAddress!!
+        } else {
+            EncoderUtil.encodeAddressDisplayName(this.personal) + " <" + mAddress + ">"
+        }
```
rafaeltonholo, line 88: You don't need to use `TextUtils` here. Additionally, we should use the string interpolation:

```suggestion
    override fun toString(): String =
        if (personal.isNullOrBlank()) {
            address
        } else {
            quoteAtoms(personal) + " <$address>"
        }

    fun toEncodedString(): String =
        if (personal.isNullOrBlank()) {
            address
        } else {
            EncoderUtil.encodeAddressDisplayName(personal) + " <$address>"
        }
```

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,307 @@
...
+     * address contains any non-ASCII characters, and false if all
+     * characters used are within ASCII.
+     *
+     * Note that this returns false for an address such as "Naïve
+     * Assumption &lt;naive.assumption@example.com&gt;", because both
+     * localpart and domain are all-ASCII. There's an ï there, but
+     * it's not in either localpart or domain.
+     */
+    fun needsUnicode(): Boolean {
+        if (mAddress == null) return false
```
rafaeltonholo: Address is a not null property. Additionally, please do not use the bang operator `!!`. This is one of the most common causes for `NullPointerException`.

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,307 @@
...
+         * @return An array of 0 or more Addresses.
+         */
+        @JvmStatic
+        fun parseUnencoded(addressList: String?): Array<Address> {
+            if (isEmpty(addressList)) return EMPTY_ADDRESS_ARRAY
+
+            val tokens = Rfc822Tokenizer.tokenize(addressList)
+            return tokens
+                .filter { !isEmpty(it.address) }
+                .map { Address(it.address!!, it.name, false) }
```
rafaeltonholo: Please avoid using the bang operator `!!`.
```suggestion
                .filterNot { it.address.isNullOrBlank() }
                .map { Address(requireNotNull(it.address), it.name, false) }
```

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,307 @@
...
+            return try {
+                val parsedList =
+                    DefaultAddressParser.DEFAULT
+                        .parseAddressList(addressList, DecodeMonitor.SILENT)
+                        .flatten()
+
+                Array(parsedList.size) { i ->
+                    val m = parsedList[i]
+                    Address("${m.localPart}@${m.domain}", m.name, false)
+                }
```
rafaeltonholo: can be simplified to:
```suggestion
                DefaultAddressParser.DEFAULT
                    .parseAddressList(addressList, DecodeMonitor.SILENT)
                    .flatten()
                    .map { mailbox ->
                        Address(
                            address = "${mailbox.localPart}@${mailbox.domain}",
                            personal = mailbox.name,
                            parse = false,
                        )
                    }
                    .toTypedArray()
```

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,307 @@
...
+            } catch (pe: MimeException) {
+                e(pe, "MimeException in Address.parse()")
```
rafaeltonholo: Avoid statically importing `Log`. This is a deprecated Logger that we need to remove in the future, but to not increase the scope of this change, please use the following:

```suggestion
                Log.e(pe, "MimeException in Address.parse()")
```

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,307 @@
...
+         * @param text String to quote.
+         * @return Possibly quoted string.
+         */
+        @JvmStatic
+        fun quoteAtoms(text: String): String? =
+            if (ATOM.matcher(text).matches()) {
+                text
+            } else {
+                quoteString(text)
+            }
```
rafaeltonholo: The Java implementation of `quoteAtoms` accepts null as a parameter.

```suggestion
        fun quoteAtoms(text: String?): String? = when {
            text == null -> null
            ATOM.matcher(text).matches() -> text
            else -> quoteString(text)
        }
```

mohitsatr: done

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
mohitsatr: done

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
mohitsatr: done

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
mohitsatr: done

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
mohitsatr: done

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
mohitsatr: done

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
mohitsatr: done

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
mohitsatr: done

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
mohitsatr: done

mail/common/src/test/java/com/fsck/k9/mail/AddressTest.kt
mohitsatr: done

mail/common/src/test/java/com/fsck/k9/mail/AddressTest.kt
mohitsatr: done

mohitsatr:
@rafaeltonholo Thanks for the detailed review. I'm new to Kotlin, and the suggested fixes really forced me to go back and review the basics.

Just a few questions:
1. Are we okay with `toString()` and `pack()` static methods returning empty string instead of null, as it was in the original Java file? Project builds successfully.
2. I think we can transfer [Address_quotes](https://github.com/thunderbird/thunderbird-android/blob/fdb8655f3a50f5bddad29399c306acf33180a1a0/mail/common/src/test/java/com/fsck/k9/mail/Address_quoteAtoms.java) tests in `AddressTest` as Address_quotes tests are testing `quoteAtoms()` of Address class, wdyt?

Goooler:
Oops, I missed this! Addressing a convert to #10599.

rafaeltonholo:
> @rafaeltonholo Thanks for the detailed review. I'm new to Kotlin, and the suggested fixes really forced me to go back and review the basics.
> 
> Just a few questions:
> 
>     1. Are we okay with `toString()` and `pack()` static methods returning empty string instead of null, as it was in the original Java file? Project builds successfully.
> 
>     2. I think we can transfer [Address_quotes](https://github.com/thunderbird/thunderbird-android/blob/fdb8655f3a50f5bddad29399c306acf33180a1a0/mail/common/src/test/java/com/fsck/k9/mail/Address_quoteAtoms.java) tests in `AddressTest` as Address_quotes tests are testing `quoteAtoms()` of Address class, wdyt?

Thanks for addressing all the comments! Answering your questions:

> 1. Are we okay with `toString()` and `pack()` static methods returning empty string instead of null, as it was in the original Java file? Project builds successfully.
 
> [!IMPORTANT]
> The original Java file, both `static toString` and `pack` return a **Nullable String**, not an empty string.

We need to preserve the current behaviour. The code will compile if there are no syntax errors, but we could face runtime errors (aka NullPointerException) because we changed the behaviour of a method by stopping it from accepting null references when converting it to Kotlin. 

The purpose of the conversion is to maintain the existing functionality, even if it might be deemed "incorrect" (I’m not sure if that’s the case here), as long as it doesn’t cause the app to crash. However, if we identify any erroneous behaviour, we should create a separate ticket to address it to ensure we don't overlook it. This, however, would be beyond the scope of the Java to Kotlin conversion task.

> 2. I think we can transfer [Address_quotes](https://github.com/thunderbird/thunderbird-android/blob/fdb8655f3a50f5bddad29399c306acf33180a1a0/mail/common/src/test/java/com/fsck/k9/mail/Address_quoteAtoms.java) tests in `AddressTest` as Address_quotes tests are testing `quoteAtoms()` of Address class, wdyt?

Honestly, I don't have a strong preference either way. It would certainly be beneficial to combine them, as it would also involve converting that class to Kotlin. However, I'm also fine with keeping it in a separate test class to prevent the `AddressTest` class from becoming too large. This should be outside the scope of this pull request.

If you’re interested in moving forward with this, I can create a ticket and assign it to you. The choice is up to you.

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,288 @@
...
+        /**
+         * Packs an address list into a String that is very quick to read
+         * and parse. Packed lists can be unpacked with unpackAddressList()
+         * The packed list is a ",\u0001" separated list of:
+         * address;\u0001personal
+         * @param addresses Array of addresses to pack.
+         * @return Packed addresses.
+         */
+        fun pack(addresses: Array<Address>?): String {
+            if (addresses == null) return ""
```
rafaeltonholo, line 222: This should return null instead. See https://github.com/thunderbird/thunderbird-android/blob/e7e3add2a8a2ce7598563a6282e4357147c4d53f/mail/common/src/main/java/com/fsck/k9/mail/Address.java#L249-L252

mail/common/src/main/java/com/fsck/k9/mail/Address.kt
```diff
@@ -0,0 +1,288 @@
...
+                Log.e(pe, "MimeException in Address.parse()")
+                // broken addresses are never added to the resulting array
+                EMPTY_ADDRESS_ARRAY
+            }
+        }
+
+        @JvmStatic
+        fun toString(addresses: Array<Address>): String =
+            if (addresses.isEmpty()) {
+                ""
```
rafaeltonholo, line 172: This method should accept a nullable Array and return a nullable String. See: https://github.com/thunderbird/thunderbird-android/blob/e7e3add2a8a2ce7598563a6282e4357147c4d53f/mail/common/src/main/java/com/fsck/k9/mail/Address.java#L192-L195

```suggestion
        fun toString(addresses: Array<Address>?): String? =
            if (addresses.isNullOrEmpty()) {
                null
```

You will also need to change `TextQuoteCreator.kt` adjusting to this change. Please use the following diff:
```diff
Index: legacy/core/src/main/java/com/fsck/k9/message/quote/TextQuoteCreator.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/legacy/core/src/main/java/com/fsck/k9/message/quote/TextQuoteCreator.kt b/legacy/core/src/main/java/com/fsck/k9/message/quote/TextQuoteCreator.kt
--- a/legacy/core/src/main/java/com/fsck/k9/message/quote/TextQuoteCreator.kt	(revision 9685530a53ebaf97fd71dabe2819887d7b330301)
+++ b/legacy/core/src/main/java/com/fsck/k9/message/quote/TextQuoteCreator.kt	(date 1772110638670)
@@ -27,7 +27,9 @@
 
     private fun prefixQuoteText(body: String, originalMessage: Message, prefix: String): String {
         val sentDate = quoteDateFormatter.format(originalMessage.sentDate)
-        val sender = Address.toString(originalMessage.from)
+        // TODO(#10610): investigate when originalMessage.from can be either null or empty.
+        //               To avoid compilation errors, we are using an empty string until we figure that out.
+        val sender = Address.toString(originalMessage.from).orEmpty()
 
         return buildString {
             val replyHeader = if (sentDate.isEmpty()) {

```

Issue #10501, opened 2026-02-09
Avatar image should support PNG


wmontwe:
The avatar image is currently limited to JPG only. The implementation is stable and we should also allow PNG images.

McBaumwolle:
Never contributed to a project this large, so this might not be perfect, but it worked on my end (via tests and on my phone).

PR #10505, opened 2026-02-09
feat(account): add support for PNG avatars

McBaumwolle:
Description
This PR updates the account avatar logic to support both `.jpg` and `.png` file extensions. Previously, all avatars were forced to use the `.jpg` extension, which caused issues when someone uploaded PNGs, especially those with transparency (I tried some for my main email and did not understand why it failed at first, or rather did not show the images in the picker).

Changes:
Dynamic extension: `LocalAvatarImageDataSource` now detects if the source image is a PNG and saves it with the correct `.png` extension.
* Smart cleanup: the `delete()` function now checks for and removes both `.jpg` and `.png` files for the given account. This ensures that, when a user switches from a JPEG avatar to a PNG avatar (or vice versa), the old file is deleted and does not remain orphaned.
* Tests: updated `LocalAvatarImageDataSourceTest` to verify PNG support and ensure the clean-up logic works for multiple file types.

Fixes:
#10501


McBaumwolle:
Never mind, needs some further adjustments.

Issue #10364, opened 2026-01-15
Rework funding feature for new Billing Library


wmontwe:
The funding feature must be updated to support the new Billing Library (migration from version 7.1.1 to 8.30).

With the updated API, historical one-time purchase data is no longer available. The previous method for retrieving purchase history is not supported in the new version.

To compensate for the missing purchase history, we will store the most recent one-time purchase locally and use it as the reference for purchase validation.

**Expected Result:**

- Purchase validation works correctly without relying on server-side purchase history.
- Users upgrading from previous app versions retain proper purchase recognition without regressions.

PR #10519, opened 2026-02-10
feat(configstore): add testing utilities and improve ConfigKey behavior

wmontwe:
Part of #10364 

This resolves an issue with not comparable config keys and adds a `TestConfigBackend` to ease test setups.


rafaeltonholo (approved):
LGTM

PR #10563, opened 2026-02-18
fix(message-list): fix display of longer date/time strings 

### LINKED ISSUE #10439: [ML 01/UI.04] - Update the Message List Compose Component for Variable Sized Date Field
The Jetpack Compose component for the Message List Item has a dynamic width for each column. This allows the date field to expand into the main subject and preview. The correct way to do this would be to allow each row to control the width, so the date could either only shorten one line, the sender name or subject, or allow it to go on to two lines when the full date is selected.

dani-zilla:
Fixes #10439 

# Why

The Date/Time field was included in the same column as the Star and Attachment icons. This would cause it to stretch the width of the column, cutting into the subject, sender, and message excerpt sections. Instead, it should not modify the width of the second line (subject or sender, depending on user option), and should be able to go onto a second line if the subject is too long. 

# How

Pulling the date out of this column, going with a 3 column design (until we split out new/unread icon from avatar), with the center content column broken up into rows. The Sender (first row) will also contain the date in a flow row, allowing the date to move to a second line if necessary. If they can't fit in the full row each, they'll end with an ellipsis. 

# Out of Scope 

Refactor of Message List items to confirm to new design and reduced complexity of MessageListItems 

messageListDateTimeFix

### DIFF
--- a/core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/organism/message/MessageItem.kt
+++ b/core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/organism/message/MessageItem.kt
@@ -118,16 +118,17 @@ internal fun MessageItem(
                     SenderText(
                         showAccountIndicator = showAccountIndicator,
                         accountIndicatorColor = accountIndicatorColor,
-                        modifier = Modifier.align(Alignment.CenterVertically),
+                        modifier = Modifier
+                            .align(Alignment.CenterVertically)
+                            .weight(1f, fill = false),
                     ) {
                         sender()
                     }
-                    if (!isCompact) {
-                        Spacer(modifier = Modifier.weight(1f))
-                    }
                     MessageItemDate(
                         receivedAt = receivedAt,
-                        modifier = Modifier.align(Alignment.CenterVertically),
+                        modifier = Modifier
+                            .align(Alignment.CenterVertically)
+                            .wrapContentWidth(),
                     )
                 }
             }
@@ -334,7 +335,7 @@ private fun MessageItemDate(
     TextTitleSmall(
         text = receivedAt,
         maxLines = 1,
-        overflow = TextOverflow.Ellipsis,
+        overflow = TextOverflow.Visible,
         modifier = modifier,
     )
 }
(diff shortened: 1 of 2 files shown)


core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/organism/message/MessageItem.kt
```diff
@@ -154,17 +154,69 @@ internal fun MessageItem(
...
+        ) {
+            senderOrSubject()
+        }
+        Spacer(modifier = Modifier.weight(1f))
+        MessageItemDate(
+            receivedAt = receivedAt,
+            modifier = Modifier.align(Alignment.CenterVertically),
+        )
+    }
+}
```
rafaeltonholo, line 198: The FlowRow should only be used when the screen size is within the small breakpoint (WindowSizeClass.Compact; we might need to create a new one called Small with the size of 350dp). You can utilize `movableContentOf` to place the same header row content inside the appropriate container based on this condition.

```suggestion
    val windowSizeInfo = getWindowSizeInfo()
    val isCompact = windowSizeInfo.screenWidthSizeClass == WindowSizeClass.Compact

    val headerRowContent =
        remember(showAccountIndicator, accountIndicatorColor, receivedAt, senderOrSubject, isCompact) {
            movableContentOf<RowScope> { scope ->
                with(scope) {
                    SenderText(
                        showAccountIndicator = showAccountIndicator,
                        accountIndicatorColor = accountIndicatorColor,
                        modifier = Modifier.align(Alignment.CenterVertically),
                    ) {
                        senderOrSubject()
                    }
                    if (!isCompact) {
                        Spacer(modifier = Modifier.weight(1f))
                    }
                    MessageItemDate(
                        receivedAt = receivedAt,
                        modifier = Modifier.align(Alignment.CenterVertically),
                    )
                }
            }
        }

    when (windowSizeInfo.screenWidthSizeClass) {
        WindowSizeClass.Compact -> FlowRow(
            verticalArrangement = Arrangement.Center,
            horizontalArrangement = Arrangement.Start,
            maxLines = 2,
            modifier = modifier
                .fillMaxWidth()
                .defaultMinSize(
                    minHeight = AccountIndicatorIcon.ACCOUNT_INDICATOR_DEFAULT_HEIGHT,
                ),
        ) {
            headerRowContent(this)
        }

        else -> Row(
            horizontalArrangement = Arrangement.SpaceBetween,
            modifier = modifier.defaultMinSize(
                minHeight = AccountIndicatorIcon.ACCOUNT_INDICATOR_DEFAULT_HEIGHT,
            ),
        ) {
            headerRowContent(this)
        }
    }
```

dani-zilla:
Had to rebase a merge conflict. Should be good to go now

PR #10568, opened 2026-02-20
fix: add Toast to warn if attempting to send a message while offline

### LINKED ISSUE #10134: Warn "you are offline"  when attempting to edit or send a message
When composing or trying to send a message, if the user is offline at least let them know. 

Sure it just goes into the outbox for a while until the user finally connects. Yes but they thought they sent the boss a message a half an hour ago only to realize now they weren't even connected to the Internet.

mohitsatr:
fixes #10134 

**Behaviour**

https://github.com/user-attachments/assets/0c9de995-4635-4fbf-aa2e-3e8bfd2f484a


mohitsatr:
@wmontwe Got it. Fixed

Issue #10566, opened 2026-02-19
Fix: Add ISO to Message List Item date time format customization


alpharesearch:
### Checklist

- [x] I have used the search function to see if someone else has already submitted the same bug report.
- [x] I will describe the problem with as much detail as possible.

### App

Thunderbird for Android

### App version

nightly 

### Where did you get the app from?

Other

### Android version

V16

### Device model

android studio 

### Steps to reproduce

After test implementing and testing the Auto and Full modes locally, I noticed one gap that might be valuable to address for power users.
​I suggest we add a third option: ISO 8601.
​Rationale:
While Full is great because it respects the Android system locale, technical users often prefer a deterministic, unambiguous format (YYYY-MM-DD HH:mm) that doesn't change based on the device language.
​Full (Locale US): 1/17/2026 12:45 PM
​Full (Locale DE): 17.01.2026 12:45
​ISO 8601: 2026-01-17 12:45 (Always the same)
​Since the architecture (Enum/ListPreference) is already being put in place, adding ISO to the MessageListDateTimeFormat enum would be a low-effort addition that aligns well with Thunderbird's technical user base.

### Expected behavior

​ISO 8601: 2026-01-17 12:45 (Always the same)

### Actual behavior

not yet implemented 
https://github.com/thunderbird/thunderbird-android/issues/10202#issuecomment-3927852088
@rafaeltonholo per your request, here is the is the bug, thank you.

PR #10574, opened 2026-02-20
feat: Add setting to display ISO date and time in message list.

alpharesearch:
This PR adds an **ISO 8601** (`YYYY-MM-DD HH:mm`) option to the message list date/time display settings.

**Context:**
While the `Full` setting is great for respecting device locales, power users and technical users often prefer the deterministic, unambiguous ISO format, which sorts perfectly and remains consistent regardless of the system language. 

**Changes:**
- Added `ISO` enum to `MessageListDateTimeFormat`
- Added formatting logic to `RelativeDateTimeFormatter` using `Locale.US` to ensure strict standard compliance.
- Added UI string for the settings menu.
- Added corresponding unit tests.

**Related Issue:**
Relates to #10566 (Fix: Add ISO to Message List Item date time format customization.

**Screenshots:**
*(Drag and drop your screenshots of the settings menu and the message list here!)*
image

image

### DIFF
--- a/core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/display/visualSettings/message/list/MessageListDateTimeFormat.kt
+++ b/core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/display/visualSettings/message/list/MessageListDateTimeFormat.kt
@@ -23,9 +23,8 @@ enum class MessageListDateTimeFormat {
     Full,
 
     /**
-     * ISO 8601 option, which is often requested by power users who prefer a
-     * deterministic, unambiguous format over the system locale defaults.
-     *      - Examples: "2026-01-23 13:23"
+     * ISO 8601 - like format
+     *  - Examples: "2026-01-23 13:23"
      */
     ISO,
 }
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/helper/RelativeDateTimeFormatter.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/helper/RelativeDateTimeFormatter.kt
@@ -9,9 +9,11 @@ import android.text.format.DateUtils.FORMAT_SHOW_DATE
 import android.text.format.DateUtils.FORMAT_SHOW_TIME
 import android.text.format.DateUtils.FORMAT_SHOW_WEEKDAY
 import android.text.format.DateUtils.FORMAT_SHOW_YEAR
+import java.text.SimpleDateFormat
 import java.util.Calendar
 import java.util.Calendar.DAY_OF_WEEK
 import java.util.Calendar.YEAR
+import java.util.Locale
 import kotlin.time.Clock
 import kotlin.time.ExperimentalTime
 import kotlin.time.Instant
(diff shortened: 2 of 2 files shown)


core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/display/visualSettings/message/list/MessageListDateTimeFormat.kt
```diff
@@ -21,4 +21,11 @@ enum class MessageListDateTimeFormat {
      *      - Examples: "1/23/2026 1:23PM" "23/1/2026 13:23"
      */
     Full,
+
+    /**
+     * ISO 8601 option, which is often requested by power users who prefer a
+     * deterministic, unambiguous format over the system locale defaults.
+     *      - Examples: "2026-01-23 13:23"
```
dani-zilla: I think this could just include the relevant data, something like:
```
/**
 * ISO 8601-like format
 *     - Example: "2026-01-23 13:23"
 */
```
We don't have to explain why we're including it in the code (maybe the PR, but not the code).

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/helper/RelativeDateTimeFormatter.kt
```diff
@@ -31,17 +31,25 @@ constructor(
...
+                    else -> FORMAT_SHOW_DATE or FORMAT_SHOW_YEAR or FORMAT_NUMERIC_DATE
+                }
+                DateUtils.formatDateRange(context, timestamp, timestamp, flags)
+            }
+            MessageListDateTimeFormat.Full -> {
+                val flags = FORMAT_SHOW_TIME or FORMAT_SHOW_DATE or FORMAT_SHOW_YEAR or FORMAT_NUMERIC_DATE
+                DateUtils.formatDateRange(context, timestamp, timestamp, flags)
+            }
+            MessageListDateTimeFormat.ISO -> {
+                val sdf = java.text.SimpleDateFormat("yyyy-MM-dd HH:mm", java.util.Locale.US)
```
dani-zilla: You should be able to add `import java.text.SimpleDateFormat` to not need the `java.text.` prefix here.

The Locale should be based on the user's locale. `Locale.getDefault()` should work.

I would normally recommend using the  `DateTimeFormatter` constant for ISO date/time, but I think for this usage, a space makes more sense here instead of the standard `T`, as it's more legible. This would be displaying with a large amount of text around it, and skimming legibility feels more important than strict adherence to ISO. Also, that would be `"${DateTimeFormatter.ISO_LOCAL_DATE} HH:mm"`, and I'm not sure if it's worth it to pull in the date time formatter just to pull in the localized date, so I think leaving it as a string like this is fine.

dani-zilla (changes requested):
Thanks for the contribution! The expansion of date/time formats we could display in message list items was something we were considering when adding contextual and full date formats, and I had actually pictured ISO being a date/time format we could add in the future. Glad to see it being used like this. I just had a few small suggestions before we can merge it.

alpharesearch:
Hi @dani-zilla, 

Thanks for the review and the helpful feedback! I agree that `Locale.getDefault()` makes more sense here for the localized digits.

I've pushed a new commit that addresses all your suggestions:
* Simplified the KDoc comment for the `ISO` enum to just show the format and example.
* Added the proper imports for `SimpleDateFormat` and `Locale` to clean up the inline code.
* Switched the locale from `Locale.US` to `Locale.getDefault()`.

Let me know if everything looks good now or if there is anything else I should adjust!

PR #10580, opened 2026-02-24
fix: App drawer closes upon account selection

### LINKED ISSUE #10120: App drawer no longer closes upon account selection, regression??
### Checklist

### App

Thunderbird for Android

### App version

TB v14 and v15.1

Other

### Android version

Android 15

### Device model

Oneplus 10 Pro

### Steps to reproduce

I'm raising this ticket in response to another users issue reported on SUMO.
https://support.mozilla.org/en-US/questions/1548980

> In the older K-9, when one changed the active account, the inbox for the newly selected account would just be shown immediately. No need for any other operations on my part.

> I am really commenting on how the side bar does not get immediately out of the way after choosing a different account to reveal the newly selected account's inbox - OPENED.

CKetti advises how this should function in PR [#5477](https://github.com/thunderbird/thunderbird-android/pull/5477)

> Previously, when selecting an account the drawer would stay open and display the folder list of the newly selected account. Now, it only stays open if the auto-expand folder is set to None. If an auto-expand folder is configured the message list for that folder is displayed and the drawer is closed.

In the current app version, the account side drawer remains open after selecting an account, irrespective of a folder being assigned in settings > account > folders > auto-expand folder.

**Is this a regression or a design decision?**

Steps to reproduce.
1. Have 2 or more accounts
2. Set the auto-expand folder to "Inbox" on each account via settings > account > folders
3. Open the account drawer and select account.

### Expected behavior

The account side drawer automatically closes and displays the "Inbox" folders message list

### Actual behavior

The account side drawer remains open

### Logs

_No response_

mohitsatr:
Fixes #10120 

Behaviour: 

**Bin** is set as AutoExpandFolder for both accounts

https://github.com/user-attachments/assets/c5b84f84-4640-48dc-b56a-bce9d9c971d8

### DIFF
--- a/feature/navigation/drawer/dropdown/src/debug/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/FakeData.kt
+++ b/feature/navigation/drawer/dropdown/src/debug/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/FakeData.kt
@@ -55,6 +55,7 @@ internal object FakeData {
         unreadMessageCount = 0,
         starredMessageCount = 0,
         hasError = false,
+        hasAutoExpandFolder = false,
     )
 
     val FOLDER = Folder(
@@ -162,6 +163,7 @@ internal object FakeData {
                 unreadMessageCount = 2,
                 starredMessageCount = 0,
                 hasError = false,
+                hasAutoExpandFolder = false,
             ),
             MailDisplayAccount(
                 id = "account2",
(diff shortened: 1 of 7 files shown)


mohitsatr:
I feel creating a use case for getting the `autoExpandAccount` Id is an unnecessary abstraction. Is there a better way to do this?
 In fact, we don't even need ID, just the boolean value to let us know whether or not `AutoExpandAccount` is set.

wmontwe:
> I feel creating a use case for getting the `autoExpandAccount` Id is an unnecessary abstraction. Is there a better way to do this? In fact, we don't even need ID, just the boolean value to let us know whether or not `AutoExpandAccount` is set.

A separate use case is not needed, you could update the `GetDisplayAccounts` and the `DisplayAccount` to hold that property.

mohitsatr:
@wmontwe Please review

PR #10585, opened 2026-02-24
chore: declare python dependencies

### LINKED ISSUE #10584: Resolve python related fluidattacks findings
Fluidattacks has identified problems related to Python libraries used within the scripts directory – these libraries are not version-pinned. To resolve this, we need to explicitly declare the required versions for each library.

wmontwe:
Resolves #10584

### DIFF
--- a/scripts/test_python_scripts.sh
+++ b/scripts/test_python_scripts.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+# Test Python scripts and dependencies in a temporary environment
+# Usage: ./scripts/test_python_scripts.sh
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+echo "Python Scripts Test"
+echo "==================="
+echo ""
+
+# Check if Python is available
+if ! command -v python3 &> /dev/null; then
+    echo "✗ Error: python3 not found"
+    exit 1
+fi
+
+echo "✓ Python: $(python3 --version)"
+
+# Create temporary virtual environment
+VENV_DIR=$(mktemp -d)
+python3 -m venv "$VENV_DIR" > /dev/null 2>&1
+source "$VENV_DIR/bin/activate"
+
+# Install dependencies
+echo "Installing dependencies..."
+pip install --quiet --upgrade pip
+pip install --quiet -r "$SCRIPT_DIR/requirements.txt"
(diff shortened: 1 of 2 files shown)


scripts/requirements.txt
```diff
@@ -1,4 +1,20 @@
...
+# YAML parser for release notes
+PyYAML==6.0.1
+
+# Required by: scripts/ci/render-notes.py
+# Template engine for rendering changelog files
+Jinja2==3.1.6
+
+# Required by: scripts/ci/setup_release_automation, scripts/ci/render-notes.py
+# HTTP library for GitHub API interactions
+requests==2.32.4
```
coreycb, line 20: Will dependabot update these in the future?

scripts/test_python_scripts.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+# Test Python scripts and dependencies in a temporary environment
+# Usage: ./scripts/test_python_scripts.sh
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
```
coreycb, line 8: I think this is unused

scripts/test_python_scripts.sh
```diff
@@ -0,0 +1,54 @@
...
+pip install --quiet -r "$SCRIPT_DIR/requirements.txt"
+echo "✓ Dependencies installed"
+echo ""
+
+# Test scripts compile
+echo "Testing script compile..."
+echo ""
+python3 -m py_compile "$SCRIPT_DIR/ci/render-notes.py" && echo "  ✓ render-notes.py"
+python3 -m py_compile "$SCRIPT_DIR/ci/setup_release_automation" && echo "  ✓ setup_release_automation"
+python3 -m py_compile "$SCRIPT_DIR/ci/merges/merge_gradle.py" && echo "  ✓ merge_gradle.py"
```
coreycb, line 39: One more comment, not sure if we want to use `set -e` or not because the script will bail early on failure. Not sure if if that affects these.

scripts/requirements.txt
wmontwe: I'll create a dependabot config

scripts/test_python_scripts.sh
wmontwe: Right, if any of these fails, there will be no cleanup

wmontwe:
@coreycb I overlooked your comments, here is a fix #10612

PR #10586, opened 2026-02-24
refactor: change FolderType to use @Discouraged annotation

wmontwe:
This removed the deprecation warnings during compilation to reduce noise.


rafaeltonholo (approved):
LGTM

PR #10604, opened 2026-02-25
chore(message-list): move swipe_action_* from :legacy:ui to :feature:mail:message:list:api

rafaeltonholo:
Part of #10603

Move all swipe_action_* string resources to :feature:mail:message:list:api


wmontwe (approved):
LGTM

PR #10624, opened 2026-02-28
Use fixed width font for composing as well (new messages and replies)

AlainKnaff:
This pull request makes thunderbird use a fixed width font for composing a well (both new message and reply), if user has chosen a fixed width font for viewing.

_Fixes #907_ , at least partly (re-uses fixed-width option for viewing, rather than create a new one. Creating a new setting seems to not work in Thunderbird currently, that new setting won't be remembered)

This aligns Thunderbird on Android's behavior more with Android on desktop's behavior where composition can use a fixed width font since long ago.

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
@@ -22,6 +22,7 @@ import android.content.Intent;
 import android.content.IntentSender;
 import android.content.IntentSender.SendIntentException;
 import android.content.pm.ActivityInfo;
+import android.graphics.Typeface;
 import android.content.pm.PackageManager;
 import android.net.Uri;
 import android.os.AsyncTask;
@@ -543,6 +544,8 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
         recipientMvpView.setFontSizes(K9.getFontSizes(), fontSize);
         quotedMessageMvpView.setFontSizes(K9.getFontSizes(), fontSize);
         K9.getFontSizes().setViewTextSize(subjectView, fontSize);
+        if(generalSettingsManager.getConfig().getDisplay().getVisualSettings().isUseMessageViewFixedWidthFont())
+            messageContentView.setTypeface(Typeface.MONOSPACE);
         K9.getFontSizes().setViewTextSize(messageContentView, fontSize);
         K9.getFontSizes().setViewTextSize(signatureView, fontSize);
 
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/compose/QuotedMessageMvpView.java
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/compose/QuotedMessageMvpView.java
@@ -7,6 +7,7 @@ import android.view.View.OnClickListener;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
 import android.widget.EditText;
+import android.graphics.Typeface;
 
 import app.k9mail.legacy.di.DI;
 import com.fsck.k9.FontSizes;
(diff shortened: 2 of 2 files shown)


dani-zilla:
Do you have some test screenshots? I don't seem to be able to produce a fixed result in the TB app. Monospace on does show in the reader for plain text messages, as expected, but not in the message compose view, which this should fix.

AlainKnaff:
> Do you have some test screenshots? I don't seem to be able to produce a fixed result in the TB app. Monospace on does show in the reader for plain text messages, as expected, but not in the message compose view, which this should fix.

Here is one example with a reply (Thunderbird on Desktop):

Screenshot_Thunderbird_Reply

AlainKnaff:
And here are 2 with the mobile app (after my proposed changes)

Screenshot_20260311-192403
Screenshot_20260311-192255

AlainKnaff:
And finally, new message for desktop thunderbird
Screenshot_Thunderbird_New_Message

dani-zilla:
I'm just not seeing it. For example, here's the setting turned off and on, on the main branch and on your branch, both in the reader and the compose view. In the latter, it's not using a monospaced font. 

combinedTestShowingNoChangeTB

AlainKnaff:
I'm just wondering now whether you are actually running with my changes or accidentally have an unpatched version? Maybe inserting some Log.i statements near my changes (both inside and outside the condition) to see whether they show up in logcat?

dani-zilla:
It seems to hit the code while debugging, and it shows the monotype font on the preview, but it never shows it in the compose message view. I tried a clean build and uninstalling/reinstalling, as well as using both my personal device and the simulator. Nothing showed monotype font on the compose view.

AlainKnaff:
You mean, you see `messageContentView.setTypeface(Typeface.MONOSPACE);` and `mQuotedText.setTypeface(Typeface.MONOSPACE);` being hit, but they have no effect?

or could it be that somehow your font definitions were messed up, i.e. the monospace font was not actually monospace? (not sure whether that such a thing is possible in Android...)
Or some code, which would only execute under certain conditions (?), would undo the effect of these statements?

dani-zilla:
The setting is detected properly, the code is executed, but monotype is not set. I'm not entirely sure why, but I have two devices that do not display it. We may have to test it further

AlainKnaff:
> The setting is detected properly, the code is executed, but monotype is not set. I'm not entirely sure why, but I have two devices that do not display it. We may have to test it further

So, how do we go from here?

dani-zilla:
Do you also see a fixed width font in the new message area above the quoted text? I think I'll have a few other people look at it to see if anyone else sees monospace font on the quoted message or in the reply text area.

AlainKnaff:
Do you mean the attribution line (just above quoted text)? => Yes, that's fixed too.
Or do you mean the "Message text" label? => Telling by the M, that's fixed too.

Or do you mean anything in the header, such as the Subject? => That's proportional. However, a fixed font is less needed here, IMHO, as alignment plays a lesser role.

dani-zilla:
savedScreenshotMonoWithQuote

AlainKnaff:
Yes

AlainKnaff:
Just on the off chance that between the time I submitted this, and you tried it something might have changed elsewhere that broke it, I synced my fork, recompiled, and retried: it still works, both in replies and in new messages.

A number of apks are build, I use the ./app-thunderbird/build/outputs/apk/foss/debug/app-thunderbird-foss-debug.apk

AlainKnaff:
With app-thunderbird-full-debug.apk, it works too

dani-zilla:
Had a few people here test it and we found something funny, it is a per-device issue. I found an old Pixel and got it set up to find that the composer text will be monospace but the quoted text won't be until you go to edit it. No situation makes it monospace on at least two Motorola phones. So, seems it's one of those fun Android distro quirks.

AlainKnaff:
Thanks for having taken this despite it not working on some phones :-)

PR #10625, opened 2026-02-28
fix(drawer): keep legacy account selector available as explicit opt-in

libnewton:
## Why
Many users still miss the old account selection drawer.
This change ensures that path remains available, but it is **purely opt-in**: nothing is enabled automatically and default drawer behavior does not change unless the user explicitly changes it in settings.
## What changed
- Re-introduce old siderail account menu based off K9 v13.
- Added tests

## Behavior impact
- **Opt-in only:** account selector visibility changes only when the user triggers it.
- **No default behavior change:** this PR does not force-enable legacy account selection drawer behavior for all users.
## Screenshot
tempFileForShare_20260228-205534

 ## Risks / Trade-offs
- Low risk: mostly test/fake alignment and formatting.
- Main risk is future behavior drift in selector toggling; updated tests should catch regressions.

**AI Assistance / Code Authors**
The code mainly was ported from v13 of K9.
This PR includes AI-assisted changes. I however checked each line of code myself and ensured the AI-code is solid.


Alecaddd:
Thank you so much for your contribution and your suggestion.

Unfortunately this diverges from the product direction and vision of the future of Thunderbird for Android.
The account drawer is also not fully completed on our side so more iterations are coming in the near future.
Since we're a very small team, we are not able to support all features, no matter how self contained they are, as that adds maintenance burden for a feature we're not planning to support.

Also, this PR is very difficult to review.
If you're planning to submit other proposals, or maybe continue with proposing this feature, we would really appreciate if you could split this PR into smaller chunks, making it easier to review and land it in small parts.
Thank you.

libnewton:
alright I will fork then

PR #10649, opened 2026-03-06
refactor: funding to kmp module

wmontwe:
Part of #10364 

This pull request refactors the funding feature modules to KMP modules. Removed direct `Activity` dependency in he `FundingManager` and refactors the funding reminder logic to use the `ActivityProvider` instead.

The `googleplay` module was not changed to kmp as it depends on the design system that is not KMP yet.


dani-zilla (approved):
Seems to work as well as it can. Admittedly haven't been able to get the funding working with the Play Billing Lab yet, but it appears the same as it does on the other branch (and now main).

wmontwe:
@dani-zilla I guess we need a signed app, I'll push the actual billing library update and add a signed test version. Not sure why it works on my side without having to sign the app 🤷

PR #10658, opened 2026-03-08
fix: smooth drawer animation and respect animation preferences

### LINKED ISSUE #10480: Animation in the sidebar when changing accounts is too fast and hectic
### Checklist

### App

K-9 Mail

### App version

16.0

F-Droid

### Android version

GrapheneOS

### Device model

Google Pixel 6

### Steps to reproduce

1. Open sidebar
2. Tap on "Unified Account" to switch to another account
3. Tap on one account

### Expected behavior

The animation to display the account list (step 2) or the folders for the selected account (step 3) should be smooth and not irritating.

### Actual behavior

With version 16 the animation has changed. It is now too fast and hectic, irritating.

### Logs

_No response_

jasi381:
## Summary
- Fixes #10480 — Sidebar animation too fast/hectic
- Increases drawer animation duration from 200ms to 300ms with `FastOutSlowInEasing` for a smoother feel
- Adds `isShowAnimations` to `DrawerConfig` so the drawer respects the user's "Show animations" setting
- When animations are disabled (via app settings or OS-wide), all drawer transitions use `snap()` (instant, no animation) instead of slide/fade

## Test plan
- [x] Open sidebar and toggle the account switcher — animation should feel smooth (not jarring)
- [x] Go to Settings → disable "Show animations" → open sidebar and toggle account switcher — transitions should be instant (no animation)
- [x] Set Developer Options → Window/Transition animation scale = 0 → verify drawer transitions are instant
- [x] Verify folder list expand/collapse icon still works correctly
- [x] Verify switching between accounts animates the account name smoothly

### DIFF
--- a/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/DrawerContent.kt
+++ b/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/DrawerContent.kt
@@ -1,5 +1,6 @@
 package net.thunderbird.feature.navigation.drawer.dropdown.ui
 
+import android.provider.Settings
 import androidx.compose.animation.AnimatedContent
 import androidx.compose.animation.AnimatedContentTransitionScope
 import androidx.compose.animation.ContentTransform
@@ -20,7 +21,6 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.unit.IntOffset
-import android.provider.Settings
 import app.k9mail.core.ui.compose.designsystem.atom.DividerHorizontal
 import app.k9mail.core.ui.compose.designsystem.atom.Surface
 import app.k9mail.core.ui.compose.theme2.MainTheme


dani-zilla:
The detekt issue should be a simple fix:
```
Task :feature:navigation:drawer:dropdown:detekt FAILED
/home/runner/work/thunderbird-android/thunderbird-android/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/common/AnimatedExpandIcon.kt:15:14: Parameters in a composable function should be ordered following this pattern: params without defaults, modifiers, params with defaults and optionally, a trailing function that might not have a default param.
Current params are: [isExpanded: Boolean, isShowAnimations: Boolean = true, modifier: Modifier = Modifier, tint: Color? = null] but could be [isExpanded: Boolean, modifier: Modifier = Modifier, isShowAnimations: Boolean = true, tint: Color? = null].
```
Basically just swapping the `isShowAnimations: Boolean = true` to be after the modifier param. 

Besides that, would you be willing to make a screen recording of this issue? I seem to be having trouble reproducing the initial issue to test the fix.

Alecaddd:
Thank you so much for this contribution.
I would recommend to simply listen for the OS-wide animation disabled state rather than adding an extra option inside the application.
From a UX point of view, it feels weird to only have an app with disabled animations but not the OS, or vice versa, so I think this feature should be always in sync with the OS and respect the global users' choice.

jasi381:
> The detekt issue should be a simple fix:
> 
> ```
> Task :feature:navigation:drawer:dropdown:detekt FAILED
> /home/runner/work/thunderbird-android/thunderbird-android/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/common/AnimatedExpandIcon.kt:15:14: Parameters in a composable function should be ordered following this pattern: params without defaults, modifiers, params with defaults and optionally, a trailing function that might not have a default param.
> Current params are: [isExpanded: Boolean, isShowAnimations: Boolean = true, modifier: Modifier = Modifier, tint: Color? = null] but could be [isExpanded: Boolean, modifier: Modifier = Modifier, isShowAnimations: Boolean = true, tint: Color? = null].
> ```
> 
> Basically just swapping the `isShowAnimations: Boolean = true` to be after the modifier param.
> 
> Besides that, would you be willing to make a screen recording of this issue? I seem to be having trouble reproducing the initial issue to test the fix.

Thanks for pointing that out! I'll swap the `isShowAnimations` parameter after `modifier` and push an update shortly. Also I'll record a short video showing the original issue so it's easier to reproduce.

jasi381:
Thanks for the feedback!

For the animation improvements, I adjusted the drawer animation to make it smoother by increasing the duration from **200ms to 300ms** and applying **FastOutSlowInEasing** so the motion feels less abrupt.

I'm currently addressing the **Detekt parameter order issue** mentioned by @dani-zilla and will push the fix shortly.

Regarding the UX suggestion from @Alecaddd, I’m also working on updating the implementation so the app respects the **OS-wide animation preference** instead of relying on a separate app-level option.

I've attached short screen recordings below showing the behaviour **before and after the animation changes** to help with reproducing the issue.

**Before fix (original behaviour):**

https://github.com/user-attachments/assets/fd3119b3-992e-4f28-b84a-d303def1080d

**After animation improvement:**

https://github.com/user-attachments/assets/0024ee3f-74fe-46d0-8004-75b9162e57d2

untitaker:
Just glanced over the code and didn't test it, but this fix seems to make the drawer respect system settings, but the application setting ("Use gaudy") still exists with an independent default. This means that all non-drawer animations still do not respect the system-wide settings. I think the correct fix is to refactor the setting (not any views) to a triple-state where the default is to inherit the system-wide setting. I'm working on that fix right now.

dani-zilla:
Soft freeze lifted, so I wanted to come back to this. Spoke with a few other people and they do see the improvement in the animation this provides. We do need to get the Spotless issue fixed though. Locally you can run `./gradlew spotlessCheck` to see what's wrong. Then you can either fix it yourself or do `./gradlew spotlessApply` to automatically fix the issues. 

As for the suggestion to blend the system and local settings, I tend to agree. Having all the decision making for whether or not to display animations in one place does seem preferable to potentially needing to check the system settings everywhere it's implemented. 

I think I'd be okay with merging this regardless though. It would be a small change to make it rely on the local settings alone later, and could fix the problem while also improving the animation now.

Issue #10554, opened 2026-02-17
[ML 0.1/UI.01.7] - Restructure of Message Item UI body


rafaeltonholo:
## Description
We conducted a review of our new design and need to restructure the content in the message list item body.

## Goal
The user interface of the Message List Item body must align with the [high-fidelity message list item design](https://www.figma.com/design/9xVi5XMbfpqVULr9Fe39ZL/TFAndroid_2025_Q4_MessageList_v02.0.0?node-id=2448-15854&t=tszVi1uohoinjWXf-4) in Figma.

## Acceptance Criteria
- **First row:**
   - Account Indicator (#10446)
   - Sender (display the Subject if `swapSenderWithSubject` is `true`)
   - Date (#10439); show this on larger screens, but on smaller screens, move it to the second row and push down the other rows.
  
- **Second row:**
  - In case users set excerpt max lines to 0:
      - Thread Number, if greater than 1
      - Attachment Icon, if applicable
   - Subject (display the Sender if `swapSenderWithSubject` is `true`)
  
- **Third row (inlined):**
  - In case users set excerpt max lines > 0:
      - Thread Number, if greater than 1
      - Attachment Icon, if applicable
  - Excerpt (Preview text). If the excerpt is multiline, wrap the text so that the icons appear integrated with the text. Add an overflow mechanism if the excerpt exceeds the maximum lines defined by the user configuration.

## Out of Scope
- Modifications to the trailing icons column
- Changes to the Avatar column
- Alterations to the Excerpt when the message is encrypted
- Changes to font styles
- Adjustments to the date field (for now, we will only include the date field slot; changes to the date will be addressed in #10439).
- Sender with conversation participants (#xxxxxxxx)

PR #10679, opened 2026-03-12
refactor(design-system): rework on SwipeableRow to include Dismiss, Reveal and Reset behaviours; ditch out from M3 implementation

rafaeltonholo:
Part of #10554.

* Introduced the new `SwipeBehaviour` sealed interface, which encapsulates swipe behaviour (threshold, animation, auto-reset, and disabled state) and replaces the previous threshold-based approach. This allows for more flexible and reusable swipe configurations. (`SwipeBehaviour.kt`)
* Refactored the swipeable row item API to accept a `SwipeBehaviour` object and a set of swipe directions, replacing the old per-direction threshold and enabling/disabling flags. This simplifies usage and improves composability. (`SwipeableRowItems.kt`)

These changes make swipeable row components more flexible, accessible, and easier to use throughout the codebase.

### DIFF
--- a/app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/molecule/items/SwipeableRowItems.kt
+++ b/app-ui-catalog/src/main/kotlin/net/thunderbird/ui/catalog/ui/page/molecule/items/SwipeableRowItems.kt
@@ -111,10 +111,10 @@ fun SwipeableRowItems(modifier: Modifier = Modifier) {
             ) { threshold -> SwipeBehaviour.Reveal(threshold = threshold) }
             SwipeSection(
                 threshold = thresholdDp,
-                subtitle = "Swipe to Reveal with Auto reset",
+                subtitle = "Swipe Action",
                 onSwipeEnd = ::onSwipeEnd,
             ) { threshold ->
-                SwipeBehaviour.Reveal(threshold = threshold, autoReset = true)
+                SwipeBehaviour.Action(threshold = threshold)
             }
         }
     }
@@ -270,10 +270,14 @@ private fun SwipeableRowItems(
                     text = backgroundItemText(direction),
                     modifier = Modifier
                         .then(
-                            if (behaviour is SwipeBehaviour.Reveal) {
-                                Modifier.width(behaviour.threshold)
-                            } else {
-                                Modifier.fillMaxWidth()
+                            when (behaviour) {
+                                is SwipeBehaviour.Reveal, is SwipeBehaviour.Action -> {
+                                    Modifier.width(behaviour.threshold)
+                                }
+
+                                else -> {
+                                    Modifier.fillMaxWidth()
+                                }
                             },
                         )
                         .swipeableCommonTextPadding(),
@@ -356,7 +360,7 @@ private val SwipeBehaviour.actionId: Int
     get() = when (this) {
         SwipeBehaviour.Disabled -> -1
         is SwipeBehaviour.Dismiss -> R.string.swipe_accessibility_dismiss_custom_action
-        is SwipeBehaviour.Reveal if autoReset -> R.string.swipe_accessibility_reveal_and_reset_custom_action
+        is SwipeBehaviour.Action -> R.string.swipe_accessibility_reveal_and_reset_custom_action
         is SwipeBehaviour.Reveal -> R.string.swipe_accessibility_reveal_custom_action
     }
 
--- a/core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeBehaviour.kt
+++ b/core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeBehaviour.kt
@@ -12,7 +12,6 @@ import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.milliseconds
-import net.thunderbird.core.ui.compose.designsystem.molecule.swipe.SwipeBehaviour.Companion.DEFAULT_AUTO_RESET_DELAY_MILLIS
 
 /**
  * Defines the behaviour of a swipeable component when a swipe gesture is
@@ -42,28 +41,30 @@ sealed interface SwipeBehaviour {
     /**
      * Defines a reveal swipe behaviour that allows content to be shown or hidden
      * through swipe gestures.
-     *
-     * @property autoReset Whether the row should automatically return to its initial
-     *  position after being revealed after a while. Defaults to `false`
-     * @property autoResetDelayMillis The duration to wait before automatically resetting
-     *  the revealed state back to the initial position. Only applies when [autoReset]
-     *  is enabled. Defaults to [DEFAULT_AUTO_RESET_DELAY_MILLIS].
      */
     data class Reveal(
         override val threshold: Dp = DEFAULT_THRESHOLD,
         override val settleAnimationSpec: AnimationSpec<Float> = DefaultSettleAnimation,
         override val enableHapticFeedback: Boolean = true,
-        val autoReset: Boolean = false,
-        val autoResetDelayMillis: Duration = DEFAULT_AUTO_RESET_DELAY_MILLIS.milliseconds,
+    ) : SwipeBehaviour
+
+    /**
+     * Defines a swipe behaviour that triggers an action and then automatically
+     * resets the swipe state after a specified delay.
+     *
+     * @property autoCloseDelayMillis The duration after which the swipe state will be reset
+     * to its default state.
+     */
+    data class Action(
+        override val threshold: Dp = DEFAULT_THRESHOLD,
+        override val settleAnimationSpec: AnimationSpec<Float> = DefaultSettleAnimation,
+        override val enableHapticFeedback: Boolean = true,
+        val autoCloseDelayMillis: Duration = DEFAULT_AUTO_RESET_DELAY_MILLIS.milliseconds,
     ) : SwipeBehaviour
 
     /**
      * Defines a reveal swipe behaviour that allows content to be dismissed by swiping
      * past the threshold.
-     *
-     * This behaviour differs from Reveal in that it is typically used for destructive
-     * or final actions where the row should be removed from view after completing the
-     * swipe gesture.
      *
      * @property dismissTransition The exit transition that will be applied when the
      *  item is dismissed.
--- a/core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeableRowState.kt
+++ b/core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeableRowState.kt
@@ -176,9 +176,8 @@ class SwipeableRowState internal constructor(
         if (swipeState != SwipeState.Swiping && swipeState != SwipeState.Revealed) return@DraggableState
 
         // When Revealed with non-auto-reset, only allow closing gestures (toward zero)
-        if (swipeState == SwipeState.Revealed) {
-            val revealBehaviour = activeBehaviour as? SwipeBehaviour.Reveal
-            if (revealBehaviour != null && !revealBehaviour.autoReset && !isClosingGesture(delta)) {
+        if (swipeState == SwipeState.Revealed && activeBehaviour is SwipeBehaviour.Action) {
+            if (!isClosingGesture(delta)) {
                 return@DraggableState
             }
         }
@@ -188,7 +187,7 @@ class SwipeableRowState internal constructor(
         val targetBehaviour = if (targetOffset >= 0f) startToEndBehaviour else endToStartBehaviour
         val targetMaxOffset = when (targetBehaviour) {
             is SwipeBehaviour.Dismiss -> layoutWidth
-            is SwipeBehaviour.Reveal ->
+            is SwipeBehaviour.Reveal, is SwipeBehaviour.Action ->
                 (targetBehaviour.percentageThreshold * layoutWidth) + SWIPE_BEHAVIOUR_REVEAL_EXTENSION
 
             is SwipeBehaviour.Disabled -> 0f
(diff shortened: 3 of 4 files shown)


core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeableRow.kt
```diff
@@ -28,105 +34,111 @@ import net.thunderbird.core.ui.compose.designsystem.molecule.swipe.SwipeDirectio
...
+ *
+ * @param onSwipeEnd Callback function invoked when an accessibility action is triggered, receiving
+ * the [SwipeDirection] corresponding to the triggered action (either [SwipeDirection.StartToEnd]
+ * or [SwipeDirection.EndToStart]).
+ * @return A list of [CustomAccessibilityAction] to be used by accessibility services through Compose
+ * semantics.
+ */
 @Composable
-private fun RowScope.SwipeableRowBackgroundContent(
+private fun rememberAccessibilityActions(
```
wmontwe, line 158: How to test that?

rafaeltonholo:
> I tested this on different screen sizes and it doesn't feel as polished as the current implementation. I played with the threshold value and based on the screen size the revealed content is easily cut. Or in landscape you need to swipe large distances to trigger the actions.

The "revealed content being cut" is because of how the behind element is drawn. I believe it is being cut off only in the "swipe to reveal", but the Swipe component shouldn't be responsible for how their slot composables are drawn.

Regarding the "you need to swipe large distances to trigger the actions", the threshold is based on a % of the actual width available to swipe. We could change to be an actual value, or both. How do you feel about this?

I will also add an input to let you play with the threshold directly from the ui-catalog

> When swiping the current implementation had a threshold and visual guide before the action is fired. Which is neat from my point of view.

I'm not sure what kind of visual guide you are talking about, so I will leave a few videos here:
| Swipe not complete |  Swipe completed | Swipe canceled |
| -- | -- | -- |
| <video src="https://github.com/user-attachments/assets/3e7739e3-c81b-41e3-9ec1-8bc2b54f3cc4" /> | <video src="https://github.com/user-attachments/assets/d2f4d694-62fd-4882-9751-5c809456f361" />| <video src="https://github.com/user-attachments/assets/c59ae5e4-4598-4631-969f-d2b15abebec8" /> |

Could you please point me to the visual guide you were referring to?


> I couldn't get "Swipe to Reveal with Auto reset" working for "Both directions". While other actions worked.

Have you tried to swipe it before swiping the others? Might be related to the next issue you mentioned. Here is a screen record of it working:

https://github.com/user-attachments/assets/3feb9f64-8ec4-4b95-900b-4be874b6c86b


> While testing the swipe functionality completely stopped working ~2~ -> multiple (updated) times and only after navigating away (Atoms) and back (Molecules) it started to work again. The logs contained nothing that would indicate what is failing. I also had the effect that some stopped working while other swipes where still functio

I was able to reproduce this issue. I will investigate the reason.

> When doing wild swipes with auto reset enabled and keep on swiping multiple times in different directions (especially on the version with both directions) it acts out, trying to animate to a closed state being interrupted by another swipe in the other direction and trying to keep up processing all the swipe gestures and eventually triggering a lot of end state events and inconsistent end states as not always the last swipe wins. Doing the same with our current version keeps working.

I will investigate that too, thanks for bringing this

core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeableRow.kt
rafaeltonholo: You need to turn on TalkBack, but I guess it isn't working as I expected. I will investigate that too.

rafaeltonholo: I forgot to set up the accessibility actions in the UI Catalog, but I will take care of that.

To test the functionality, please follow these steps:

1. Enable TalkBack.
2. Tap on the SwipeableRow.
3. Tap three times on the screen.
4. Select the "Actions" option and double-tap to open it.
5. Choose the action you want to test and double-tap to activate it.

dani-zilla: I always hate having to pull up the talkback actions menu, but I will say, this does work better for accessibility than our current implementation on the Message List, which doesn't expose the swipe actions to the user.

rafaeltonholo:
@wmontwe, I've made several updates to the animations and addressed issues related to accessibility, behaviour, and gestures that I identified.

Additionally, I've added a slider to the UI-Catalogue to allow for threshold adjustments.

I also wanted to give you a heads-up about the Saver implementation: During testing, I noticed that when swiping and rotating the screen, the state was saved but only for one orientation. There is a bug in the `ResponsiveContent` that is unrelated to these changes, and I will address it in a separate branch.

Please let me know if you want any other change.

core/ui/compose/designsystem/src/main/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeableRow.kt
dani-zilla: It does feel like a set size might be better than a screen percentage. Without it, landscape would have to swipe very far, and narrower screens might not reveal all the text for the action under the item until it's well past the threshold. Might be cleaner to do a set amount.

rafaeltonholo: > It does feel like a set size might be better than a screen percentage. Without it, landscape would have to swipe very far, and narrower screens might not reveal all the text for the action under the item until it's well past the threshold. Might be cleaner to do a set amount.

Yeah, I agree! The only problem that I had with setting up a fixed size was: How to define the correct threshold size that would fit on all screens? That was the reason I picked percentage instead.

With a percentage, we could set different percentages based on the `WindowSize`, which would automatically adjust for any screen size. Not sure it that actually make sense haha

dani-zilla: Oh, I was kind of assuming we could set the threshold on a per-item basis. Although one option could also be using a percentage dependent on the screen width/height ratio. But if the end goal is to make it always feel like you're swiping a set amount regardless of screen size or orientation, as well as make the item under the swipeable area reveal the same amount, a fixed value would do that better. 

The text under the swipeable row could be completely visible but the swipe threshold hasn't been reached yet, and it'll snap back. On the other hand, on narrower screens, perhaps the full text won't display before the swipe is registered if they let go.

rafaeltonholo: Here is what I was trying to explain with the threshold as a percentage:

### Phone portrait 

https://github.com/user-attachments/assets/356ba994-f0ea-4c8f-a753-f8df0f6d09f6

---

### Phone landscape 

https://github.com/user-attachments/assets/b84e2d34-6fd0-4b71-a22e-260c23180e4c

---

### Tablet landscape

https://github.com/user-attachments/assets/509d14ca-3fcb-4bb3-9454-16890d7a3927

---
 
### The behaviours creation:

```kotlin
@Composable
private fun SwipeAction.rememberSwipeBehaviour(): SwipeBehaviour {
    val windowInfo = getWindowSizeInfo()
    val threshold = when (windowInfo.screenWidthSizeClass) {
        WindowSizeClass.Compact -> 0.35f
        WindowSizeClass.Medium -> 0.25f
        else -> 0.15f
    }
    return remember(this) {
        when (this) {
            SwipeAction.None, SwipeAction.ArchiveDisabled -> SwipeBehaviour.Disabled

            SwipeAction.ToggleSelection,
            SwipeAction.ToggleRead,
            SwipeAction.ToggleStar,
            SwipeAction.ArchiveSetupArchiveFolder,
                ->
                SwipeBehaviour.Reveal(threshold = threshold, autoReset = true)

            SwipeAction.Archive,
            SwipeAction.Delete,
            SwipeAction.Spam,
            SwipeAction.Move,
                -> SwipeBehaviour.Dismiss(threshold = threshold)
        }
    }
}
```

---

Another good reason I see to use percentages over fixed sizes is that we could allow users to customize the swipe sensitivity easily.

Let me know what your thoughts about that are.

dani-zilla: But do we want them to be able to customize that? I assume the area revealed under the swipeable row is an interaction element that we'd want to control the padding, sizing, etc on. If the ratio's not just right for an item under the row, they may not know what an item is doing as they swipe on it. By not making it a ratio, we can make the swipe distance, and therefore the size of the item that displays under the row, constant. That ensures it's the right shape, the right length to reveal the text, and the same swipe distance regardless of screen size. 

swipeDistance

This kind of illustrates what I'm worried about. If the swipe distance isn't defined based on the size of what we want to reveal, we could have some situations that don't look good or hide the functionality. 

Besides that, the swipe distance itself could change between orientations. Even if we define different ratios for different orientations, we can't guarantee the same swipe distance each time. 

We could still keep this independent of the item below it, just allow the length of the swipe to be a hard size that the implementation can decide. 

Also, what if we eventually implement a short/far swipe thing? Some apps have different options depending on whether or not you swiped only far enough to reveal the first item or if you kept swiping to do something else. It could be mark read for a short swipe and reply for a full swipe, something like that.

And then there's the concern that we may have more than just swipeable items on the message list. Would the same ratio for the message list make sense elsewhere?

rafaeltonholo: I see your point, thanks for sharing the images, it helps a lot to understand better!!

> But do we want them to be able to customize that?

I don't see harm on allowing users to customize the triggering threshold. I have seen many complaints about the sensitivity of the swipe action and defining a sweet spot for that is really difficult. Letting then customize it would be helpful.

> I assume the area revealed under the swipeable row is an interaction element that we'd want to control the padding, sizing, etc on. If the ratio's not just right for an item under the row, they may not know what an item is doing as they swipe on it. By not making it a ratio, we can make the swipe distance, and therefore the size of the item that displays under the row, constant. That ensures it's the right shape, the right length to reveal the text, and the same swipe distance regardless of screen size. 


You are right! I guess the problem here is that I'm linking the swipe action trigger threshold with the reveal background element size, which will create that issue. I guess I need to have a different logic for swipe to dismiss, which the background will always be the size of the row, and the swipe to reveal, which will have a fixed background size.

The question is: should the swipe action trigger threshold be tied to the size of the background element or also a fixed size?

Let's say we set the swipe action trigger threshold to a fixed size, e.g. 100dp, but the reveal background element's size is 50dp. Then, users would need to swipe the whole background element's size to reveal it. Is that what we want?

With percentage, we could let the percentage be calculated with the background element's size, but that could create a sensitivity issue as well...


> Besides that, the swipe distance itself could change between orientations. Even if we define different ratios for different orientations, we can't guarantee the same swipe distance each time. 

I'm not sure if it is important to guarantee the same swipe distance, tbh. I think it is more important to guarantee how easy is to trigger the action rather than how far I need to swipe (same distance).
Additionally, we need to remember that as we use dp, a fixed size will vary depending on user's device resolution.


> We could still keep this independent of the item below it, just allow the length of the swipe to be a hard size that the implementation can decide. 

Answered above.


> Also, what if we eventually implement a short/far swipe thing? Some apps have different options depending on whether or not you swiped only far enough to reveal the first item or if you kept swiping to do something else. It could be mark read for a short swipe and reply for a full swipe, something like that.

That might be a good idea, but I would keep for the future. Currently we only support one swipe action per direction.

> And then there's the concern that we may have more than just swipeable items on the message list. Would the same ratio for the message list make sense elsewhere? 

I don't think so... I can't see a use case for that, tbh.

Thanks for all the inputs! Let me know your thoughts and let's keeping doing these kinds of discussion! I feel we will bring a better UX with this!

dani-zilla: > I'm not sure if it is important to guarantee the same swipe distance, tbh. I think it is more important to guarantee how easy is to trigger the action rather than how far I need to swipe (same distance).
Additionally, we need to remember that as we use dp, a fixed size will vary depending on user's device resolution.

More we could define the swipe distance to be limited by the size of the item that it's revealing. When we design the item below it, we'd set it to either a length that design says. Honestly, at the end of the day, I don't think this is an engineering decision anyway. This should be something the design team weighs in on. It might not be a bad idea to ensure the swipeable item could have a limit based on screen size _or_ a hard pixel limit

rafaeltonholo: @laurelterlesky, @Alecaddd, here is the thread! I asked GPT to create a TL;DR, but feel free to read all the comments:

**TL;DR:** 
The thread is debating whether swipe actions should use a **percentage-based threshold** or a **fixed/limited swipe distance**. 

The main concern with percentages is that the revealed action area can feel inconsistent across screen sizes and orientations, which may make the UI look less polished or hide functionality. 

Rafael agrees that the current approach is coupling the trigger threshold too closely to the revealed background size, and suggests **dismiss** and **reveal** may need different logic. 

The discussion ends with a soft consensus that this should probably be guided by **design**, with swipe distance constrained by the size of the revealed action — or at least capped — rather than using one ratio everywhere.

Alecaddd: Swipe actions by their own nature are a quick entry point to trigger repeated action, a convenient and more faster way to do things rather than accessing menus.
With that in mind, if we make the threshold based on %, it'll become very inconvenient to do long swipes when in landscape mode or tablet.

My recommendation is to define a fixed value for the threshold.
Based on my tests a comfortable distance on mobile portrait orientation is around 150px.
We shouldn't worry too much about needing to reveal the whole text of the action underneath because that can drastically change based on translations and it won't always meet our size expectations.

The important things are:
- When the threshold is met, the background color of the action  changes (as it already happens)
- If the user doesn't let go and drags back the swipe action nothing gets triggered. This allows users to drag as much as they want to read the text underneath without triggering anything.
- Acceleration is important as some users expect the swipe action to pass the threshold naturally if the swipe is quick.

**Bonus**
We might consider having different px distance for different resolutions, but that's for later and I would start with the simple and consistent implementation first.

rafaeltonholo: Thank you for sharing that! I've changed the `SwipeBehaviour` threshold to a fixed size instead of a percentage. The default threshold size is now `150.dp`, but you can customize it.

> Bonus
> We might consider having different px distance for different resolutions, but that's for later and I would start with the simple and consistent implementation first.

For this PR, we are focusing on the creation of design system components. The adjustments for different resolutions will be addressed in a separate branch. The last screen recording I shared demonstrated the current component already integrated, but it was intended just as an example.

PR #10725, opened 2026-03-23
Don't use deployments when they aren't needed

kewisch:
Avoiding deployments where not needed. I wasn't sure about shippable builds, I'm not sure we need deployments there at all since we don't use them, but since that is the once spot where we "deploy" something I left some of them as is. Please review where you need deployments.

https://github.blog/changelog/2026-03-19-github-actions-late-march-2026-updates/

### DIFF
--- a/.github/workflows/shippable_builds.yml
+++ b/.github/workflows/shippable_builds.yml
@@ -61,7 +61,9 @@ jobs:
     name: Show Release Environment
     runs-on: ubuntu-latest
     needs: get_environment
-    environment: ${{ needs.get_environment.outputs.releaseEnv }}
+    environment:
+      name: ${{ needs.get_environment.outputs.releaseEnv }}
+      deployment: false
     outputs:
       matrixInclude: ${{ steps.dump.outputs.matrixInclude }}
       releaseDate: ${{ steps.dump.outputs.releaseDate }}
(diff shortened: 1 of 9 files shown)


.github/workflows/shippable_builds.yml
```diff
@@ -623,7 +631,9 @@ jobs:
     name: Wait for Approval
     needs: [ dump_config, sign_mobile ]
     if: "${{ contains(fromJSON('[\"beta\", \"release\"]'), needs.dump_config.outputs.releaseType) }}"
-    environment: publish_hold
+    environment:
+      name: publish_hold
+      deployment: false
```
coreycb, line 636: I think this might need to be true since it has a deployment protection rule.

kewisch: From the docs at https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/control-deployments#using-environments-without-deployments

> When deployment is set to false:
> Required reviewers still apply—reviewers must still approve before the job runs.

coreycb: Thanks for the link. This will work as-is then. I guess the next question is do we want to have an audit record of deployments for publish hold or not. I think we might want to retain them for security purposes.

PR #10751, opened 2026-03-26
fix(message-reader): adds font scaling to match user preferences

dani-zilla:
Fixes: https://github.com/thunderbird/thunderbird-android/issues/10692 

Before:

Screenshot 2026-03-26 at 5 39 50 PM

Notice the font size in the information header, which is based on the system font size, and how the text below it is unchanged compared to...

After: 
tinyToHugeFontSizing

Where the font scales with the user's settings.

Users noticed that the font size in messages did not scale with the device preferences. This is fixed by getting the current system fontScale and multiplying it by the incoming font size. This does not alter the display if the user has chosen to force messages to fit in the window by shrinking them, that option will still be respected. If they turn this off, font will scale with the system font.


mikhailzolikoff:
I presume it'll take "a minute" for this to be deployed to the Play Store?

PR #10757, opened 2026-03-30
chore(ci): add feature flag report flag and update script

wmontwe:
Add flag to signal user-facing changes that are guarded by a feature-flag.

### DIFF
--- a/.github/workflows/pr-request-report-labels.yml
+++ b/.github/workflows/pr-request-report-labels.yml
@@ -33,16 +33,17 @@ jobs:
 
           include_count="$(jq '[.[] | select(.name == "report: include")] | length' <<< "$labels_json")"
           exclude_count="$(jq '[.[] | select(.name == "report: exclude")] | length' <<< "$labels_json")"
+          feature_flag_count="$(jq '[.[] | select(.name == "report: feature-flag")] | length' <<< "$labels_json")"
           highlight_count="$(jq '[.[] | select(.name == "report: highlight")] | length' <<< "$labels_json")"
 
-          total_count=$((include_count + exclude_count + highlight_count))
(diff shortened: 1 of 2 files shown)


rafaeltonholo:
I was considering the possibility of allowing the `feature flag` PR label to have a broader application rather than just being a report-only flag.

Additionally, we could enforce that whenever the `feature flag` PR label is used, the PR body or a comment must explicitly state the name of the feature flag that the PR corresponds to. 

This would enable us to maintain a report of features that are currently in progress but not yet available to users, while also providing them with a heads-up about upcoming features.

What are your feelings about this idea?

wmontwe:
I think a comment with the actual feature flag name makes total sense. I would need to check if that could be extracted easily.

The dedicated flag might make sense, but then it can't be enforced. It will be an additional thing you need to do and remember to do. You would have 3 steps as a result: add the report label, add the feature label and comment the name of the feature flag. If that's not too complicated we could do that instead.

@dani-zilla What do you think?

dani-zilla:
Yeah, adding the feature flag label and commenting what the flag is to use it sounds reasonable. Makes it easy to see what flag it is at a glance and, when the feature is hidden by a flag that's introduced in a previous PR, you'll still be able to see what it relates to

rafaeltonholo:
We can use the following command with `gh` to retrieve the feature-flag from the PR description:
```
gh pr view "$PR_NUMBER" \
  --repo thunderbird/thunderbird-android \
  --json body \
  --jq 'try (.body | gsub("\r"; "") | capture("(?m)^feature-flag:\\s*`(?<flag>[^`]+)`$").flag) catch ""'
```

Tested with #10785 and got: `enable_message_list_new_state`

wmontwe:
Closing this in favor of adding a `feature-flag` label along with a comment stating the name of the feature-flag.

PR #10771, opened 2026-04-02
fix(deps): downgrade androidxSwiperefreshlayout from 1.2.0 to 1.1.0

### LINKED ISSUE #10243: Deleted message remains visible in message list
### Checklist

### App

Thunderbird for Android

### App version

16.0b2

Other

### Android version

15

### Device model

Xiaomi 14

### Steps to reproduce

show message list
click on top message to show it
delete the message by click on the trashbin -> returns to message list

### Expected behavior

deleted message should disappear from list

### Actual behavior

the deleted message remains visible in the message list until the next refresh or sync. The message looks as if it had not been read, but it cannot be selected again. Actually, it has been correctly moved to trash.

This happens only if the concerned message is th top one in the list.

https://github.com/user-attachments/assets/30322137-5529-4f86-ad32-d7619cf57ae1

### Logs

_No response_

rafaeltonholo:
Fixes #10243.

The upgrade to 1.2.0 caused the Message List RecyclerView to get the stale state again.

As we are rewriting the Message List on Compose, the `androidx.swiperefreshlayout:swiperefreshlayout` won't be necessary in the future.

### DIFF
--- a/app-k9mail/dependencies/fossReleaseRuntimeClasspath.txt
+++ b/app-k9mail/dependencies/fossReleaseRuntimeClasspath.txt
@@ -157,7 +157,7 @@ androidx.sqlite:sqlite-framework-android:2.5.0
 androidx.sqlite:sqlite-framework:2.5.0
 androidx.sqlite:sqlite:2.5.0
 androidx.startup:startup-runtime:1.2.0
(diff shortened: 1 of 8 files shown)


wmontwe (approved):
LGTM

PR #9652, opened 2025-08-26
Show avatars on message notifications

Goooler:
Closes #7089.

PR #10786, opened 2026-04-03
fix(notifications): respect showContactPicture setting for notification avatars

### LINKED ISSUE #10750: New Mail notification: disable avatar
### Checklist

### App

Thunderbird for Android

### App version

17.0

F-Droid

### Android version

16

### Device model

Galaxy S26 Ultra

### Steps to reproduce

With Thunderbird 17, I have now the Avatar or contact picture in the notification. I see that this was a feature request and maybe it's useful for some. On the other hand, I don't use Avatar / contact pictures and that's why I disabled them in the list view.

Is it possible that also the Avatar in the notification can be disabled, when I disable them in the mail overview? Either as separate option or dependent if I disable them in the overview?

### Expected behavior

No Avatar/  contact picture in the notification. 

### Actual behavior

Avatar is visible, when I disable them in the list overview 

### Logs

_No response_

mvanhorn:
## Summary

When contact pictures are disabled in the message list settings, notification avatars are still shown. This checks the `isShowContactPicture` preference before loading contact avatars for new mail notifications.

## Why this matters

PR #9652 added avatar support to notifications but didn't check the existing contact picture visibility setting. Users who disable contact pictures in the message list expect notifications to also respect that preference.

## Changes

- `SingleMessageNotificationCreator.kt`: Added `MessageListPreferencesManager` as a constructor dependency and an early return in `setAvatar()` when `isShowContactPicture` is false (line 68)
- `CoreNotificationKoinModule.kt`: Injected `messageListPreferencesManager` into `SingleMessageNotificationCreator`

## Testing

- Verified the preference check reads from the same `showContactPicture` key used by the message list UI (`MessageViewHolder.kt:63`)
- When `isShowContactPicture` is false, `setAvatar()` returns without loading or setting the large icon
- When `isShowContactPicture` is true (default), behavior is unchanged

Fixes #10750

This contribution was developed with AI assistance (Claude Code).


rafaeltonholo (changes requested):
Hi @mvanhorn,

Thank you for taking the time to work on this PR.

While your current change allows users to enable or disable the Avatar in notifications, the configuration you're using is incorrect. 

The `MessageListPreferencesManager` is meant for the Message List screen, not for Notifications. There may be instances where users prefer not to have Avatars displayed in notifications but still want to see them in the Message List, and vice versa.

To properly address issue #10750, we should add a new user preference under `NotificationPreference`. This would allow users to disable Avatars specifically for notifications.

For reference, you can look at commit 8dfa32ea8800af7db64f786b69c894f39d2898b1, which demonstrates how to add a new Boolean preference using the `NotificationPreferenceManager`. 

Thank you!

mvanhorn:
Thanks for the review, @rafaeltonholo. Reworked this in d847439f to use a dedicated `NotificationPreference` value instead of the message-list contact picture preference.

Changes:
- Added `isShowContactPictureInNotification` to `NotificationPreference` (defaults to `true` to preserve current behavior).
- Persisted via `KEY_SHOW_CONTACT_PICTURE_IN_NOTIFICATION` in `DefaultNotificationPreferenceManager`.
- `SingleMessageNotificationCreator.setAvatar()` now reads from `NotificationPreferenceManager` instead of `MessageListPreferencesManager`.
- Added a separate "Show contact pictures in notifications" toggle in general settings, alongside the existing notification preferences.
- Pattern mirrors commit 8dfa32ea8800af7db64f786b69c894f39d2898b1 you referenced.
- Added focused unit coverage (`SingleMessageNotificationCreatorTest`) for both branches of the new preference.

Force-pushed to replace the original commit since the underlying approach changed.

rafaeltonholo (approved):
LGTM, please, just update your branch with the latest main version, so we can get rid of the CI failure.

mvanhorn:
Thanks for the approval @rafaeltonholo. Rebased onto current main (172694bbeb on top of fb1baad114) and force-pushed. The diff is unchanged.

The `require-report-label` check is gated on a `report: include`/`report: exclude`/`report: highlight` label that I can't add as a non-maintainer; if you can drop one of those on the PR I think the rest of the gate clears.

rafaeltonholo:
> Thanks for the approval @rafaeltonholo. Rebased onto current main ([172694b](https://github.com/thunderbird/thunderbird-android/commit/172694bbeb889e3375ad25f3228a5d8be98adb13) on top of [fb1baad](https://github.com/thunderbird/thunderbird-android/commit/fb1baad114168cb49873adc8f6165df1ce55210a)) and force-pushed. The diff is unchanged.
> 
> The `require-report-label` check is gated on a `report: include`/`report: exclude`/`report: highlight` label that I can't add as a non-maintainer; if you can drop one of those on the PR I think the rest of the gate clears.

I've placed the label already; you should be good now. Once the CI finishes, we can merge these changes! Thanks for this contribution!

PR #10796, opened 2026-04-07
fix(message-view): Respect animation setting when displaying message body

### LINKED ISSUE #1903: Fade-in animation when opening message plays even when animations are disabled
### Expected behavior
When disabling the "Global settings > Display > Animation" setting, when opening a mail, no animation should play.

### Actual behavior
Opening a mail results in a fade-in animation, with the message contents appearing progressively (the duration is around 0.5s-1s).

### Steps to reproduce
1. Disable the setting "Global settings > Display > Animation"
2. Open a message
3. Observe a fade-in animation

### Environment
K-9 Mail version: 5.200

Android version: 6.0.1

Account type (IMAP, POP3, WebDAV/Exchange): IMAP, if it matters

Other users have complained about this behavior on a mailing-list: https://www.mail-archive.com/k-9-mail@googlegroups.com/msg03953.html

Thanks for your help!

untitaker:
**Description:**

Disable fade-in animation when opening messages if the user has disabled animations in Thunderbird's own app settings.

The fix was written with AI, but testing, code review and commit messages were done manually.

Fixes #1903

**Testing:**

1. Go to Settings > General > Display > Use gaudy... and disable it
2. Open any email message
3. Verify the message content appears without fade-in

In most cases the message view still loads in steps and feels jumpy, but without any animation.

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessageTopView.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessageTopView.kt
@@ -73,6 +73,10 @@ class MessageTopView(
         layoutInflater = LayoutInflater.from(context)
 
         viewAnimator = findViewById(R.id.message_layout_animator)
+        if (!visualSettingsPrefManager.getConfig().isShowAnimations) {
+            viewAnimator.inAnimation = null
+            viewAnimator.outAnimation = null
+        }
         progressBar = findViewById(R.id.message_progress)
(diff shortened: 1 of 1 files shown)


untitaker:
`ViewSwitcher.java` works different than this fix: It sets up animations explicitly in code only if the animation setting is enabled, while this view sets up animations in `message.xml` and here I added code to conditionally remove them again. Just went for the most minimal fix, but happy to do a broader refactor so animation settings are applied the same way across views.

rafaeltonholo (approved):
LGTM! Thank you for contributing to our project!

PR #10800, opened 2026-04-07
feat(settings): animation setting inherits systemwide settings

### LINKED ISSUE #6027: The default value for Animations is not following a system-wide accessibility setting
### Checklist

### App version

5.807

Google Play

### Android version

12

### Device model

Pixel 4a

### Steps to reproduce

1. Go to the accessibility settings, and enable “Remove animations” under “Text and Display” section. In previous versions of Android, this option may have been hidden in the Developer Options menu.
2. Install K-9.
3. Go to the application settings, display.

### Expected behavior

The animations should be disabled by default if the system-wide setting disables the animations.

### Actual behavior

The animations are enabled by default unconditionally.

### Logs

N/A

untitaker:
Convert the "Use gaudy visual effects" boolean checkbox to a tri-state option (On/Off/Use system default) that defaults to following Android's system animation accessibility setting.

New installations follow system settings, old installations inherit their on/off state.

Fixes #6027

Note that this only changes the default to respect system settings. Unlike https://github.com/thunderbird/thunderbird-android/pull/10658 this does not touch any views that previously didn't respect the application-level setting. Any view or UI component that didn't respect the application-level setting will continue to not respect any animation settings, and should be fixed in a separate PR IMO.

**Implementation:**

Similar to ThemeManager, introduce a new AnimationManager and funnel all access to this option through there. I'm on the fence as to whether that isn't too much overkill, IMO it would make sense to merge them both into one "settings manager".

Right now there is only one place where the setting is being read, in ViewSwitcher. However, there are many other places where the setting is supposed to be respected but isn't (sidebar, messagetopview #10796) so I figured it's better to centralize the resolution outside of ViewSwitcher.

Written with AI-assistance, however like #10796 testing and commit messages are written manually.

Screenshot_20260407-171904
Screenshot_20260407-171908

### DIFF
--- a/legacy/core/src/main/java/com/fsck/k9/preferences/upgrader/GeneralSettingsUpgraderTo111.java
+++ b/legacy/core/src/main/java/com/fsck/k9/preferences/upgrader/GeneralSettingsUpgraderTo111.java
@@ -1,0 +1,25 @@
+package com.fsck.k9.preferences.upgrader;
+
+
+import java.util.Map;
+
+import com.fsck.k9.preferences.SettingsUpgrader;
+import net.thunderbird.core.preference.AnimationPreference;
+
+
+/**
+ * Convert <em>animations</em> from boolean to {@link AnimationPreference} enum.
+ *
+ * {@code true} maps to {@link AnimationPreference#ON}, {@code false} maps to {@link AnimationPreference#OFF}.
+ */
+public class GeneralSettingsUpgraderTo111 implements SettingsUpgrader {
+
+    @Override
+    public void upgrade(Map<String, Object> settings) {
+        Object animations = settings.get("animations");
+        if (animations instanceof Boolean) {
+            boolean value = (Boolean) animations;
+            settings.put("animations", value ? AnimationPreference.ON : AnimationPreference.OFF);
+        }
+    }
+}
--- a/legacy/storage/src/main/java/com/fsck/k9/preferences/migration/StorageMigrationTo30.kt
+++ b/legacy/storage/src/main/java/com/fsck/k9/preferences/migration/StorageMigrationTo30.kt
@@ -1,0 +1,26 @@
+package com.fsck.k9.preferences.migration
+
+import android.database.sqlite.SQLiteDatabase
+
+private const val ANIMATION_KEY = "animations"
+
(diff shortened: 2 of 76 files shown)


untitaker:
ack on the lint issues, will fix this week

legacy/core/src/main/java/com/fsck/k9/preferences/upgrader/GeneralSettingsUpgraderTo111.java
wmontwe, line 1: Upgraders must be covered by a test.

legacy/ui/legacy/src/main/res/values/strings.xml
```diff
@@ -701,6 +701,9 @@
 
     <string name="animations_title">Animation</string>
     <string name="animations_summary">Use gaudy visual effects</string>
```
wmontwe: The "animations_summary" string could be removed.

legacy/storage/src/main/java/com/fsck/k9/preferences/migration/StorageMigrationTo30.kt
wmontwe, line 1: Migrations must be covered by a test.

wmontwe (changes requested):
Thank you for taking care of this issue.

Just add a test for the migration/updater please and remove the `animations_summary` string.

untitaker:
@wmontwe this PR is ready for review

@dani-zilla since you commented on https://github.com/thunderbird/thunderbird-android/pull/10658 maybe this is of interest to you too

after this PR is through I intend to fix the sidebar to respect the application setting.

wmontwe:
@untitaker I updated this PR to a mergable state. The unused string resource is now gone.

We need to merge #11123 first and then apply the string removal.

untitaker:
thanks, didn't see that CI failed

wmontwe:
#11123 got merged.

This is ready to be reviewed.

wmontwe (approved):
Thank you for the contribution!

@rafaeltonholo As I changed the code I can't approve it now.

rafaeltonholo (approved):
LGTM

Issue #10775, opened 2026-04-02
[ML 0.1/UI.04.2] - Implement legacy message list loading bridge


rafaeltonholo:
## Blocked by

- #10774

## Description
Implement the bridge between the legacy message loading system and the Compose UI. This includes creating the `LegacyMessageListBridge` interface, the `LoadMessagesLegacySideEffect` that invokes the bridge to load messages from the legacy data source, and the mapping from `MessageListItem` to `MessageItemUi`.

## Acceptance Criteria
- `LegacyMessageListBridge` interface defined in the API module.
- `LoadMessagesLegacySideEffect` loads messages from legacy and maps them to `MessageItemUi`.
- `MessageItemUi` includes `messageReference` for identifying messages.
- `MessageListFragment` implements the bridge, converting `MessageListItem` -> `MessageItemUi`.
- Manual refresh on sort criteria change is removed (handled by bridge).
- Action mode finish is guarded to only trigger when inactive and no messages are selected.
- New state feature is enabled alongside Compose for message list items.
- Unit tests for `LoadMessagesLegacySideEffect`.

PR #10805, opened 2026-04-08
chore(state-machine): add support for dispacthing ui side effects from state side effects

rafaeltonholo:
Part of #10775.
feature-flag: `enable_message_list_new_state`

- Update `StateSideEffectHandler` and its `Factory` to include a third generic type `TEffect`.
- Add `dispatchUiEffect` to `StateSideEffectHandler` to allow side effect handlers to trigger UI effects.
- Update `BaseStateMachineViewModel` and associated tests to support the updated factory and handler signatures.


wmontwe (approved):
LGTM

Issue #9709, opened 2025-09-02
Font size controls removed, now message list font size is too big and cannot be reduced


rrrrrraa:
### Checklist

- [x] I have used the search function to see if someone else has already submitted the same bug report.
- [x] I will describe the problem with as much detail as possible.

### App

Thunderbird for Android

### App version

11.1

### Where did you get the app from?

Google Play

### Android version

16

### Device model

Pixel 6

### Steps to reproduce

In Thunderbird : Go to settings, display, global ( font size option is no longer there )

In Android system settings : go do display, Screen, Display and Text size
Confirm Text size is set to the smallest possible size.

### Expected behavior

Using the smallest possible System text size, the message list font size should be smaller than in this screenshot from Thunderbird version 10, with in-app font size set to "Very small" (beyond which there is still "Smallest")

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/3abeee81-dda4-4092-9540-cb32dbd9cc6b" />

### Actual behavior

Message list font size is actually much larger than before, therefore fewer messages and less of the message previews are displayed, with the same Density and Preview lines settings as before.

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/c2ba7f8f-936c-4d28-a86c-03121a472e2f" />

rrrrrraa:
Thanks for your work on Thunderbird. I am a long time Thunderbird/K9 mail user, but I'm unhappy with the effect of the removal of the font size options because the fonts are all too big now even with the System text set to minimum size. It is less efficient to view messages, and I personally find it unpleasant to look at these large fonts.

Related issues : 
https://github.com/thunderbird/thunderbird-android/issues/8949
https://github.com/thunderbird/thunderbird-android/issues/9242

wmontwe:
We’ve removed these settings as part of the ongoing UI redesign, since the previous way of changing font sizes no longer works with the new design system. We’re working on alternative options to bring back some customization in a future update.

I added this to our epic for the message list rewrite #8838 for consideration.

rafnov:
More or less - when can we expect proper font handling?
Weeks? Months? Years?
I'm using K-9 for a few days, I wouldn't like to have to used to it to much.

mcepl:
[...]

PR #10809, opened 2026-04-08
fix: 'Font Size' setting appears in search when feature flag is off

### LINKED ISSUE #10756: The font-changing feature is not displaying as expected
### Checklist

### App

Thunderbird for Android

### App version

17

Other

### Android version

hyper os 2.0.220.0

### Device model

k60

### Steps to reproduce

I opened the general settings as usual and tried to find the font size option, but it wasn't there.

### Expected behavior

It should display properly, and you can use the search function to navigate to the font size page.

### Actual behavior

In Thunderbird 17, there isn't an option to change the font size when I try to do so; I have to search for the feature to access the font size settings. I hope this issue can be fixed.

### Logs

[k9mail-logs.txt](https://github.com/user-attachments/files/26332677/k9mail-logs.txt)

shamim-emon:
- Fixes #10756 
## Description
This PR ensures that all Font Size related preferences are hidden from the search results when the corresponding feature flag is disabled.

Previously, even when the Font Size settings screen was hidden via the feature flag, the search system would still show those preferences because they were already indexed. This caused confusion and inconsistent behavior.

## Implementation Highlights:
Updated `GeneralSettingsActivity `to dynamically ignore these keys in the search index when the `disable_font_size_config `feature flag is enabled using `ignorePreference(key)`.

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/general/GeneralSettingsActivity.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/settings/general/GeneralSettingsActivity.kt
@@ -16,12 +16,35 @@ import com.fsck.k9.ui.R
 import com.fsck.k9.ui.base.BaseActivity
 import com.fsck.k9.ui.base.extensions.fragmentTransaction
 import com.fsck.k9.ui.base.extensions.fragmentTransactionWithBackStack
+import kotlin.getValue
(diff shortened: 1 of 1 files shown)


wmontwe (approved):
LGTM

lgs11:
:-1: Boo. The only workaround for #9709 is now gone. I will need to find an old version of K-9 that still works.

PR #10848, opened 2026-04-10
fix(i18n): locale-sensitive ordering of folders (#10718)

### LINKED ISSUE #10718: [Bug] Wrong alphabetical order of sub-folders in e-mail folder list (due to accents, like ä, ö, ü)
### Checklist

### App

Thunderbird for Android

### App version

16.1

F-Droid

### Android version

GrapheneOS Android 16 Build-Number 2026030701

### Device model

Google Pixel 6a

### Steps to reproduce

1. open Thunderbird for Android
2. open sub-folders in e-mail folder list within main menu
3. accented strings like ä, ö, ü at the bottom of the list — not within the proper sorting

### Expected behavior

1. sorting like a<ä<b, o<ö<p, u<ü<v

### Actual behavior

accented strings like ä, ö, ü at the bottom of the list — not within the proper sorting

### Logs

Image

This bug is very similar to https://github.com/signalapp/Signal-Android/issues/14487, even when that is a totally different app — but the behaviour is almost just the same!

winne42:
Closes #10718

**Before:**
* folders with accented starting letters are moved after the end of the regular alphabet
* lower-case letters are treated as being equal to upper-case --> order is sometimes `lower-case < upper-case`, sometimes `upper-case < lower-case`
image

**After:**
* folders with accented characters are integrated in regular alphabet
* always `lower-case < upper-case`
image

Note that changes to the devices locale may require a restart of Thunderbird to take effect. If this was not wanted, `sortForDisplay` _could_ be converted from a stored to a computed property.

### DIFF
--- a/legacy/ui/folder/src/main/java/app/k9mail/legacy/ui/folder/DefaultDisplayFolderRepository.kt
+++ b/legacy/ui/folder/src/main/java/app/k9mail/legacy/ui/folder/DefaultDisplayFolderRepository.kt
@@ -5,6 +5,7 @@ import app.k9mail.legacy.mailstore.FolderTypeMapper
 import app.k9mail.legacy.mailstore.MessageStoreManager
 import app.k9mail.legacy.message.controller.MessagingControllerRegistry
 import app.k9mail.legacy.message.controller.SimpleMessagingListener
+import java.text.Collator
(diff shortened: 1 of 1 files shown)


wmontwe (approved):
Thank you!

PR #10862, opened 2026-04-12
Restore compose button functionality

jondaley:
fix: this PR does two things: 1, fixes compose button on overflow menu when viewing a message.  2. displays compose button on top menu when using non-floating preference, rather then putting it on the overflow menu.

The compose button behavior was recently changed.  It seem like maybe the developers are using the floating compose button so haven't noticed that the behavior broke.  I would guess that because of that, the compose menu choice isn't needed, but for those of us who prefer the non-floating version, this fixes a needed behavior.

The bug was reported here: https://github.com/thunderbird/thunderbird-android/issues/10753


wmontwe:
@jondaley Thanks for the fix. You just need to run `./gradlew spotlessApply` to resolve the issue and test with `./gradlew spotlessCheck`.

jondaley:
Great, thanks.  I wasn't sure what was missing.  I believe it is all set now?

wmontwe (approved):
This fixed the issue, thank you.

PR #10878, opened 2026-04-15
refactor: use PreferenceDataStore for PostMarkAsUnreadNavigation

### LINKED ISSUE #10754: Replace Direct Usage of K9.messageViewPostMarkAsUnreadNavigation With PreferenceDataStore
The navigation behavior after marking a message as unread is currently controlled via the static accessor: `K9.messageViewPostMarkAsUnreadNavigation` . This approach relies on legacy global state and bypasses the modern preference management architecture. As a result:
- It introduces tight coupling to the K9 class
-  It is not lifecycle-aware
- It makes testing more difficult due to the lack of dependency injection
- It does not align with the current DataStore-based preference system

## 🎯 Goal 
Replace the static accessor` K9.messageViewPostMarkAsUnreadNavigation` with a DataStore-backed preference, and update all usages across the application to access this value through the modern preference infrastructure

mohitsatr:
Resolves #10754 

This Pr replaces the static accessor **K9.messageViewPostMarkAsUnreadNavigation** with a DataStore-backed preference in **InteractionSettingsPreferenceManager** and updates the logic to read/write from the new preference source


wmontwe (approved):
Thank you!

PR #10893, opened 2026-04-18
fix: Restoring backup from file doesn't respect 'swipe actions' setting

### LINKED ISSUE #10621: [k-9 mail] restoring backup from file doesn't respect 'swipe actions' setting
### Checklist

### App

K-9 Mail

### App version

16.1

F-Droid

### Android version

11 (R)

### Device model

Redmi note 4 pro (mido)

### Steps to reproduce

1. configure `settings > interaction > swipe actions` to be something else than the default values
2. make a backup from `settings > backup > export settings`
3. fresh install k9 mail
4. restore the previously made backup, di the necessary accounts setup afterwards
5. after setting up the account, you will see that the values of `swipe actions` setting won't be the same as you configured them to be, but rather they will be reset back to their default values (every other _general setting_ should be restored properly)

### Expected behavior

The values of `swipe actions` setting should be restored correctly just like the other settings are.

### Actual behavior

The values of `swipe actions` setting get reset back to their default values despite of our exported configuration 

### Logs

_No response_

shamim-emon:
- fixes #10621 
### 🐛 Problem
 - While restoring backup from files , `GeneralSettingsWriter` uses legacy preference system to store GeneralSettings values. 
 -  New PreferenceManagers for GeneralSettings  don't listen to this update and so GeneralSettings flow does not react to this change and UI does not get update without restarting the app.
### 🛠️ Solution
- Made all preferenceManagers subscribe to `PreferenceChangeBroker`.
- Notify all preferenceManagers upon GeneralSetting restore on `GeneralSettingsWriter`.

### DIFF
--- a/core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/PreferenceChangePublisher.kt
+++ b/core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/PreferenceChangePublisher.kt
@@ -2,11 +2,15 @@ package net.thunderbird.core.preference
 
 /**
  * Publishes changes of preferences to all subscribers.
+ * The change can be scoped to a specific category of preferences or applied globally.
+ * @see PreferenceScope for available scopes of preference changes.
  */
 interface PreferenceChangePublisher {
 
     /**
      * Publish a change in the preferences.
+     * @param scope Defines which category of preferences has changed.
+     * By default, [PreferenceScope.ALL] is used, indicating a global change affecting all preferences.
      */
-    fun publish()
+    fun publish(scope: PreferenceScope = PreferenceScope.ALL)
 }
--- a/core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/PreferenceChangeSubscriber.kt
+++ b/core/preference/api/src/commonMain/kotlin/net/thunderbird/core/preference/PreferenceChangeSubscriber.kt
@@ -6,7 +6,9 @@ package net.thunderbird.core.preference
 fun interface PreferenceChangeSubscriber {
 
     /**
-     * Called when preferences change.
+     * Called when preferences within a given scope have changed.
+     * @param scope The [PreferenceScope] indicating which category of preferences
+     * has been updated.
      */
-    fun receive()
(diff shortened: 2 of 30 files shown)


wmontwe (commented):
Using the `PreferenceChangeBroker` solved the issue, but it causes side effects.

This is using a simple broadcaster that notifies all subscribers and therefore could lead to a broadcast storm as now every settings related manager reacts to any change. Just add a log to every receive and then change the "animation" setting you will see 11 managers updating.

While the individual reloads are relatively small, they quickly add up causing simultaneous updates across multiple managers that end up aggregated as `GeneralSettings` with multiple updates to consumers in the app.

Some ideas:

These are rough ideas and maybe not doable or adding to much complexity.

1. Keep the change broker, but scope the emissions (adds complexity)

```kotlin
enum class PreferenceScope { ALL, DISPLAY, PRIVACY, NETWORK, DEBUGGING }

interface PreferenceChangePublisher {
    fun publish(scope: PreferenceScope = PreferenceScope.ALL)
}

fun interface PreferenceChangeSubscriber {
    fun receive(scope: PreferenceScope)
}

// In a manager:
override fun receive(scope: PreferenceScope) {
    if (scope == PreferenceScope.ALL || scope == PreferenceScope.PRIVACY) {
        configState.update { loadConfig() }
    }
}
```

2. Use key-based notifications

```kotlin
fun interface PreferenceChangeSubscriber {
    fun receive(changedKey: String)
}

// In a manager:
override fun receive(changedKey: String) {
    if (myOwnedKeys.contains(changedKey)) {
        configState.update { loadConfig() }
    }
}
```

3. Use a reactive storage

```kotlin
class DefaultNotificationPreferenceManager(...) {
    val configFlow = combine(
        storage.observeBoolean(KEY_QUIET_TIME),
        storage.observeString(KEY_NOTIFICATION_VIBRATE)
    ) { quietTime, vibrate -> 
        NotificationSettings(quietTime, vibrate) 
    }.stateIn(scope, Eagerly, initialValue)
}
```

But our storage is not reactive yet.

---

Other ideas are welcome

wmontwe (approved):
Thank you! 

And sorry for the long wait.

PR #10920, opened 2026-04-21
Fixed cutoff behavior for notification actions

harshad1:
This PR is a follow up bugfix for https://github.com/thunderbird/thunderbird-android/pull/10301

In this PR, we fix the cutoff behavior for the notification action settings

Prior to this we always showed 3 actions. With this change, we only show actions above the cutoff (up to 3) matching the design intent of the system


harshad1:
This PR probably ought to go into the 19 release

dani-zilla (approved):
Looks good and in my testing, it fixed the issue. Thanks for the quick turnaround on this bug fix!

PR #10940, opened 2026-04-23
chore(weblate-cli): add weblate create and delete operations

### LINKED ISSUE #10865: Weblate - setup new project
The existing Weblate project was built specifically for the Android version, leading to the use of Android identifiers in various links and component names. Furthermore, the legacy code is used as the base for all translations, which will cause issues during future migrations. To ensure consistent string propagation across different repositories (Android, iOS, Bolt, Mobile Components, etc.), we need to create a new Weblate project that utilizes categories to separate these different projects but still befits from shared string propagation.

wmontwe:
Resolves #10865 

This restructures the Weblate CLI to a subcommand pattern and adds create and delete commands for components. Updates the Weblate links and updates the translation CLI.

Create searches the project for all modules that contain Android or Compose resources and checks if a corresponding component exists within Weblate. If not, it prompts to create the component.

The configuration is updated to point to the new Weblate project: https://hosted.weblate.org/projects/thunderbird/thunderbird-android/.

The managed components config is updated to include all newly created components.


cli/weblate-cli/src/main/kotlin/net/thunderbird/cli/weblate/api/Component.kt
```diff
@@ -40,7 +40,19 @@ data class Component(
             encoder: Encoder,
             value: Component,
         ) {
-            error("Component serialization is not supported")
+            require(encoder is kotlinx.serialization.json.JsonEncoder) {
```
rafaeltonholo: nit: unnecessary full qualifier

cli/weblate-cli/src/main/kotlin/net/thunderbird/cli/weblate/api/WeblateClient.kt
```diff
@@ -8,91 +8,145 @@ import io.ktor.client.plugins.logging.DEFAULT
...
             components.addAll(componentPage.results)
 
             hasNextPage = componentPage.next != null
             page++
         }
 
+        if (config.cacheEnabled) {
+            try {
+                writeComponentsToCache(components)
+            } catch (_: Exception) {
```
rafaeltonholo: Shouldn't we at least log the exception?

rafaeltonholo (dismissed):
LGTM. Just two minor comments, no need to address if you don't want.

rafaeltonholo (approved):
LGTM

PR #10947, opened 2026-04-26
chore(ci): Ensure exactly 3 milestones exist

coreycb:
Without this check, pull requests could be put in the wrong milestone. For example, if we accidentally have only 2 milestones, one for beta and one for release, pull requests for main would incorrectly be put in the beta milestone.

### DIFF
--- a/.github/workflows/pr-merged.yml
+++ b/.github/workflows/pr-merged.yml
@@ -40,10 +40,16 @@ jobs:
             release) milestone_index=2 ;;
           esac
           echo "$milestone_index"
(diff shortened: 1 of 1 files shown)


stephenking314 (approved):
LGTM

coreycb:
I'm going to briefly close 1 milestone, try to merge (should fail), then go back to 3 milestones and merge again.

coreycb:
Well this runs post-merge so the check fails but it's not super obvious. It's still useful but I'll need to add something else to make it more obvious.

PR #10949, opened 2026-04-27
Add zizmor for gh action security analysis

coreycb:
Also fix the following zizmor issues:
- Add persist-credentials: false to all actions/checkout steps (artipacked)
- Fix template injection in pr-request-report-labels.yml by moving pull_request context values to env variables instead of inline expressions
- Replace secrets: inherit in build-daily.yml with explicit secret passing, and declare secrets in shippable_builds.yml workflow_call
- Replace softprops/action-gh-release with gh release create in shippable_builds.yml (superfluous-actions)
- Fix spoofable bot actor check in pr-dependabot-dependency-guard-update.yml
- Fix various issues to tighten up pull_request_target workflows

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -34,6 +34,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 1
+          persist-credentials: false
 
       - name: Show disk usage
         uses: ./.github/actions/disk-usage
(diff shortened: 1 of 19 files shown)


dandarnell:
I'm not 100% of the caveats, but setting persist-credentials to false breaks checkouts when using SSH. Setting it on comm-l10n-unified broke its mirror action, and I suspect it might break actions here as well.

coreycb:
> I'm not 100% of the caveats, but setting persist-credentials to false breaks checkouts when using SSH. Setting it on comm-l10n-unified broke its mirror action, and I suspect it might break actions here as well.

Yeah these changes are almost guaranteed to break something. Maybe they should be phased in to limit the impact. I can split into different patches.

PR #10953, opened 2026-04-27
fix(ci): ensure lables and pr body are properly escaped

wmontwe:
Fix `pr-request-report-labels.yml` breaking on nexpected EOF while looking for matching "'"


coreycb (dismissed):

wmontwe:
@coreycb I had to tweak the action to use gh to be able to debug it locally, now it only posts one message and updates it based on the validation state.

coreycb (approved):

PR #10959, opened 2026-04-29
fix: wrong migration type for account settings upgrade 104

### LINKED ISSUE #10958: Importing settings fails to add avatar monogram
When importing older settings like the QR code (version 99), the settings migration 104 should add the avatar monogram based on the user name or email. This fails as the `name` and `email` settings are not accessible during the settings upgrade.

The issue is caused by using the wrong type of migration that just migrates account settings but not account information like name, email, identities etc.

To fix the issue we need to use the `CombinedSettingsUpgrader` instead.

This is the source of users importing settings from Desktop or from previously exported settings will end up with a `XX` avatar monogram.

wmontwe:
Resolves #10958


rafaeltonholo (approved):
LGTM

PR #10962, opened 2026-04-29
docs: add settings documentation

### LINKED ISSUE #10960: Document SettingUpgraders
The settings are hard to understand and we need to start to document our current understanding of how these work to avoid issues like #10959

wmontwe:
Resolves #10960 

This goes a bit beyond the `SettingsUpgraders` and reflects my current knowledge about the settings in general.

### DIFF
--- a/docs/architecture/settings.md
+++ b/docs/architecture/settings.md
@@ -1,0 +1,143 @@
+# ⚙️ Settings Architecture
+
+This document describes the settings architecture in Thunderbird for Android, explaining how preferences are organized,
+stored, and upgraded.
+
+## 🏗️ Overview
+
+Settings are divided into several scopes to handle different levels of configuration:
+
+```mermaid
+graph TD
+    subgraph Settings
+        APP[General Settings]
+        ACCOUNT[Account Settings]
+        IDENTITY[Identity Settings]
+        FOLDER[Folder Settings]
+    end
+
+    ACCOUNT --> IDENTITY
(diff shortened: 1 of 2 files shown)


docs/architecture/settings.md
```diff
@@ -0,0 +1,128 @@
...
+
+The upgrade process follows a specific sequence:
+
+```mermaid
+sequenceDiagram
+    participant AU as AccountSettingsUpgrader
+    participant SU as SettingsUpgradeHelper
+    participant CU as CombinedSettingsUpgrader
+    
+    rect rgb(240, 240, 240)
```
dani-zilla: I will say, this looks better on light mode than dark mode. It's a bit hard to read if your browser automatically adjusts and puts Github on a dark background based on your settings. A slightly darker gray might work for both light and dark. The issue is the text becomes white on dark mode, but the rectangle is still nearly white. 

Image
Image

wmontwe: Yeah, I'm a light mode user and didn't check the dark version. Will update.

wmontwe: Seems like GitHub has some issues with their dark theme for sequence diagrams. I added a custom theme that should improve the readability.

dani-zilla (approved):
It's funny to think that light mode was the only mode for so long, and now I often forget it exists too. But, looks a lot better now, both graphs. Thanks! Love seeing more documentation 👍

PR #10967, opened 2026-05-01
Fix: Save all option overrides attachment files

### LINKED ISSUE #10965: More than 1 attached file overridden by the option "Save all"
### Checklist

### App

Thunderbird for Android

### App version

18.0

F-Droid

### Android version

16.0 - LineageOS

### Device model

_No response_

### Steps to reproduce

1. Write a new email
2. Attach more than 1 file (for example 3 files with the same extension type: text_1.txt, text_2.txt, text_3.txt)
3. Write your subject and body content
4. Send it to yourself
5. Once received, open it and click on the attachments button "View all"
6. Your three files will be listed at the bottom popup that appears. Click on "Save all"
7. Select a folder destination and save them
8. Now open your folder explorer and check the three files... all three are the same (same size, same content). Thunderbird saved correctly the third one, but the first and the second file are just a copy of the third one (renamed).

### Expected behavior

I expect to save all my attachments with the original content

### Actual behavior

Thunderbird corrupts attachments files when you download more than two files using the option "Save all".

But if I download them individually (not clicking on "Save all") the files keep the original content and works as expected.

### Logs

_No response_

mohitsatr:
Fixes #10965

#### Description

Makes use of `OpenDocumentTree Contract` for `onSaveAllAttachments` option to let users choose a destination folder for all the attachments. Previously, we were forced to click 'save' for each attachment file.

#### Screen Shots

https://github.com/user-attachments/assets/a643b473-da0b-4b1b-a9dd-653cbe5fd65d

## AI Disclosure

## Contribution Checklist

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/AttachmentController.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/AttachmentController.kt
@@ -4,6 +4,7 @@ import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
+import android.provider.DocumentsContract
 import android.widget.Toast
 import androidx.annotation.WorkerThread
 import app.k9mail.legacy.message.controller.SimpleMessagingListener
@@ -66,6 +67,21 @@ class AttachmentController internal constructor(
         }
     }
 
+    fun saveAttachmentToDirectory(scope: CoroutineScope, directoryUri: Uri?) {
+        if (directoryUri == null) return
+        scope.launch {
+            if (!attachment.isContentAvailable) {
+                val success = downloadAttachment()
+                if (success) {
+                    attachmentDisplayController.refreshAttachmentThumbnail(attachment)
+                    saveLocalAttachmentToDirectory(directoryUri)
+                }
+            } else {
+                saveLocalAttachmentToDirectory(directoryUri)
+            }
+        }
+    }
+
     private suspend fun saveLocalAttachmentTo(documentUri: Uri) {
         val success = withContext(ioDispatcher) {
             try {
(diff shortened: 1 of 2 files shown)


legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/AttachmentController.kt
```diff
@@ -0,0 +1,188 @@
...
+import com.fsck.k9.ui.R
+import java.io.IOException
+import kotlin.coroutines.resume
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.withContext
+import net.thunderbird.core.android.account.LegacyAccountDto
+import net.thunderbird.core.logging.legacy.Log.e
```
dani-zilla: This class is one we're planning on deprecating. I understand this was a kotlin conversion of a Java class, but if we're introducing new code (and a small change at that) we should use the newer class. It's `net.thunderbird.core.logging.Logger`. You can inject an instance of it. 

We have more details of it here: https://github.com/thunderbird/thunderbird-android/blob/49e1eab7e2df55ac83ec02e6e3f8c370ac4843c1/core/logging/README.md

rafaeltonholo (changes requested):
Although we are aiming to migrate all Java to Kotlin, the conversion should not be done in bug fixes like this one. Having a Kotlin conversion in the same changeset as a bug fix makes it harder to review and to revert in case of necessity. Please, revert the AttachmentController Kotlin conversion and apply the fix on the Java side.

mohitsatr:
@rafaeltonholo
I'm not familiar with AyncTasks. 
Would it be okay if I send a separate PR for `AttachmentController.java` -> `.kt` conversion first, and then continue on this PR?

rafaeltonholo:
> @rafaeltonholo I'm not familiar with AyncTasks. Would it be okay if I send a separate PR for `AttachmentController.java` -> `.kt` conversion first, and then continue on this PR?

Yes, please! Here is the GitHub issue for the conversion: #11096

rafaeltonholo:
Hi @mohitsatr, now that the `AttachmentController` is already converted to Kotlin, would you please rebase your changes and apply the fix?

Thanks

dani-zilla:
Thanks for all the work on this! The conversion of the file to Kotlin nd then splitting this out and updating it was super helpful. 

In my testing, saving three rtf files, only the second one is preserved. The first matches the third, its contents erased, the second remains correct. I tried it twice, and had the same result both times. It does appear to be better than main, so the fix works, I just don't believe it's being applied properly to all files. Are you able to see something similar in your testing?

mohitsatr:
@dani-zilla, No, I tested with  6 .rtf files, each about 10 kb, and they all get saved properly without their content being duplicated or erased.  I also tried testing with a mix of .pdf, .jpeg, and .png files, and they too are being stored properly. Could you also check the behaviour with these other file types in your testing?

dani-zilla (approved):
I went back through and tested with txt, png, and rtf, and they all seem to work. Not sure what was the issue, but it seems it was on my end, not the branch. So, with that all said, I think this is good to approve. We're in a soft freeze until the end of the week, so I'll label it as such and come back to merge later. Thanks for the contribution and all the work you put into this!

rafaeltonholo (approved):
LGTM. Once the soft-freeze is lifted, we can merge this! Thanks for your contribution

PR #11000, opened 2026-05-09
fix: unified inbox swipe delete

### LINKED ISSUE #10998: Only the first swiped message is deleted in Unified Inbox (subsequent swipes do nothing)
### Checklist

### App

Thunderbird for Android

### App version

18.0

F-Droid

### Android version

11

### Device model

_No response_

### Steps to reproduce

1. Open Thunderbird for Android.
2. Go to **Unified Inbox**.
3. Ensure there are several messages visible in the list.
4. Swipe the first message to delete it.
5. Immediately swipe the second message.
6. Swipe the third message, and so on.
7. Restart the app.
8. Check how many messages were actually deleted.

### Expected behavior

Every swipe gesture should successfully delete (or move to Trash) the corresponding message.

If I perform 5 swipes → 5 messages should be deleted permanently from the Unified Inbox.

### Actual behavior

Only the **very first** swipe deletes the message successfully.
- All following swipes are ignored: the message disappears temporarily, but reappears after app restart.
- Effectively, only **one message** can be deleted per app session using swipe in the Unified Inbox.
- Multi-select + Delete button works correctly.
- Swipe-to-delete works fine in individual account folders.

https://github.com/user-attachments/assets/65f431c6-0959-40fe-ae7d-ef83553b6758

### Logs

[k9mail-logs-clear.txt](https://github.com/user-attachments/files/27533086/k9mail-logs-clear.txt)

namelessweakl1ng:
## Summary

Fixes an issue where swipe-to-delete in Unified Inbox only persisted for the first message deletion during an app session.

Previously:

* The first swipe deletion worked correctly.
* Subsequent swipe deletions temporarily removed messages from the UI, but the messages reappeared after restarting the app.
* Multi-select delete and swipe-delete in individual folders continued to work correctly.

## Root Cause

The Unified Inbox swipe-delete flow was not correctly persisting state updates for subsequent delete operations within the same session.

## Changes

* Updated the Unified Inbox delete handling to ensure every swipe action correctly persists message deletion state.
* Kept behavior consistent with existing delete flows used by individual folders and multi-select actions.
* Added/updated related logic to prevent stale message state from being restored after app restart.

## Testing

Tested by:

1. Opening Unified Inbox
2. Swiping multiple consecutive messages to delete
3. Restarting the app
4. Verifying all swiped messages remained deleted

Also verified:

* Multi-select delete still works
* Swipe-delete in individual folders still works

Fixes #10998 

## AI Usage Disclosure

AI assistance tools were used to help navigate the repository structure, trace relevant code paths, and assist with implementation/debugging. All code changes were reviewed and validated manually before submission.

### DIFF
--- a/legacy/core/src/main/java/com/fsck/k9/controller/MessagingController.java
+++ b/legacy/core/src/main/java/com/fsck/k9/controller/MessagingController.java
@@ -2166,14 +2166,10 @@ public class MessagingController implements MessagingControllerRegistry, Messagi
             }
 
             unsuppressMessages(account, messages);
-            // Ensure unified inbox (and other aggregated views) always reload after a delete.
-            // In contrast to multi-select delete, swipe-delete may trigger only in-memory hiding/unhiding,
-            // which relies on MessageListRepository notifications to update the UI.
             for (MessagingListener l : getListeners()) {
                 for (LocalMessage message : messages) {
                     String uid = message.getUid();
                     if (!uid.startsWith(K9.LOCAL_UID_PREFIX)) {
-                        // Use folderServerId so listeners can correlate with their search/account context.
                         String folderServerId = message.getFolder().getServerId();
                         l.synchronizeMailboxRemovedMessage(account, folderServerId, uid);
                     }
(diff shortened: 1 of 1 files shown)


wmontwe:
This doesn't appear to fix #10998. Or I can't reproduce it. I think the bug is related to the rapid swipe actions not triggering `deleteMessages()` while animating.

I'm working on test tooling that would allow to replicate the environment easily and then come back again to check this issue.

wmontwe:
Thank you for providing the fix. We haven't been able to confirm that the fix worked on our end.

I am therefore closing this ticket for now. Should you acquire more specific data that might help us isolate the root cause, please share it in the issue.

PR #11030, opened 2026-05-20
K9WebViewClient: log WebView main-frame load errors

### LINKED ISSUE #11029: K9WebViewClient swallows WebView load errors
The `K9WebViewClient` used by the legacy message view ([legacy/ui/legacy/src/main/java/com/fsck/k9/view/K9WebViewClient.kt](https://github.com/thunderbird/thunderbird-android/blob/main/legacy/ui/legacy/src/main/java/com/fsck/k9/view/K9WebViewClient.kt)) overrides `shouldOverrideUrlLoading`, `shouldInterceptRequest` and `onPageFinished`, but not `onReceivedError`. The QuotedMessageMvpView WebView client has the same gap.

When the WebView fails to load the message body, the failure is silent: the user sees an empty body, and nothing lands in adb logcat that points at the cause, which makes triage hard. This is especially annoying for failures specific to the message-rendering pipeline (a `cid:` part that cannot be resolved by the attachment resolver, a renderer process restart on a large/complex HTML mail, etc.).

`minSdk` for the legacy module is 23, so the API 23+ overload `onReceivedError(view, request, error)` covers every supported device on its own. Sub-resource failures (broken inline images, blocked http content in https mails) should be filtered out with `request.isForMainFrame` so the log does not get flooded.

Suggested fix: override `onReceivedError` on `K9WebViewClient` and log the error code, description and failing URL through the same `net.thunderbird.core.logging.legacy.Log` helper that the rest of the file already uses. A PR for this is at #11030.

jim-daf:
Closes #11029.

`K9WebViewClient` already overrides `shouldOverrideUrlLoading`, `shouldInterceptRequest` and `onPageFinished`. It does not override `onReceivedError`, so any failure while loading a message body in the legacy message view is silent: the user sees a blank message body, and nothing useful gets recorded in adb logcat.

This PR adds the API 23+ overload only, since `minSdk` for the legacy module is 23 (`build-plugin/src/main/kotlin/ThunderbirdProjectConfig.kt: const val sdkMin = 23`), so the deprecated pre-M overload is not reachable on any device the app ships to.

```kotlin
@RequiresApi(Build.VERSION_CODES.M)
override fun onReceivedError(view: WebView, request: WebResourceRequest, error: WebResourceError) {
    super.onReceivedError(view, request, error)
    if (request.isForMainFrame) {
        Log.d("Message WebView load error: %d %s for %s", error.errorCode, error.description, request.url)
    }
}
```

Notes on scope:

- Only main-frame errors are logged. Sub-resource errors are very common in HTML mails (broken remote images, blocked `http://` resources inside an `https://` mail) and would just spam the log without giving useful information about the root render failure.
- The log call uses `Log.d` via `net.thunderbird.core.logging.legacy.Log`, matching the existing imports and call style in this file (see the existing `Log.d(e, "Couldn't open URL: %s", uri)` on the same class).
- No new permission, no new dependency, no behaviour change for successful loads. The change is additive: an existing silent failure now produces a debug log line.

Out of scope but mentioned in the issue: `QuotedMessageMvpView` has the same gap. Happy to send a follow-up PR if maintainers prefer that one bundled, but `K9WebViewClient` is the dominant codepath and felt like the right place to start.

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/view/K9WebViewClient.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/view/K9WebViewClient.kt
@@ -6,6 +6,7 @@ import android.content.Intent
 import android.net.Uri
 import android.os.Build
 import android.provider.Browser
+import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
 import android.webkit.WebResourceResponse
 import android.webkit.WebView
@@ -118,6 +119,14 @@ class K9WebViewClient(
         super.onPageFinished(view, url)
 
         onPageFinishedListener?.onPageFinished()
+    }
+
+    @RequiresApi(Build.VERSION_CODES.M)
+    override fun onReceivedError(view: WebView, request: WebResourceRequest, error: WebResourceError) {
+        super.onReceivedError(view, request, error)
+        if (request.isForMainFrame) {
+            Log.d("Message WebView load error: %d %s for %s", error.errorCode, error.description, request.url)
+        }
     }
 
     companion object {
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/view/K9WebViewClient.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/view/K9WebViewClient.kt
@@ -22,6 +22,7 @@ import net.thunderbird.core.logging.legacy.Log
 /**
  * [WebViewClient] that intercepts requests for `cid:` URIs to load the respective body part.
  */
+@Suppress("TooManyFunctions")
 class K9WebViewClient(
     private val clipboardManager: ClipboardManager,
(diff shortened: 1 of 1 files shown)


wmontwe (dismissed):
Thank you!

wmontwe:
@jim-daf There is a warning about too many methods, this could safely be ignored for this class by adding `@Suppress("TooManyFunctions")`to the class header.

jim-daf:
> @jim-daf There is a warning about too many methods, this could safely be ignored for this class by adding `@Suppress("TooManyFunctions")`to the class header.

Thanks @wmontwe, done in [5ad43d2](https://github.com/thunderbird/thunderbird-android/commit/5ad43d29). Added @Suppress("TooManyFunctions") on the class header, right under the existing KDoc.

wmontwe (approved):
Thank you!

PR #11033, opened 2026-05-21
feat(smime): add S/MIME support via companion-app architecture

christine-ciphermail:
## Summary

Adds end-to-end S/MIME support to Thunderbird for Android via a companion-app architecture, mirroring the existing OpenPGP integration: key material and cryptographic operations live in a separate provider app, and Thunderbird talks to it over a stable AIDL API.

A reference provider implementation (CipherMail for Android) is the first consumer of this API, but any S/MIME provider can implement it.

## What's included

**New module `plugins/smime-api/`** — versioned AIDL + Parcelables + client helper. Pluggable across providers, mirrored in the CipherMail companion repo. Independent licensing under Apache-2.0.

**Integration in legacy core / UI:**
- Account-level Sign/Encrypt toggles, with per-account defaults (account settings → S/MIME)
- Compose: per-message Sign/Encrypt checkboxes; encrypt-implies-sign enforced; lock/sign icons in the recipient bar reflect provider's certificate availability check
- Message view: separate "signed" / "encrypted" status icons and an "Open in provider" action for messages Thunderbird can't render inline
- S/MIME state persisted separately from the legacy crypto-enabled flag so a message is never sent silently unencrypted when the provider is unavailable

**Safety behaviour:**
- If the provider isn't installed or returns `USER_INTERACTION_REQUIRED` (e.g. provider keystore is locked), the compose flow pauses and surfaces the action to the user — the message is preserved, not lost
- Drafts are offered a folder assignment when the composer is closed before a save completes

**Documentation:**
- `docs/adr/0009-smime-companion-app-architecture.md` — architectural decision and rationale (why companion app vs. built-in)
- `docs/security/smime-companion-threat-model.md` — threat model for the inter-process boundary
- `docs/developer/writing-smime-provider.md` — guide for implementers of the API
- `docs/user-guide/setup/enabling-smime.md` — end-user setup walkthrough

## Why companion-app and not built-in?

Same reasons OpenPGP went this route in K-9 Mail / Thunderbird for Android: cryptographic key material, certificate stores, PIN unlock flows, and smartcard support are non-trivial concerns that benefit from living in a single, dedicated, hardened process. Thunderbird stays a mail client; providers stay providers. See ADR-0009 for the full discussion.

## Test plan

- [ ] Build cleanly on AGP 9 / Gradle 9 / JDK 21
- [ ] Existing unit tests pass; new tests cover `RecipientPresenter.onSmimeCertCheckResult` branches
- [ ] Manual: enable S/MIME on an account, sign-only, encrypt+sign, send/receive between two devices using CipherMail provider
- [ ] Manual: uninstall provider, confirm error path (no silent unencrypted send, message preserved)
- [ ] Manual: provider keystore locked → `USER_INTERACTION_REQUIRED` flow opens the provider's unlock UI, retry succeeds

## Notes

- No breaking changes to existing behaviour for users who don't enable S/MIME on an account
- API module is versioned independently (`plugins/smime-api/CHANGELOG.md`) so providers and Thunderbird can evolve at their own pace


christine-ciphermail:
Please apply report: include

I can't seem to add a lablelmyself.

christine-ciphermail:
How do I add a label?

jbott-tbird:
Hi @christine-ciphermail, thank you for your interest in getting this into Thunderbird for Android.  We have had several discussions about encryption (GPG and S/MIME) in the app.  This isn't in our roadmap for 2026, though might be something we add to 2027.  For now we are prioritizing PRs that align with our 2026 roadmap.

PR #11046, opened 2026-05-26
refactor: quality gradle plugins

wmontwe:
This pull request introduces two new custom Gradle plugins for code quality: one for Detekt (static code analysis) and one for Spotless (code formatting). It migrates all modules to use these new plugins instead of the previous script-based approach. Additionally, applies format updates and fixes some of the `detektAll` findings.

This harmonizes the project setup with [Thunderbird Mobile Components](https://github.com/thunderbird/thunderbird-mobile-components)


dani-zilla (dismissed):

rafaeltonholo (approved):
LGTM. Just a nit: I would rather prefer instead of suppressing the detekt issues having them in the baseline, but not a blocker for this PR.

PR #11061, opened 2026-05-28
Option to clean up old attachments on a schedule

harshad1:
## Summary

Adds an account-level setting to remove locally downloaded attachments from IMAP messages older than a selected age. Intention is to save local space.

Message and attachment metadata are preserved so attachments can still be fetched again from the server when requested.

I wired this into the existing legacy account settings/storage path because account settings are still mostly there. 

## Notes

- Adds configurable attachment cleanup age with a default of disabled.
- Schedules periodic WorkManager cleanup for eligible IMAP accounts.
- Runs a one-time cleanup when the setting changes.
- Removes only local attachment payloads for old messages; message records and attachment parts remain.
- Does not automatically re-download attachments when the retention window is changed or disabled.
- Keeps cleanup scoped to explicit attachments, with tests covering inline/folded disposition cases.
- I used GPT codex to help me explore the codebase and draft things. Also had it add more tests.

### DIFF
--- a/legacy/core/src/main/java/com/fsck/k9/job/AttachmentCleanupWorker.kt
+++ b/legacy/core/src/main/java/com/fsck/k9/job/AttachmentCleanupWorker.kt
@@ -2,11 +2,13 @@ package com.fsck.k9.job
 
 import android.content.ContentResolver
 import android.content.Context
+import android.database.sqlite.SQLiteDatabaseLockedException
 import androidx.work.Worker
 import androidx.work.WorkerParameters
 import app.k9mail.legacy.mailstore.MessageStoreManager
 import com.fsck.k9.Preferences
 import java.util.concurrent.TimeUnit
+import net.thunderbird.core.android.account.AccountDefaultsProvider.Companion.MIN_ATTACHMENT_CLEANUP_DAYS
 import net.thunderbird.core.logging.legacy.Log
 import net.thunderbird.core.preference.BackgroundOps
 import net.thunderbird.core.preference.GeneralSettingsManager
@@ -22,7 +24,6 @@ class AttachmentCleanupWorker(
 
     override fun doWork(): Result {
         if (isBackgroundCleanupDisabled()) {
-            Log.d("Background cleanup is disabled. Skipping attachment cleanup.")
             return Result.success()
         }
 
@@ -30,7 +31,7 @@ class AttachmentCleanupWorker(
         val account = preferences.getAccount(accountUuid) ?: return Result.success()
         val retentionDays = account.attachmentCleanupDays
 
-        if (retentionDays <= 0) {
+        if (retentionDays < MIN_ATTACHMENT_CLEANUP_DAYS) {
             return Result.success()
         }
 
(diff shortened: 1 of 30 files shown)


legacy/ui/legacy/src/main/res/xml/account_settings.xml
```diff
@@ -53,6 +53,15 @@
...
             />
 
+        <ListPreference
+            android:dialogTitle="@string/account_settings_attachment_cleanup_label"
+            android:entries="@array/attachment_cleanup_entries"
+            android:entryValues="@array/attachment_cleanup_values"
+            android:key="account_attachment_cleanup"
+            android:summary="@string/account_settings_attachment_cleanup_summary"
+            android:title="@string/account_settings_attachment_cleanup_label"
+            />
```
harshad1, line 63: Intentionally wired into the existing legacy account settings screen because account settings still use this

legacy/core/src/main/java/com/fsck/k9/job/AttachmentCleanupWorkerManager.kt
```diff
@@ -0,0 +1,81 @@
...
+import androidx.work.OneTimeWorkRequestBuilder
+import androidx.work.PeriodicWorkRequestBuilder
+import androidx.work.WorkManager
+import androidx.work.workDataOf
+import java.util.concurrent.TimeUnit
+import net.thunderbird.core.android.account.LegacyAccountDto
+import net.thunderbird.core.common.mail.Protocols
+import net.thunderbird.core.logging.legacy.Log
+
+class AttachmentCleanupWorkerManager(
```
harshad1, line 16: Cleanup work is scheduled only for IMAP accounts with a non-zero retention value. Disabling the setting or switching to a non-IMAP account cancels both periodic and pending one-time cleanup work.

legacy/core/src/main/java/com/fsck/k9/job/AttachmentCleanupWorker.kt
```diff
@@ -0,0 +1,61 @@
...
+            Result.retry()
+        }
+    }
+
+    private fun isBackgroundCleanupDisabled(): Boolean {
+        return when (generalSettingsManager.getConfig().network.backgroundOps) {
+            BackgroundOps.NEVER -> true
+            BackgroundOps.ALWAYS -> false
+            BackgroundOps.WHEN_CHECKED_AUTO_SYNC -> !ContentResolver.getMasterSyncAutomatically()
+        }
```
harshad1, line 88: Attachment cleanup follows the existing background-work policy, including WHEN_CHECKED_AUTO_SYNC via Android master sync, so it should not run when the user has disabled background account work.

legacy/storage/src/main/java/com/fsck/k9/storage/messages/AttachmentCleanupOperations.kt
```diff
@@ -0,0 +1,157 @@
+package com.fsck.k9.storage.messages
+
+import android.content.ContentValues
+import android.database.sqlite.SQLiteDatabase
+import com.fsck.k9.mail.internet.MimeHeader
+import com.fsck.k9.mail.internet.MimeUtility
+import com.fsck.k9.mailstore.LockableDatabase
+import net.thunderbird.core.common.mail.Flag
+
+internal class AttachmentCleanupOperations(
```
harshad1, line 11: The SQL query only pre-filters candidate parts; attachment detection is done by parsing disposition so inline parts with filenames like 'attachment.png' are not removed.

legacy/storage/src/main/java/com/fsck/k9/storage/messages/AttachmentCleanupOperations.kt
```diff
@@ -0,0 +1,157 @@
...
+                                messageId = cursor.getLong(2),
+                            ),
+                        )
+                    }
+                }
+            }
+        }
+    }
+
+    private fun SQLiteDatabase.markPartsAsMissing(messagePartIds: List<Long>) {
```
harshad1, line 79: Cleanup removes only the local payload and marks the part as missing. It preserves message part metadata such as filename/server reference so the attachment can be fetched again on demand.

harshad1:
IMO this is ready for review.

The jankiest part is the attachment detection - we finds candidates by doing a SQL LIKE over raw headers, then parse in code to decide whether the part is really an attachment. This is kinda fragile and expensive - lots of redundant work on every run.

The cleaner version would store attachment-ness as structured data when saving the message part, then the cleanup can query an indexed column instead of interpreting MIME headers.

But I did not want to introduce a schema change and a migration rn

harshad1:
It looks like I will not have the time to address feedback etc going forward. 

Closing this now.

If anyone wants to use this code, please feel free to open a new pr using it

PR #11082, opened 2026-06-02
fix(message-reader): inlined image not displaying

### LINKED ISSUE #10889: Inline images no longer display inline after v18.0 update
### Checklist

### App

K-9 Mail

### App version

18.0

Google Play

### Android version

16

### Device model

Pixel 8a

### Steps to reproduce

View a message with a multi-part message in MIME format including an image Content-Disposition: inline
In account settings->Reading mail->Alway show images is set to From anyone

### Expected behavior

Image is seen inline when opening the message

### Actual behavior

Image is shown as an attachment which has to be viewed in an external viewer.

### Logs

This was working before updating to v18.0

rafaeltonholo:
## Contribution Summary

Fixes: #10889

#### Description

This PR brings back the attachment list at the bottom of the message. Instead of just reverting the removal of the old one, I've implemented a new one using Compose, since it will be useful in the future when we migrate the Message Reader to Compose.

**Changes included:**
- Convert `AttachmentViewInfo` to Kotlin
- Added `BoxScope` to `RemoteImage`
- Added the Attachment list back to the message reader
- Make sure the Attachment header bar is fully clickable, instead of only the `View` button.

#### Screenshots

|     |
| :-----: | 
| Inlined images only |
|  |
| Inlined images + attached image | 
|  |
| Inlined images + attached files/image |
|  |
| Attachments without images | 
|  |
| Inlined image AND file attachment (Before download complete message) |
|  |
| Inlined image AND file attachment (After download complete message) |
| image |

## AI Disclosure


dani-zilla:
Sometimes I could get inline images, but other times they appear below. I'm trying to figure out when this happens, but I think it could be down to your account settings, if you always allow images. With this account, I have it set to always allow images, yet the photo I sent ended up below the email, with a divider line where the image should have been between the text. 
testEmailInline

That's not to say it doesn't always work:
inlineWorkingThough

dani-zilla:
> Sometimes I could get inline images, but other times they appear below. I'm trying to figure out when this happens, but I think it could be down to your account settings, if you always allow images. With this account, I have it set to always allow images, yet the photo I sent ended up below the email, with a divider line where the image should have been between the text. 

Looking into this now and I'll come up with a more detailed test plan, but what I found so far is that it is interpreted as a jpeg mimetype and should display. I'll try to piece out which cases fail or if it's from a specific email service.

dani-zilla:
imageSmallEnoughToNotNeedDownload
Might have figured out at least one issue. It's not putting the image inline if it's downsized by the email sender. I used my mail client to downsize this image to "Small" when sending so it wouldn't hit the kb threshold for being too large and requiring a separate download. 

Also, I found another issue with displaying inline images that **are** too large. When you download the image and it re-draws to display the image inline, it's HUGE. Too large to fit in the view area. It will either shrink the text to display the entire image if the user has selected the "Auto-fit messages" option in Settings > General > Display. If auto-fit messages is off, it's just a huge image. We may want to consider auto-scaling such images, but perhaps that's out of scope for this fix, as I do believe that was the behavior before we encountered this bug.

rafaeltonholo:
> Might have figured out at least one issue. It's not putting the image inline if it's downsized by the email sender. I used my mail client to downsize this image to "Small" when sending so it wouldn't hit the kb threshold for being too large and requiring a separate download. 

Could you please share with me the `.eml` file of the email you are experiencing this? This could be how the email sender has built the email.

> Also, I found another issue with displaying inline images that **are** too large. When you download the image and it re-draws to display the image inline, it's HUGE. Too large to fit in the view area. It will either shrink the text to display the entire image if the user has selected the "Auto-fit messages" option in Settings > General > Display. If auto-fit messages is off, it's just a huge image. We may want to consider auto-scaling such images, but perhaps that's out of scope for this fix, as I do believe that was the behavior before we encountered this bug.

I will create a separate issue to investigate this matter. Resizing images in this way would require JavaScript integration, which is disabled in our webview.

dani-zilla:
I really don't know what to say. I've sent dozens of emails to myself from a variety of accounts. iCloud, Gmail, Thundermail, Proton. I used the Thunderbird desktop client, Gmail on the web, my mail app, the Thunderbird app (can't do inline attachments anyway), and Apple's Mail app on iOS. With the exception of Proton on Android, they all showed up with the inline image at the bottom and a divider line where the image should be. I think it may have to do with what they send through, how they denote its place in the email, whether or not they did resizing with or without my permission, I don't know. 

How are you testing that you're not seeing similar results?

rafaeltonholo:
> I really don't know what to say. I've sent dozens of emails to myself from a variety of accounts. iCloud, Gmail, Thundermail, Proton. I used the Thunderbird desktop client, Gmail on the web, my mail app, the Thunderbird app (can't do inline attachments anyway), and Apple's Mail app on iOS. With the exception of Proton on Android, they all showed up with the inline image at the bottom and a divider line where the image should be. I think it may have to do with what they send through, how they denote its place in the email, whether or not they did resizing with or without my permission, I don't know.
> 
> How are you testing that you're not seeing similar results?

I'm doing nothing special. Using Gmail (both app and Webmail) to send the emails with the inline images, I was able to get the inline images inside the app, after downloading the whole message:

image

---

I've also tested via webmail, resizing the images as you reported, which was a different behaviour, but shows fine here:
image
image
image

dani-zilla:
Have you tried clients other than Gmail? Because I did just use the web view on Gmail with my personal account and it worked, though it hadn't always and obviously not for all photos. I also found that anything sent from my iPhone's mail client or my Mac's would also lead to the image not displaying inline, regardless of whether or not it came from iCloud or Gmail. So I believe it's the mail client itself that's the issue here. Clearly the fix doesn't account for all the ways an inline image could be attached. 

icloudFromiPhone

Screenshot 2026-06-12 at 11 20 16 AM

rafaeltonholo:
> Have you tried clients other than Gmail? Because I did just use the web view on Gmail with my personal account and it worked, though it hadn't always and obviously not for all photos. I also found that anything sent from my iPhone's mail client or my Mac's would also lead to the image not displaying inline, regardless of whether or not it came from iCloud or Gmail. So I believe it's the mail client itself that's the issue here. Clearly the fix doesn't account for all the ways an inline image could be attached. 
> 
> icloudFromiPhone
> 
> Screenshot 2026-06-12 at 11 20 16 AM
> 
> 

Yes, I also used the Thunderbird Desktop. You can see examples in the description of the PR.

Can you share with me the `.eml` files for the emails you are not seeing the image inlined, but attached? That could be on how the email client you are using to send the message is building it. Have you also checked how the message you sent looks like on other clients?

dani-zilla:
> Yes, I also used the Thunderbird Desktop. You can see examples in the description of the PR.

I can confirm that thunderbird desktop works to Gmail and Thundermail

> Can you share with me the `.eml` files for the emails you are not seeing the image inlined, but attached? That could be on how the email client you are using to send the message is building it. Have you also checked how the message you sent looks like on other clients?

Here's the same email that isn't displaying right up there opened in the Mail app on iOS 26:

IMG_1049

And that exact email in thunderbird:

icloudFromiPhone

dani-zilla:
We found that, while this update does fix the initial issue, we found a **separate** issue, and may be primarily linked to how Apple Mail formats their messages. They use the format of the .eml text itself to show where inline images should display, instead of the content ID (CID). We'll have to make a separate ticket to investigate and fix that, but have declared it out of scope for this.

dani-zilla (approved):

PR #11098, opened 2026-06-04
chore(deps): bump dependencies

wmontwe:
Resolves #11095 

Bumped:

 - androidGradlePlugin [9.1.0 -> 9.2.1]
 - androidxActivity [1.12.4 -> 1.13.0]
     https://developer.android.com/jetpack/androidx/releases/activity#1.13.0
 - androidxAnnotation [1.9.1 -> 1.10.0]
     https://developer.android.com/jetpack/androidx/releases/annotation#1.10.0
 - androidxCamera [1.5.2 -> 1.6.1]
     https://developer.android.com/jetpack/androidx/releases/camera#1.6.1
 - androidxComposeBom [2026.01.00 -> 2026.05.01]
     https://developer.android.com/jetpack/compose/bom/bom-mapping
 - androidxCore [1.17.0 -> 1.18.0]
     https://developer.android.com/jetpack/androidx/releases/core#1.18.0
 - androidxDatastore [1.2.0 -> 1.2.1]
     https://developer.android.com/jetpack/androidx/releases/datastore#1.2.1
 - androidxGraphicsPath [1.1.0-alpha01 -> 1.1.0]
     https://developer.android.com/jetpack/androidx/releases/graphics#1.1.0
 - androidxNavigation [2.9.7 -> 2.9.8]
     https://developer.android.com/jetpack/androidx/releases/navigation#2.9.8
 - androidxWindow [1.3.0 -> 1.5.1]
     https://developer.android.com/jetpack/androidx/releases/window#1.5.1
 - androidxWork [2.11.0 -> 2.11.2]
     https://developer.android.com/jetpack/androidx/releases/work#2.11.2
 - clikt [5.0.3 -> 5.1.0]
     https://github.com/ajalt/clikt/releases/tag/5.1.0
 - landscapist [2.9.5 -> 2.9.8]
     https://github.com/skydoves/landscapist/releases/tag/2.9.8
 - androidMaterial [1.13.0 -> 1.14.0]
     https://github.com/material-components/material-components-android/releases/tag/1.14.0
 - jetbrainsCompose [1.11.0 -> 1.11.1]
     https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.11.1
 - kotlinBom [2.3.21 -> 2.4.0]
     https://github.com/JetBrains/kotlin/releases/tag/v2.4.0

PR #11113, opened 2026-06-05
feat: add core UI testing

### LINKED ISSUE #11112: Add KMP UI testing
Due to the migration to AGP 9 (#10424) the Compose tests are not working anymore for Android targets. We need to fix them and add support for the new `androidHostTest` source set for UI testing.

This is necessary for the Bolt extraction #10391

wmontwe:
## Contribution Summary

Depends on #11098 

Resolves #11112 

#### Description

Added a new `core:ui:testing` module and a `ComposeUiTestHarness` to allow Compose UI testing for Android and JVM. The `ComposeUiTest` is wrapped in a `ComposeUiTestScope` to decouple platform specific test execution for `androidHostTest`.

`Mockito` is removed from the shared dependencies in the version catalog to discourage it's use.

Partial changes are cherry-picked from the bolt refactor.

## AI Disclosure

## Contribution Checklist

### DIFF
--- a/core/ui/testing/src/commonMain/kotlin/net/thunderbird/core/ui/testing/ComposeUiTestScope.kt
+++ b/core/ui/testing/src/commonMain/kotlin/net/thunderbird/core/ui/testing/ComposeUiTestScope.kt
@@ -1,0 +1,58 @@
+package net.thunderbird.core.ui.testing
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.test.SemanticsNodeInteraction
+import androidx.compose.ui.test.SemanticsNodeInteractionCollection
+
+/**
+ * A scope for Compose UI tests. It provides a platform agnostic way to use ComposeUiTest API.
+ */
+public interface ComposeUiTestScope {
+
+    /**
+     * Sets the content of the Compose UI test scope.
+     */
+    public fun setContent(content: @Composable () -> Unit)
+
+    /**
+     * Find a node with the given [tag].
+     *
+     * @param tag The tag of the node to find.
+     * @param useUnmergedTree Whether to use the unmerged tree.
+     * @return A [SemanticsNodeInteractionCollection] containing the found nodes.
+     */
+    public fun onNodeWithTag(
+        tag: String,
+        useUnmergedTree: Boolean = false,
+    ): SemanticsNodeInteraction
+
+    /**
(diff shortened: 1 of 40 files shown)


core/ui/testing/src/commonMain/kotlin/net/thunderbird/core/ui/testing/ComposeUiTestScope.kt
```diff
@@ -0,0 +1,63 @@
+package net.thunderbird.core.ui.testing
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.test.SemanticsNodeInteraction
+import androidx.compose.ui.test.SemanticsNodeInteractionCollection
+
+/**
+ * A scope for Compose UI tests. It provides a platform agnostic way to use ComposeUiTest API.
+ */
+public interface ComposeUiTestScope {
```
rafaeltonholo, line 10: I'm not sure I understand the reasoning behind having this interface. The `ComposeUiTest` interface is already platform-agnostic and provides many other testing functionalities that we are lacking by creating our own interface.

Unless I am missing some important reasoning, I would suggest using `ComposeUiTest` directly instead of creating a new one.

wmontwe: Current `ComposeUiTest` is experimental and requires agressive opt-in. You could add a general opt-in, but that also covers other api's that I don't wan't to enable.
There is also no estimate on when this API will be stable. So it's just convenience, either have a wrapper or opt in whenever you write a test.

This also allows to verify what works within the custom harness, by writing a test.

rafaeltonholo: Understood. In that case, it's fair; however, I recommend that this interface implement the `SemanticsNodeInteractionsProvider` an then use the `SemanticsNodeInteractionsProvider by delegate` in the actual implementations. 

This approach will allow us to access the stable Espresso methods used for testing. Otherwise, we'll have to re-implement or delegate many Espresso methods that we already use in the codebase.

rafaeltonholo (approved):
LGTM. We can address the `SemanticsNodeInteractionsProvider` later if we need.

PR #11118, opened 2026-06-08
fix: stop linkifier from swallowing trailing punctuation in URLs

### LINKED ISSUE #11060: Link highlighting is incorrect when there's `),` at the end
### Checklist

### App

K-9 Mail

### App version

18.0

F-Droid

### Android version

Android 15 (e/OS)

### Device model

Fairphone

### Steps to reproduce

1. Receive an email with text like this
```
* Trung Nguyen kicked off a discussion on [adding support for the old GCC2 C++
  ABI](https://discourse.llvm.org/t/rfc-clang-with-gcc2-c-abi/90825),
  motivated by projects like HaikuOS that need ABI compatibility.
```
```
* Nikita Popov made a proposal, '[yet another strict
  FP](https://discourse.llvm.org/t/rfc-yet-another-strict-fp/90798)'. "I
```
2. Try to click the links

### Expected behavior

The links should work correctly.
In both cases they work correctly in the Desktop version of Thunderbird.

### Actual behavior

K9 thinks that the trailing `),` or `)'` is part of the URL, making the link not work.

Links with trailing `)` or `).` (followed by a space) work fine in both K9 and Thunderbird.

### Logs

_No response_

mvanhorn:
## Contribution Summary

Linked Issue/Ticket: Fixes #11060

#### Description

The URL linkifier was swallowing trailing punctuation into the linked href. A URL followed by punctuation, like `see https://example.com.` or `(https://example.com)`, had the trailing `.`/`,`/etc. pulled into the link, producing a broken URL. The existing trimming logic only handled a single end-of-sentence character.

This change extracts the trailing-punctuation trimming into a dedicated function that walks backward from the URL end, iteratively stripping clause punctuation (`,;:'"`) and sentence-ending characters, while preserving the existing bracket (`skipChar`) behavior and the "don't trim before `>`" rule. Because it trims iteratively, it correctly handles multiple trailing punctuation characters, not just one.

#### Screen Shots

_No UI changes; this is text-parsing logic in the HTML linkifier._

## AI Disclosure

## Contribution Checklist

### DIFF
--- a/legacy/core/src/main/java/com/fsck/k9/message/html/HttpUriParser.kt
+++ b/legacy/core/src/main/java/com/fsck/k9/message/html/HttpUriParser.kt
@@ -39,16 +39,36 @@ internal class HttpUriParser : UriParser {
             currentPos = matchUnreservedPCTEncodedSubDelimClassesGreedy(text, currentPos + 1, ":@/?")
         }
 
-        if (text.isEndOfSentence(currentPos - 1)) {
-            currentPos--
-        }
-
-        if (text[currentPos - 1] == skipChar) {
-            currentPos--
-        }
+        currentPos = text.trimTrailingPunctuation(startPos, currentPos, skipChar)
 
         val uri = text.subSequence(startPos, currentPos)
         return UriMatch(startPos, currentPos, uri)
+    }
+
+    private fun CharSequence.trimTrailingPunctuation(startPos: Int, endPos: Int, skipChar: Char?): Int {
+        var currentPos = endPos
+        var shouldKeepTrimming = true
+
+        while (shouldKeepTrimming && currentPos > startPos) {
+            val position = currentPos - 1
+
+            if (skipChar != null && this[position] == skipChar) {
+                currentPos--
+                shouldKeepTrimming = false
+            } else if (isTrailingPunctuation(position)) {
+                currentPos--
+            } else {
+                shouldKeepTrimming = false
+            }
+        }
+
+        return currentPos
+    }
+
+    private fun CharSequence.isTrailingPunctuation(position: Int): Boolean {
+        if (position < lastIndex && this[position + 1] == '>') return false
+
+        return this[position] in CLAUSE_PUNCTUATION || isEndOfSentence(position)
     }
 
     private fun getSkipChar(text: CharSequence, startPos: Int): Char? {
@@ -269,6 +289,7 @@ internal class HttpUriParser : UriParser {
     companion object {
         // This string represent character group sub-delim as described in RFC 3986
         private const val SUB_DELIM = "!$&'()*+,;="
+        private const val CLAUSE_PUNCTUATION = ",;:'\""
         private val SCHEME_REGEX = "(https?|rtsp)://".toRegex(RegexOption.IGNORE_CASE)
         private val DOMAIN_REGEX =
             "[\\da-z](?:[\\da-z-]*[\\da-z])*(?:\\.[\\da-z](?:[\\da-z-]*[\\da-z])*)*(?::(\\d{0,5}))?"
--- a/legacy/core/src/main/java/com/fsck/k9/message/html/HttpUriParser.kt
+++ b/legacy/core/src/main/java/com/fsck/k9/message/html/HttpUriParser.kt
@@ -39,36 +39,10 @@ internal class HttpUriParser : UriParser {
             currentPos = matchUnreservedPCTEncodedSubDelimClassesGreedy(text, currentPos + 1, ":@/?")
         }
 
-        currentPos = text.trimTrailingPunctuation(startPos, currentPos, skipChar)
+        currentPos = text.endIndexWithoutTrailingPunctuation(startPos, currentPos, skipChar)
 
         val uri = text.subSequence(startPos, currentPos)
         return UriMatch(startPos, currentPos, uri)
-    }
-
-    private fun CharSequence.trimTrailingPunctuation(startPos: Int, endPos: Int, skipChar: Char?): Int {
-        var currentPos = endPos
-        var shouldKeepTrimming = true
-
-        while (shouldKeepTrimming && currentPos > startPos) {
-            val position = currentPos - 1
-
-            if (skipChar != null && this[position] == skipChar) {
-                currentPos--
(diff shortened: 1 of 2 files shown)


legacy/core/src/main/java/com/fsck/k9/message/html/HttpUriParser.kt
```diff
@@ -39,18 +39,38 @@ internal class HttpUriParser : UriParser {
...
-        if (text[currentPos - 1] == skipChar) {
-            currentPos--
-        }
+        currentPos = text.trimTrailingPunctuation(startPos, currentPos, skipChar)
 
         val uri = text.subSequence(startPos, currentPos)
         return UriMatch(startPos, currentPos, uri)
     }
 
+    private fun CharSequence.trimTrailingPunctuation(startPos: Int, endPos: Int, skipChar: Char?): Int {
```
dani-zilla: I think that because this function does not return a modified CharSequence it may need a different name, reflecting that it returns the index of the last non-punctuation character in the URI. 

This change removes the use of the `isEndOfSentence` function, moving it into this one. I think the function name should reflect that.

legacy/core/src/main/java/com/fsck/k9/message/html/HttpUriParser.kt
```diff
@@ -39,18 +39,38 @@ internal class HttpUriParser : UriParser {
...
+        var currentPos = endPos
+        var shouldKeepTrimming = true
```
dani-zilla: Is there any reason we'd need this instead of simply breaking out of the while loop when the condition is met?

legacy/core/src/main/java/com/fsck/k9/message/html/HttpUriParser.kt
```diff
@@ -39,18 +39,38 @@ internal class HttpUriParser : UriParser {
...
+                currentPos--
+            } else {
+                shouldKeepTrimming = false
+            }
+        }
+
+        return currentPos
+    }
+
+    private fun CharSequence.isTrailingPunctuation(position: Int): Boolean {
```
dani-zilla: This probably should be down with the `isEndOfSentence` function, just to keep these extension functions together in the code.

legacy/core/src/test/java/com/fsck/k9/message/html/HttpUriParserTest.kt
```diff
@@ -276,6 +276,24 @@ class HttpUriParserTest {
...
+    fun `URI wrapped in parentheses followed by a comma`() {
+        val input = "(https://discourse.llvm.org/t/rfc-clang-with-gcc2-c-abi/90825),"
+
+        val uriMatch = parser.parseUri(input, 1)
+
+        assertUriMatch("https://discourse.llvm.org/t/rfc-clang-with-gcc2-c-abi/90825", uriMatch, 1)
+    }
+
+    @Test
+    fun `URI wrapped in parentheses followed by an apostrophe`() {
```
dani-zilla, line 305: I think I'd like to see a test that checks for multiple punctuation symbols at the end. One of the benefits of your new function is that it can move the URL's endpoint back by more than a single character

dani-zilla (commented):
I should note, we do prefer you stick to the template now that it's been update to serve as a better template. This also defines better if AI was primarily used or only assisted the contributor. Thanks for the contribution! There are just a few small notes I had.

mvanhorn:
Thanks, all four addressed in d16858f:

- Renamed `trimTrailingPunctuation` -> `endIndexWithoutTrailingPunctuation` (and the call site), since it returns the index of the last non-punctuation character rather than a trimmed CharSequence.
- Dropped the `shouldKeepTrimming` flag in favor of `break`, same behavior, less state to track.
- Moved the helper down next to the other `CharSequence` extension functions so they're grouped.
- Added a test for multiple trailing punctuation characters (`.../path,;`), exercising the iterative trim walking back more than one character.

I also moved the PR description over to the current template with the AI disclosure filled in. One heads-up: I don't have a JVM locally, so I couldn't run `testDebugUnitTest`/`spotlessCheck` myself, relying on CI for those. If anything trips, I'll fix it.

dani-zilla:
You may have seen it already, but the test `IPv4 address with empty port` in `HttpUriParserTest.kt` is failing. The ':' at the end of the URI is being removed, causing a failure. 

```
Expected :...tIndex=0, endIndex=1[7, uri=http://127.0.0.1:])
Actual   :...tIndex=0, endIndex=1[6, uri=http://127.0.0.1])
```
Here's the thing, **your** code is right, the test is wrong, according to [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986). 

> "URI producers and normalizers should omit the ":" delimiter that separates host from port if the port component is empty"

So, I'm wondering if you could update the test to reflect the fact that you fixed a bug that the test was failing to report. 

Thanks, and what a funny way to find we were doing something unexpected incorrectly!

mvanhorn:
CI flagged one pre-existing test that conflicts with this fix by design: 'IPv4 address with empty port' asserted that http://127.0.0.1: keeps the trailing colon. Since a bare trailing colon after a host is exactly the prose-punctuation case this PR addresses (and an empty port isn't a fetchable URI), I updated the expectation in e327c338 so the match ends at the host. If you'd rather preserve the old empty-port behavior I can special-case it instead - just say the word.

dani-zilla:
> CI flagged one pre-existing test that conflicts with this fix by design: 'IPv4 address with empty port' asserted that http://127.0.0.1: keeps the trailing colon. Since a bare trailing colon after a host is exactly the prose-punctuation case this PR addresses (and an empty port isn't a fetchable URI), I updated the expectation in [e327c33](https://github.com/thunderbird/thunderbird-android/commit/e327c338b3ac585f5e8929fdc226c29b115cca62) so the match ends at the host. If you'd rather preserve the old empty-port behavior I can special-case it instead - just say the word.

Yup, that works for me. The way we were doing it was technically wrong, so this is a good fix in more than one way. Thanks!

dani-zilla (approved):

mvanhorn:
Thanks @dani-zilla for the merge - the linkifier stops swallowing trailing punctuation into URLs now.

PR #11120, opened 2026-06-08
Automatically generate contributor list for releases

stephenking314:
## Contribution Summary

Generate contributor list and compare url for github release notes.

The goal of this change is to fully automate the generation of github release notes removing an error prone, tedious, manual step.

Additionally this includes a minor readability update for the render notes script.

Pipeline runs performed in fork repo. Note, there was no attempt to fully model the shippable builds pipeline in the private repo since this pipeline uses secrets for various task, and completion of these tasks was not necessary to exercise the desired code paths.  See Release Bumps:Append Contributors for the important bits.
[modeling a full release](https://github.com/stephenking314/thunderbird-android/actions/runs/27154244459/job/80152915795)
[modeling a beta release](https://github.com/stephenking314/thunderbird-android/actions/runs/27165549002/job/80191669222)

## AI Disclosure

## Contribution Checklist

### DIFF
--- a/scripts/ci/contributor_list.py
+++ b/scripts/ci/contributor_list.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import subprocess
+import sys
+import requests
+import re
+
+
+def git(*args):
+    return subprocess.check_output(["git", *args], text=True).strip()
+
+
+def get_recent_matching_tags(pattern):
+    output = git(
+        "for-each-ref",
+        "--sort=-creatordate",
+        "--format=%(refname:short)",
+        "refs/tags",
+    )
+    filtered = [line for line in output.splitlines() if re.search(pattern, line)]
+    return filtered[:2]
+
+
+def github_get(url):
+    headers = {
+        "Accept": "application/vnd.github+json",
+        "User-Agent": "contributors-between-tags",
+    }
+    token = os.getenv("GITHUB_TOKEN")
+    if token:
+        headers["Authorization"] = f"Bearer {token}"
+
+    response = requests.get(url, headers=headers)
+    response.raise_for_status()
+
+    return response
+
+
+def github_compare(older, newer):
+    commits = []
+
+    url = f"https://api.github.com/repos/thunderbird/thunderbird-android/compare/{older}...{newer}?per_page=100"
+
+    while True:
+        response = github_get(url)
+        data = response.json()
+
+        page_commits = data.get("commits", [])
+        commits.extend(page_commits)
+
+        if "next" in response.links:
+            url = response.links["next"]["url"]
+        else:
+            break
+
+    return commits
+
+
+def get_contributors(commits):
+    contributors = {}
+    for commit in commits:
+        author = commit.get("author")
+
+        if not author or not author.get("login"):
+            continue
+
+        author = commit.get("author").get("login")
+        date = commit.get("commit").get("author").get("date")
+
+        # Track the earliest commit of each author in this set
+        if author not in contributors:
+            contributors[author] = commit
+        elif date < contributors[author].get("commit").get("author").get("date"):
+            contributors[author] = commit
+
+    return contributors
+
+
+def is_first_commit(author, date):
+    url = f"https://api.github.com/repos/thunderbird/thunderbird-android/commits?author={author}&until={date}&per_page=2"
+    data = github_get(url).json()
+    return len(data) == 1
+
+
+def get_first_contributions(contributors):
+    first_contributions = {}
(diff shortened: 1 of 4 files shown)


wmontwe:
@stephenking314 We're working on changing our in app changelog implementation. Please have a look at: [RFC](https://github.com/wmontwe/thunderbird-android/blob/b0af0256a18b5500ece59a6988bbaf00db4520f8/docs/engineering/rfcs/0001-changelog-system-replacement.md) and [technical design](https://github.com/wmontwe/thunderbird-android/blob/b0af0256a18b5500ece59a6988bbaf00db4520f8/docs/engineering/technical-designs/0001-changelog-system-replacement.md)

TLDR, we want to get rid of the XML format and use a JSON version defined in the above docs. It should be closer to `thunderbird-notes`, but as YAML support on Kotlin is weak, we will use JSON.

Feel free to comment.

stephenking314:
@wmontwe Those changes seem reasonable and it shouldn't be a big impact to the notes generation since we use an xml template file to generate the current changelog used in the app.  Switching over to use a json format should be simple, and we already have separate template files for github notes and the short form notes submitted to play store.
I'll give it a more thorough look over and leave any comments on the related issue.

Just a quick note about this PR, in case it isn't clear.  This addition is only for the [github release](https://github.com/thunderbird/thunderbird-android/releases/) notes.

scripts/ci/contributor_list.py
```diff
@@ -0,0 +1,185 @@
...
+
+        page_commits = data.get("commits", [])
+        commits.extend(page_commits)
+
+        if "next" in response.links:
+            url = response.links["next"]["url"]
+        else:
+            break
+
+    return commits
```
kryoseu: Probably not a big deal, but I believe the `/compare` API does not provide pagination. Instead the amount of commits is capped at 250, so this will always break at first iteration. 

A way to confirm this is by running a compare against an old and recent tag:

```
curl -v https://api.github.com/repos/thunderbird/thunderbird-android/compare/THUNDERBIRD_12_1...THUNDERBIRD_20_0b2 > /tmp/git-compare.json
```   

This has over 2.5k commits, but only 250 are returned.. No pagination.

scripts/ci/contributor_list.py
```diff
@@ -0,0 +1,185 @@
...
+        newer, older = tags[0], tags[1]
+        changelog_newer = newer
+
+    commits = github_compare(repo, older, newer)
+    contributors = get_contributors(commits)
+    first_contributions = get_first_contributions(repo, contributors)
+
+    contributors = {f"@{author}" for author in contributors}
+    excludes = {'@dependabot[bot]', '@thunderbird-botmobile[bot]', '@weblate'}
+    contributors -= excludes
```
kryoseu: Since we're excluding the bots after `get_first_contributions()`, is it possible we include bots as first contributors?

scripts/ci/contributor_list.py
```diff
@@ -0,0 +1,185 @@
...
+        print("\nContributors:", file=f)
+        thanks_to = "Thanks to: " + ', '.join(sorted(contributors, key=str.casefold))
+        print(thanks_to, file=f)
+
+        if first_contributions:
+            print("\nNew Contributors:", file=f)
+            for author, pr in first_contributions.items():
+                print(f"* @{author} made their first contribution in {pr}", file=f)
+
+        changelog = f"https://github.com/thunderbird/thunderbird-android/compare/{older}...{changelog_newer}"
```
kryoseu, line 145: nit: hardcoded repo instead of `{repo}`.

kryoseu (dismissed):
Added a few comments @stephenking314. None of those issues seem blocking, just some improvements worth noting.

scripts/ci/contributor_list.py
stephenking314: I was missing the hook to enable pagination, which is `?per_page=x`.  I actually had this earlier but accidentally dropped it while doing a refactor.  What you pointed out is correct, without pagination the results are capped at 250 commits, which could end up cutting out contributors between major revisions.

scripts/ci/contributor_list.py
stephenking314: Good call, might not be too likely we add a new bot, but if we do we'll want to be able to just add it to the list to exclude it.

scripts/ci/contributor_list.py
stephenking314: Yeah, that's cleaner.  If we want to fully generalize still can.  No need to pass around a parameter that will never change

kryoseu (approved):

wmontwe (approved):
LGTM

PR #11129, opened 2026-06-11
fix: printed header uses US date/time format rather than the device locale

### LINKED ISSUE #11109: Printing: Date not reflecting locales
### Checklist

### App

Thunderbird for Android

### App version

20.0b1

F-Droid

### Android version

16.0 / LineageOS 23.2

### Device model

instandnoodle

### Steps to reproduce

Printing a message.

### Expected behavior

Header should show the local date format (in my case European, DD.MM.YYYY HH:MM).

### Actual behavior

When printing, the header shows the US date format (M/D/YYYY H:MM AM/PM). I found no setting to format the header data.

### Logs

_No response_

mohitsatr:
Fixes #11109

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessagePrinter.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessagePrinter.kt
@@ -3,12 +3,11 @@ package com.fsck.k9.ui.messageview
 import android.content.Context
 import android.print.PrintAttributes
 import android.print.PrintManager
+import android.text.format.DateFormat
 import com.fsck.k9.mail.Address
 import com.fsck.k9.mail.Message
 import com.fsck.k9.mailstore.MessageViewInfo
 import com.fsck.k9.view.MessageWebView
-import java.text.SimpleDateFormat
-import java.util.Locale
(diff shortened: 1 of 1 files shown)


AvikMakwana:
Thanks for working on this.

I had also investigated #11109 locally and reached the same production fix: replacing the hardcoded `M/d/yyyy, h:mm a` print-header formatter with Android’s locale-aware `DateFormat.getDateFormat(context)` and `DateFormat.getTimeFormat(context)`.

I manually verified the behavior on a Motorola Edge 60 Fusion with 24-hour time enabled, using the same email in both builds:

- Thunderbird Beta 20.0b1: `Date: 6/11/26, 4:31 PM`
- Patched local build: `Date: 11/06/26, 16:31`

So this confirms the fix respects local date ordering and the 24-hour time preference.

I also have a small Robolectric regression test for a German region / 24-hour setup. Happy to share that patch if test coverage would be useful here.

dani-zilla (approved):
Thanks for the contribution! In the future, please fill out the full PR template. However, thank you for including the ticket and for fixing this for us!

antroyd:
Thanks for fixing it.
Which will be the version / beta this fix will be implemented first?

PR #11131, opened 2026-06-11
refactor: design system to kmp

### LINKED ISSUE #11116: Convert design-system to KMP
In order to extract Bolt, the design system module should be converted to KMP.

wmontwe:
## Contribution Summary

Resolves #11116 

#### Description

This converts the design system to KMP and resolves a couple of testing and dependency issues.

Downstream changes have been kept at a minimum to avoid impact on existing work.

The larger renames and reorganization steps will happen up-on the bold library extraction task.

## AI Disclosure

## Contribution Checklist


core/ui/compose/designsystem/src/commonMain/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeableRow.kt
```diff
@@ -161,28 +160,37 @@ private fun rememberAccessibilityActions(
...
+                template = stringResource(action.descriptionStringRes),
+                actionLabel = action.actionLabel(),
+            )
+
+            CustomAccessibilityAction(label = label) {
+                onSwipeEnd(direction)
+                true
             }
         }
     }
```
rafaeltonholo, line 179: Why have you removed the ` remember` block here?

core/ui/compose/designsystem/src/commonMain/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeableRow.kt
```diff
@@ -161,28 +160,37 @@ private fun rememberAccessibilityActions(
...
 }
 
+private fun formatAccessibilityActionLabel(
+    template: String,
+    actionLabel: String,
+): String {
+    return template
+        .replace("%1\$s", actionLabel)
+        .replace("%s", actionLabel)
+}
```
rafaeltonholo: `org.jetbrains.compose.resources.stringResource` also supports parameters, the only requirement is that they must be numbered (unless they fixed that.

This formatting function will only work in case we use string arguments, but will fail if we plan to use others, such as %d, %t, etc.

core/ui/compose/designsystem/src/commonMain/kotlin/app/k9mail/core/ui/compose/designsystem/atom/textfield/TextFieldOutlinedPassword.kt
```diff
@@ -39,7 +37,7 @@ fun TextFieldOutlinedPassword(
     Material3OutlinedTextField(
         value = value,
         onValueChange = stripLineBreaks(onValueChange),
-        modifier = modifier.applyLegacyPasswordSemantics().semantics { contentType = ContentType.Password },
+        modifier = modifier.applyPasswordSemantics(),
```
rafaeltonholo, line 40: Isn't the `.semantics { contentType = ContentType.Password }` required here?

wmontwe: `applyPasswordSemantics()` is an expect/actual method that applies the Android compat password semantics while always applying .semantics { contentType = ContentType.Password } for Android/JVM

core/ui/compose/designsystem/src/commonMain/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeableRow.kt
wmontwe: Right this is more a hack and instead we need to change the strings to use `%1$s`. As I anyway need to recreate the Weblate component (Android to composable resource change). It's probably fine to replace them in place, without introducing new keys.

core/ui/compose/designsystem/src/commonMain/kotlin/net/thunderbird/core/ui/compose/designsystem/molecule/swipe/SwipeableRow.kt
wmontwe: Within `remember` `stringResource` doesn't work as it's a composable function. The other `getString` are suspendable. The old setup doesn't play well with compose resources and I had no better idea as removing `remember`.

I used `runBlocking` for tests, but that doesn't seem suitable for production code.

I'm open for better ideas.

rafaeltonholo: What about:
```kotlin
@Composable
private fun rememberAccessibilityActions(
    state: SwipeableRowState,
    gesturesEnabled: Boolean,
    onSwipeEnd: (SwipeDirection) -> Unit,
): List<CustomAccessibilityAction> {
    if (!gesturesEnabled) return emptyList()
    val currentOnSwipeEnd by rememberUpdatedState(onSwipeEnd)
    val actions = state.accessibilityActions
    val enableEndToStart = state.enableSwipeFromEndToStart
    val enableStartToEnd = state.enableSwipeFromStartToEnd

    return actions.mapNotNull { action ->
        key(action) {
            val direction = when (action) {
                is EndToStartAccessibilityAction if enableEndToStart -> EndToStart
                is StartToEndAccessibilityAction if enableStartToEnd -> StartToEnd
                else -> return@key null
            }

            val label = stringResource(action.descriptionStringRes, action.actionLabel())
            remember(label, direction) {
                CustomAccessibilityAction(label = label) {
                    currentOnSwipeEnd(direction)
                    true
                }
            }
        }
    }
}

```

rafaeltonholo (approved):
LGTM. If the AccessibilityActions become a performance issue, we can revisit it later

PR #11167, opened 2026-06-19
chore(deps): bump detekt

wmontwe:
Follow-up of #11144

This bumps Detekt to [2.0.0-alpha.5](https://github.com/detekt/detekt/releases/tag/v2.0.0-alpha.5) and the detekt compose plugin to [0.6.2](https://github.com/mrmans0n/compose-rules/releases/tag/v0.6.2). See also [migration guide](https://detekt.dev/docs/introduction/migration/).

This resolves issues with AGP compatibility that require to use the alpha version. The benefit is that type resolution works now and calling `detektAll` also includes KMP. I didn't fix all the reported issues as this is a task for another time.

The detekt baseline path has been changed to allow me to move modules and keep the baseline with the module.


config/detekt/detekt.yml
```diff
@@ -89,27 +71,29 @@ comments:
...
 complexity:
   active: true
   CognitiveComplexMethod:
     active: false
-    threshold: 15
+    allowedComplexity: 15
   ComplexCondition:
     active: true
-    threshold: 5
+    allowedConditions: 3
```
rafaeltonholo, line 83: Was it intentional to change the value from 5 to 3?

config/detekt/detekt.yml
```diff
@@ -125,34 +109,31 @@ complexity:
...
     active: true
-    threshold: 600
+    allowedLines: 600
   LongMethod:
     active: true
-    threshold: 60
-    ignoreAnnotated:
-      - 'Preview'
-      - 'PreviewLightDark'
-      - 'PreviewLightDarkLandscape'
```
rafaeltonholo, line 135: Why are we removing these ignores?

config/detekt/detekt.yml
```diff
@@ -164,24 +145,28 @@ complexity:
   StringLiteralDuplication:
     active: false
     excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/androidUnitTest/**', '**/androidInstrumentedTest/**', '**/jsTest/**', '**/iosTest/**']
-    threshold: 3
+    allowedDuplications: 2
```
rafaeltonholo, line 148: Was it intentional decreasing the value from 3 to 2?

config/detekt/detekt.yml
```diff
@@ -164,24 +145,28 @@ complexity:
...
+    allowedFunctionsPerFile: 11
+    allowedFunctionsPerClass: 11
+    allowedFunctionsPerInterface: 11
+    allowedFunctionsPerObject: 11
+    allowedFunctionsPerEnum: 11
     ignoreDeprecated: false
     ignorePrivate: false
+    ignoreInternal: false
     ignoreOverridden: false
+    ignoreAnnotatedFunctions: ['Preview']
```
rafaeltonholo:
```suggestion
    ignoreAnnotatedFunctions: ['Preview', 'PreviewLightDark', 'PreviewLightDarkLandscape']
```

config/detekt/detekt.yml
```diff
@@ -164,24 +145,28 @@ complexity:
...
 
 coroutines:
   active: true
+  CoroutineLaunchedInTestWithoutRunTest:
+    active: false
```
rafaeltonholo: Shouldn't this be `true` instead?

config/detekt/detekt.yml
```diff
@@ -322,20 +313,24 @@ naming:
...
     active: false
+    aliases: ['FunctionMinNameLength']
     minimumFunctionNameLength: 3
   FunctionNaming:
     active: true
-    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/androidUnitTest/**', '**/androidInstrumentedTest/**', '**/jsTest/**', '**/iosTest/**']
-    functionPattern: '[a-z][a-zA-Z0-9]*'
+    aliases: ['FunctionName']
+    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/androidUnitTest/**', '**/androidInstrumentedTest/**', '**/jsTest/**', '**/iosTest/**', '**/nativeTest/**']
+    functionPattern: '[a-zA-Z][a-zA-Z0-9]*'
```
rafaeltonholo: Is this change related to Compose? It this required?

config/detekt/detekt.yml
```diff
@@ -391,17 +402,21 @@ performance:
     active: true
   CouldBeSequence:
     active: false
-    threshold: 3
+    allowedOperations: 2
```
rafaeltonholo, line 405: Was it intentional decreasing the value from 3 to 2

config/detekt/detekt.yml
```diff
@@ -620,14 +679,10 @@ style:
...
     ignorePropertyDeclaration: false
-    ignoreLocalVariableDeclaration: false
+    ignoreLocalVariableDeclaration: true
     ignoreConstantDeclaration: true
     ignoreCompanionObjectPropertyDeclaration: true
     ignoreAnnotation: false
-    ignoreAnnotated:
-      - 'Preview'
-      - 'PreviewLightDark'
-      - 'PreviewLightDarkLandscape'
```
rafaeltonholo, line 630: I would keep this

feature/mail/message/list/api/src/main/kotlin/net/thunderbird/feature/mail/message/list/ui/component/molecule/MessageItemAvatarCircle.kt
```diff
@@ -53,11 +53,11 @@ fun MessageItemAvatarCircle(
     Box(
         modifier = modifier
             .clip(CircleShape)
-            .combinedClickable(enabled = enabled, onClick = onClick)
```
rafaeltonholo, line 56: Please, remove the comments on lines 48 up to 52 as this is not necessary anymore

config/detekt/detekt.yml
wmontwe: The rule is inactive and that's the default value from detekt.

config/detekt/detekt.yml
wmontwe: The rule is inactive and that's the default value from detekt.

config/detekt/detekt.yml
wmontwe: I applied the detekt default values where possible and where they didn't raise new violations, I simply kept them.

config/detekt/detekt.yml
wmontwe: I think it makes sense to enable this.

config/detekt/detekt.yml
wmontwe: It didn't raise a violation for me.

config/detekt/detekt.yml
wmontwe: Good catch, it's not needed as composables are excluded anyway.

config/detekt/detekt.yml
wmontwe: This is now covered by `ignoreLocalVariableDeclaration: true`

config/detekt/detekt.yml
wmontwe: I didn't see any violations, but it might be handy.

config/detekt/detekt.yml
rafaeltonholo: It seems now that detekt is ignoring the `debug` folder. If we ever plan to move the Previews outside the `debug` folder, this will be a common detekt issue.

rafaeltonholo (approved):
LGTM, thanks for these changes!

config/detekt/detekt.yml
wmontwe: This might fail with the design system changes. I'll adjust if needed.

PR #11180, opened 2026-06-23
docs: add RFC for rendering inline images in plain-text messages

### LINKED ISSUE #11179: Create an RFC for the Positional Inline Images feature
Create an RFC for the Positional Inline Images feature, which will allow us to achieve feature parity with Desktop for messages with inline images.

rafaeltonholo:
## Contribution Summary

Closes #11179. 

#### Description

Adds a new RFC for rendering inline images in plain-text messages using positional body placement.

## AI Disclosure

        **Disclaimer**: Use AI for proofreading.

### DIFF
--- a/docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
+++ b/docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
@@ -99,12 +99,12 @@ consistently to users.**
 ## Risks & Drawbacks
 
 - **False positives.** Broad inlining may render something a sender or user thought of as an attachment in the body. The
-  default-on preference mitigates this: users who dislike it can turn it off. Scoping to plain-text-only further limits
-  exposure.
+  risk is limited by requiring `Content-Disposition: inline`, supported image types, and scoping the first iteration to
+  plain-text messages without an HTML alternative.
 
-- **Behavioural divergence within Android.** Existing `cid:` inline images are hidden from the attachment list; named
-  positional inline images will be listed. Two inline images could be treated differently in the list depending on how
-  they were authored. This matches Desktop's actual behaviour, but is an internal inconsistency worth noting.
+- **Attachment-list noise.** Named inline images, such as logos or email signatures, may appear in the attachment UI. 
+  This increases clarity for saving user-visible images but can add noise on small screens. **The UI should rely on the
+  existing attachment presentation rather than adding a separate inline-image setting.**
 
 - **Large images.** Inlining a multi-megabyte photo affects body rendering cost. The chosen rendering path loads the
   image lazily (it is not embedded in the HTML document), which keeps this in line with existing `cid:` images.
--- a/docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
+++ b/docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
@@ -1,4 +1,4 @@
-# RFC 0003: Render inline images in plain-text messages (Apple Mail positional images)
+# RFC 0003: Render inline images in plain-text messages
 
 - Issue: [#11179](https://github.com/thunderbird/thunderbird-android/issues/11179)
 - Technical design: TBD
@@ -14,8 +14,9 @@ mirroring Desktop's `mail.inline_attachments`. Additionally, this new behaviour 
 
 ## Motivation
 
-Some senders, most notably Apple Mail composing in plain text, attach images *inline* by position rather than by HTML
-reference. A representative message is a `multipart/mixed` whose children are, in order:
+Some senders attach images *inline* by position rather than by HTML reference. Apple Mail composing in plain text is a
+common producer of this standards-based message shape, but **the behaviour is not Apple-specific**. A representative 
+message is a `multipart/mixed` whose children are, in order:
 
 ```
 multipart/mixed  
--- a/docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
+++ b/docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
@@ -79,8 +79,8 @@ consistently to users.**
 
 - **Strict, Apple-Mail-only heuristic.** Inline only images that exactly match the Apple Mail shape (explicit
   `Content-Disposition: inline`, sibling of text parts, no HTML). Lowest false-positive risk, but narrower than Desktop
-  and would still render some messages differently from desktop. **Rejected** in favour of Desktop parity, which the
-  preference makes safe to default on.
+  and would still render some standards-based messages differently from desktop. 
+  - **Rejected** in favour of handling the standards-based shape directly.
 
 - **Add a user-visible setting.** This would mirror Desktop's `mail.inline_attachments` preference and provide an 
   opt-out for users who prefer an attachment-only view. 
@@ -89,11 +89,12 @@ consistently to users.**
     correctly by default.
 
 - **Hide all inline images from the attachment list (body-only).** Consistent with how Android treats `cid:` inline
-  images today. **Rejected** because it removes the easy save affordance and diverges from Desktop, which lists named
-  inline images.  ***Note**: TfA supports long-press to save on inline images, but it isn't as explicit as on Desktop.*
+  images today. 
+  - **Rejected** because it removes the explicit save/discovery affordance. Thunderbird for Android supports
+    long-press save on inline images, but that is less visible than the attachment UI.
 
-- **Do nothing.** Leave these images as attachments. **Rejected:** It is a visible cross-product inconsistency on a
-  common message shape.
+- **Do nothing.** Leave these images as attachments. 
+  - **Rejected:** It is a visible cross-product inconsistency on a common message shape.
 
 ## Risks & Drawbacks
 
--- a/docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
+++ b/docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
@@ -1,0 +1,115 @@
+# RFC 0003: Render inline images in plain-text messages (Apple Mail positional images)
+
+- Issue: [#11179](https://github.com/thunderbird/thunderbird-android/issues/11179)
+- Technical design: TBD
+- Status: **Proposed**
+
+## Summary
+
+Display images in the message body when a message carries them as positionally placed parts rather than `cid:`
+-referenced HTML images. Today these render only as attachments. This brings Thunderbird for Android in line with
+Thunderbird Desktop, which displays such images inline. The behaviour is gated behind a new preference (default on),
+mirroring Desktop's `mail.inline_attachments`. Additionally, this new behaviour must be secured under a feature flag
+(`enable_inline_images_positional_body_message`)
+
+## Motivation
+
+Some senders, most notably Apple Mail composing in plain text, attach images *inline* by position rather than by HTML
+reference. A representative message is a `multipart/mixed` whose children are, in order:
+
+```
+multipart/mixed  
+├── text/plain   "Text before"  
+├── image/jpeg   Content-Disposition: inline; filename=DSC01740.jpeg   (no Content-ID)  
+└── text/plain   "Text after"  
+```
+
+There is no HTML part and no `cid:` reference. The sender intends that the image appear between the two text blocks.
+
+Currently, the app only inlines images that are referenced from an HTML body via `cid:` and have a matching
+`Content-ID`. An image like the one above satisfies neither condition, so it is shown solely as an attachment. The user
+sees an attachment card/entry instead of the picture the sender placed in the message.
+
+Thunderbird Desktop displays this image inline by default. The two products diverging on a common, real-world message
+shape is a usability gap and the same message to look the same.
+
+This is **not** a regression: the `cid:` inline path works correctly. It is a class of message that has never been
+supported on the app.
+
+## Proposal
+
+Render any images which contain the header `Content-Disposition: inline`, that a message places among its body parts
+inline, at its position, even when it has no `Content-ID`, and it is not referenced by any HTML.
+
+Concretely:
+
+1. When a message body contains an image part that is not already handled as a `cid:` attachment, display it inline in
+   the rendered body at the position it occupies in the MIME structure. This applies to supported image types only, and
+   images that contain the header `Content-Disposition: inline`; other media (PDF, video, ...) continue to be shown as a
+   regular attachment because the message body cannot render them.
+
+2. Introduce an Android equivalent of the desktop's `mail.inline_attachments` preference. When enabled (the default),
+   images display inline as described. When disabled, behaviour escape hatch if they prefer a compact, attachment-only
+   view, and matches the control Desktop already exposes.
+
+3. Following Desktop's rule, an inline image that has a filename is shown **both** in the body and as an attachment
+   chip (so it remains easy to find and save). An inline image with no filename is shown in the body only. This differs
+   from how Android currently hides `cid:` inline images from the attachment list; see Risks & Drawbacks.
+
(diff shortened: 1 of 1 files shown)


docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
```diff
@@ -0,0 +1,115 @@
...
+
+- Issue: [#11179](https://github.com/thunderbird/thunderbird-android/issues/11179)
+- Technical design: TBD
+- Status: **Proposed**
+
+## Summary
+
+Display images in the message body when a message carries them as positionally placed parts rather than `cid:`
+-referenced HTML images. Today these render only as attachments. This brings Thunderbird for Android in line with
+Thunderbird Desktop, which displays such images inline. The behaviour is gated behind a new preference (default on),
```
dani-zilla: Would this also apply to cid-referenced images from other email clients? It seems like it should, but I wanted to be sure. Also, would we want to do this as a per-account setting, as we do currently for whether or not remote images should always be displayed automatically, or just for all accounts? I lean towards the latter, and if we have complaints about the lack of a per-account setting, we could always add an override to shut it off on individual accounts.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
```diff
@@ -0,0 +1,115 @@
...
+1. When a message body contains an image part that is not already handled as a `cid:` attachment, display it inline in
+   the rendered body at the position it occupies in the MIME structure. This applies to supported image types only, and
+   images that contain the header `Content-Disposition: inline`; other media (PDF, video, ...) continue to be shown as a
+   regular attachment because the message body cannot render them.
+
+2. Introduce an Android equivalent of the desktop's `mail.inline_attachments` preference. When enabled (the default),
+   images display inline as described. When disabled, behaviour escape hatch if they prefer a compact, attachment-only
+   view, and matches the control Desktop already exposes.
+
+3. Following Desktop's rule, an inline image that has a filename is shown **both** in the body and as an attachment
```
dani-zilla: Would this mean also showing images from email signatures at the bottom to be downloaded? I think the reduced screen size of an Android device vs desktop does warrant slightly different behavior to match what mobile users expect. Someone may not expect precise parity with desktop. 

We could either have it as an option to always show attachments chips for images, defaulted to off. I think most mobile users who want to download an image know to press and hold on it, then select save image. Users can already do this, and I don't think we've had requests to add the image chips to the bottom. 

It might be a decision we have to work with design on.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
```diff
@@ -0,0 +1,115 @@
...
+- **Do nothing.** Leave these images as attachments. **Rejected:** It is a visible cross-product inconsistency on a
+  common message shape.
+
+## Risks & Drawbacks
+
+- **False positives.** Broad inlining may render something a sender or user thought of as an attachment in the body. The
+  default-on preference mitigates this: users who dislike it can turn it off. Scoping to plain-text-only further limits
+  exposure.
+
+- **Behavioural divergence within Android.** Existing `cid:` inline images are hidden from the attachment list; named
```
dani-zilla: I usually default to, "Users don't want to think about this." They don't want to think about whether or not the person who emailed them used Proton Mail, Apple Mail, Gmail, or, preferably, Thunderbird. I therefore think we should make it as similar as possible between the implementations.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
```diff
@@ -0,0 +1,115 @@
...
+## Open Questions
+
+- **HTML + loose images.** When a message has both an HTML alternative and loose inline images, should the loose images
+  be appended/inlined (as Desktop does), or should we trust the HTML and leave them as attachments? **Proposed:** out of
+  scope for this RFC, follow-up issue.
+
+- **Nested under `multipart/alternative`.** Positional images nested inside an alternative are an unusual shape. Exclude
+  or handle? **Proposed:** exclude.
+
+- **Reply/quote/forward.** Should positional inline images be reproduced when quoting a message (*Desktop has a separate
```
dani-zilla: One thing we could do is just list the file name or even just a text placeholder like, "<image>"

But yeah, I agree that this consideration would be out of scope right now. Getting it working and looking the same as any other inline attachment should be the only focus here.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
wmontwe: I think this proposal is standards-compliant:

- [RFC 2183, section 2.1](https://www.rfc-editor.org/rfc/rfc2183.html#section-2.1) says `inline` body parts should be displayed automatically and in occurrence order.
- [RFC 2046, section 5.1.3](https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1.3) defines `multipart/mixed` as independent parts bundled in a particular order.

Existing `cid:` images are HTML references to MIME parts, as described by [RFC 2392](https://www.rfc-editor.org/rfc/rfc2392.html).

So this is not Apple-specific; Apple Mail is just a common producer of this standards-based message shape.

I also found [RFC 2387 `multipart/related`](https://www.rfc-editor.org/rfc/rfc2387.html), which is worth considering for HTML plus related resources. That is distinct from this plain-text positional case, but relevant when deciding not to disturb existing `cid:`/HTML behavior.

Sorry for the info dump, email protocols are inherently complex, but we have to be knowledgable about them and keep as close to them to avoid surprises.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+There is no HTML part and no `cid:` reference. The sender intends that the image appear between the two text blocks.
+
+Currently, the app only inlines images that are referenced from an HTML body via `cid:` and have a matching
+`Content-ID`. An image like the one above satisfies neither condition, so it is shown solely as an attachment. The user
+sees an attachment card/entry instead of the picture the sender placed in the message.
+
+Thunderbird Desktop displays this image inline by default. The two products diverging on a common, real-world message
+shape is a usability gap and the same message to look the same.
```
wmontwe: Yes, there is no specific rule for how the image must be displayed, it's really a UX decision the client can decide on.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
```diff
@@ -0,0 +1,115 @@
...
+  they were authored. This matches Desktop's actual behaviour, but is an internal inconsistency worth noting.
+
+- **Large images.** Inlining a multi-megabyte photo affects body rendering cost. The chosen rendering path loads the
+  image lazily (it is not embedded in the HTML document), which keeps this in line with existing `cid:` images.
+
+## Open Questions
+
+- **HTML + loose images.** When a message has both an HTML alternative and loose inline images, should the loose images
+  be appended/inlined (as Desktop does), or should we trust the HTML and leave them as attachments? **Proposed:** out of
+  scope for this RFC, follow-up issue.
```
wmontwe: I support out of scope: Loose inline images without `cid` are complicated, appending them might duplicate content or show tracking assets. This requires real samples to decide if it's a viable path.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
```diff
@@ -0,0 +1,115 @@
...
+
+- **Nested under `multipart/alternative`.** Positional images nested inside an alternative are an unusual shape. Exclude
+  or handle? **Proposed:** exclude.
```
wmontwe: Exclude for the first iteration. `multipart/alternative` parts are alternatives, not content to render together, so we should pick the best supported alternative and ignore the others. This also avoids fragile behavior for malformed or deliberately misleading MIME structures, including spam and phishing messages.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
```diff
@@ -0,0 +1,115 @@
...
+
+- **Reply/quote/forward.** Should positional inline images be reproduced when quoting a message (*Desktop has a separate
+  preference for this*)? **Proposed:** out of scope for this RFC.
```
wmontwe: Out of scope: This RFC only changes message display. Reply, quote, and forward behavior should remain unchanged and be handled in a follow-up proposal if needed.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
```diff
@@ -0,0 +1,115 @@
...
+
+- **Preference placement and naming.** Where in Settings should the toggle live, and what should it be called?
+  (Implementation detail deferred to the technical design, but the user-facing label needs product/UX input.)
```
wmontwe: We could put it in the current `messages` section, under "Show inline images in plain text messages". But an ideal solution would be to allow users to change that setting contextually, while they are actually reading a message.

wmontwe (commented):
I think this is standard compliant, but it would help to align this with the RFCs.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
rafaeltonholo: @dani-zilla:

> Would this also apply to cid-referenced images from other email clients?

Not sure if I got this question, but no changes are planned to the `cid-`referenced images behaviour.

> Also, would we want to do this as a per-account setting, as we do currently for whether or not remote images should always be displayed automatically, or just for all accounts?

I don't see any reason for having the gate preference per account. I would keep it for all accounts for simplicity.

rafaeltonholo: > I think this proposal is standards-compliant:
> 
>     * [RFC 2183, section 2.1](https://www.rfc-editor.org/rfc/rfc2183.html#section-2.1) says `inline` body parts should be displayed automatically and in occurrence order.
> 
>     * [RFC 2046, section 5.1.3](https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1.3) defines `multipart/mixed` as independent parts bundled in a particular order.
> 
> 
> Existing `cid:` images are HTML references to MIME parts, as described by [RFC 2392](https://www.rfc-editor.org/rfc/rfc2392.html).
> 
> So this is not Apple-specific; Apple Mail is just a common producer of this standards-based message shape.
> 
> I also found [RFC 2387 `multipart/related`](https://www.rfc-editor.org/rfc/rfc2387.html), which is worth considering for HTML plus related resources. That is distinct from this plain-text positional case, but relevant when deciding not to disturb existing `cid:`/HTML behavior.
> 
> Sorry for the info dump, email protocols are inherently complex, but we have to be knowledgable about them and keep as close to them to avoid surprises.

If you think it is worth, I can either include the RFCs as part of the summary or in a separate section such as Resources.

Let me know your preference.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
rafaeltonholo: > Would this mean also showing images from email signatures at the bottom to be downloaded?

If it was inlined AND it has a filename, then yes.

>  I think the reduced screen size of an Android device vs desktop does warrant slightly different behavior to match what mobile users expect. Someone may not expect precise parity with desktop.

I may have used the wrong term by saying "parity," but my intention is not to create an identical copy of the feature; rather, I want to ensure the same functionality.

If the desktop version displays the image inline and allows users to download the image as an attachment, we should implement a similar approach for mobile, using the mobile UX. Having the image that is inlined with a filename inside the attachment list will not disrupt the current user experience, as this is how it operates today. The only difference is that the image will also be visible within the body of the message.

> We could either have it as an option to always show attachments chips for images, defaulted to off. I think most mobile users who want to download an image know to press and hold on it, then select save image. 

I would prefer keeping the implementation simpler if possible.

> Users can already do this (long-press to download), and I don't think we've had requests to add the image chips to the bottom.

I would say that is a hidden feature that users might not even know that it exists. I myself didn't know about this before working on this RFC.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
rafaeltonholo: > I usually default to, "Users don't want to think about this." They don't want to think about whether or not the person who emailed them used Proton Mail, Apple Mail, Gmail, or, preferably, Thunderbird. I therefore think we should make it as similar as possible between the implementations.

I'm unsure if I got this comment in this section. The point here is to point out the divergence on our Android app, not on others. I may need to make it clearer in the text.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
rafaeltonholo: > One thing we could do is just list the file name or even just a text placeholder like, ""

The problem with only forwarding the filename or a placeholder is when someone forwards an email that is composed of only an inline image. In that situation, the person would be forwarding a broken message, and the receiver won't be able to read it.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
rafaeltonholo: > But an ideal solution would be to allow users to change that setting contextually, while they are actually reading a message.

Would the contextual preference change the global preference value, or just temporarily for the given message they turned it on? I lean towards only for the given message.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
dani-zilla: I guess I was mostly wondering if we needed an option to enable this. If the sender meant to have the images inline, shouldn't they be displayed as such? Is there a particular use case where you would want specifically emails sent in this format displaying without inline images?

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
dani-zilla: It might be coming from iOS that I instinctively press and hold to download images. I don't see the harm in adding the chips at the bottom to increase clarity though 👍

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
dani-zilla: I suppose I just meant, should we really treat these as any different to the end user, whether they came from our currently supported inline methods or a new one? Does it make sense to make them display differently at all? Basically, regardless of the client of the sender formatting it a particular way, should we make the attachments display the same whether it was inlined with one method or the other?

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
wmontwe: Adding the RFC as references is good enough. 

I think @dani-zilla is right, we don't need the setting, especiall if we follow the path to remove settings from the app.

docs/engineering/rfcs/0003-render-inline-images-in-plain-text-messages.md
wmontwe: If we store it for the given message, this is limited to the device as there is no easy way to sync or export/import that setting.

But as commented above, we might not need the setting at all.

rafaeltonholo:
@wmontwe and @dani-zilla, thanks for all the comments in this RFC.

I've updated the RFC text to address all the comments. 

Let me know your thoughts!

wmontwe (approved):
LGTM

dani-zilla (approved):
Looks good. Thanks!

PR #11187, opened 2026-06-24
feat: add Script to Convert Historical Changelog Data from XML to JSON

### LINKED ISSUE #11186: Add XML-to-JSON Conversion Script for Historical Changelog Data
The main, beta, and release changelogs are currently loaded locally from `changelog_master.xml`. 
As part of the ongoing replacement of the legacy changelog library, this issue involves creating a script that reads the corresponding changelog XML file (`main`, `beta`, or `release`) and generates JSON output files, with one JSON file per version and an additional mapping JSON file that references all version-specific JSON files.

shamim-emon:
- fixes #11186


scripts/changelog/migrate_changelog_to_json.py
```diff
@@ -0,0 +1,238 @@
+#!/usr/bin/env python3
+
```
wmontwe, line 17: A comment on how to use the script would help, see `validate-pr-report-labels.sh` for an example.

scripts/changelog/migrate_changelog_to_json.py
wmontwe, line 1: You need to give the script execution rights by running `chmod +x scripts/changelog/migrate_changelog_to_json.py`

wmontwe (commented):
I've run this and found some corner cases, K-9 has a special case, where 5.200 has two release candidates 5.200-RC1 and 5.200-RC2. I think they could be joined into one release to avoid having to change the schema. Having this as logic is not worth the effort. So this could be done by hand. Just add a warning in the script when processing RC versions.

I think having the json schema would allow us to validate the output of this migration script. If you like doing it, otherwise I'm happy to help.

scripts/changelog/migrate_changelog_to_json.py
shamim-emon: Done

scripts/changelog/migrate_changelog_to_json.py
shamim-emon: Done

wmontwe (approved):
Thank you!

PR #11193, opened 2026-06-25
fix: thread counted as single message in confirmation dialog

### LINKED ISSUE #8223: Misleading confirmation dialog deleting messages that are part of a thread
### Checklist

### App version

8.0b1

Google Play

### Android version

LineageOS for microG

### Device model

Sony Xperia XZ1 Compact 'lilac'

### Steps to reproduce

1. Enable threaded view
2. Ensure you have some threaded messages
3. Go to  either Unified inbox or the Inbox for the account with the threaded eamils
4. Select a message that is part of a thread (i.e. has the number of thread emails .e.g. 4 in the Correspondent line)
5. Select the message - the top bar shows e.g '4 selected'
6. Select Delete 

### Expected behavior

I would expect to see a confirmation dialog saying 'Confirm thread deletion: Do you really want to delete all 4 messages in this thread?'

### Actual behavior

The confirmation dialog says: 'Confirm deletion: Do you really want to delete this message?', implying that only ***a single message*** will be deleted

### Logs

_No response_

mohitsatr:
Fixes #8223 

#### Description

- Display the correct count of messages in the confirmation dialog when trying to delete/spam a thread by either swiping or selecting it.
- MessageListFragment does not yet have swipe actions, so for now, `selectedMessagesCount` can be used to get the correct count.

#### Updated Behaviour

https://github.com/user-attachments/assets/7cf1b537-3718-4296-a7d4-353e1c50e80b

## AI Disclosure

## Contribution Checklist

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/LegacyMessageListFragment.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/LegacyMessageListFragment.kt
@@ -1008,11 +1008,11 @@ class LegacyMessageListFragment :
         changeSort(nextSortType)
     }
 
-    private fun onDelete(messages: List<MessageReference>) {
+    private fun onDelete(messages: List<MessageReference>, count: Int) {
         if (interactionSettings.isConfirmDelete) {
             // remember the message selection for #onCreateDialog(int)
             activeMessages = messages
-            showDialog(R.id.dialog_confirm_delete)
+            showDialog(R.id.dialog_confirm_delete, count)
         } else {
             onDeleteConfirmed(messages)
         }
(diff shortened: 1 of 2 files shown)


legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/LegacyMessageListFragment.kt
```diff
@@ -1117,6 +1091,41 @@ class LegacyMessageListFragment :
...
+                    selectionCount,
+                    selectionCount,
+                )
+                val confirmText = getString(R.string.dialog_confirm_delete_confirm_button)
+                val cancelText = getString(R.string.dialog_confirm_delete_cancel_button)
+                ConfirmationDialogFragment.newInstance(dialogId, title, message, confirmText, cancelText)
+            }
+
+            else -> {
+                throw RuntimeException("Called showDialog(int) with unknown dialog id.")
```
dani-zilla: This started off as a bit of a nit, that you should probably specify this is the `showDialog(int, int)` version of this function, as the other is capable of processing different dialog IDs. 

But that got me thinking, how would someone know which one to call and would they know to test all cases prior to submitting code? Perhaps these should be one function, with the secondary int for the number of items being optional (defaulting to 1 or making it nullable), and, if no value is passed in, handle it like you would without the second parameter, otherwise, handle it as you do for deleting threads. I just want to account for a future contributor who might not be sure which one to pick.

dani-zilla (changes requested):

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/LegacyMessageListFragment.kt
mohitsatr: Made it one function and `selectionCount` parameter defaults to 1

dani-zilla (approved):

PR #11212, opened 2026-06-29
fix: allow scrolling the notification actions list on small screens

### LINKED ISSUE #11191: Add a way to scroll on notification actions page on small displays
Currently, it's incredibly difficult, if not impossible, to select notification actions on a small screen if the desired option is below the displayed options. This is because the same gesture for selecting items would change the location of the notification actions instead of scrolling. 

To fix this, we'll define "grab areas" to grab each item, similar to the way we allow reordering of accounts in settings. This makes the rest of each item a scroll surface. 

Alternatively, if all items cannot display, we could turn off the gesture so only the arrow buttons on the side can reorder items. 

Image

Something like this may be preferable: 
Image

AC:
- Users can scroll on small displays to bring all items into view for reordering notification actions

mvanhorn:
### Prior to submitting a pull request, please familiarize yourself with...

- Mozilla's [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/)
- The repository [contribution code quality guides](https://github.com/thunderbird/thunderbird-android/tree/main/docs/contributing)

Linked Issue/Ticket: Fixes #11191

#### Description

On the Notification actions reorder screen, every row attached a full-row drag gesture. `NotificationReorderRow` in `NotificationActionsSettingsScreen.kt` put `.immediateDragGesture(...)` on the root `Row`, so any vertical drag anywhere on a row was consumed as a reorder drag. When the action list is taller than the screen, that left no way to scroll the `LazyColumn`, and actions below the fold were unreachable on small displays.

This moves the gesture off the row and onto a dedicated leading drag handle (`NotificationDragHandle`). With the gesture confined to the handle, vertical drags on the rest of a row fall through to the `LazyColumn` and scroll the list, while dragging the handle still reorders the item. This mirrors how account reordering already exposes a grab area in settings.

The reorder logic in `NotificationActionsReorderController` is unchanged. The existing up/down arrow buttons remain the accessible reordering path, so the handle is marked decorative for screen readers (`clearAndSetSemantics`) to avoid exposing a repeated, non-actionable element to TalkBack.

#### Screen Shots

This is a touch-gesture and behavior change. A short video of dragging the handle versus scrolling the row body would best show it; I was not able to capture device media in this environment.

## AI Disclosure

### Please ensure that your pull request meets the following requirements - thanks!

- [ ] This contribution does not break existing unit tests (run `gradlew testDebugUnitTest`; will be checked by CI).
- [ ] This contribution adheres to our [Engineering process](https://github.com/thunderbird/thunderbird-android/tree/main/docs/engineering) (RFC/Technical Design/ADR)

AI was used for assistance.


dani-zilla:
So, this is in line with what I was thinking. And it does re-enable scrolling on small devices. However, I do think the divider line is a bit inelegant here. We could either allow just the line to be dragged or fix the handle on it. 

I also wonder if we wouldn't be better off only showing these if the display size is too small to display all the content. That could be out of scope, and we could look at it for a future improvement. Probably best for accessibility and to re-add this feature to small displays to just get the handles aligned right and working for now. 


https://github.com/user-attachments/assets/956dd29c-15b8-4a2c-b851-6f308c367249

notificationActionsSettingsWithHandles

dani-zilla:
Just checking in, are you still interested in coming back to this? We have a few solutions in mind, but I'm mostly leaning towards using drag handles on all the options as you started with here, and simply removing dragging capability from the divider bar. What do you think?

mvanhorn:
Still interested, yes. Drag handles on all options plus removing drag from the divider bar sounds right to me - it keeps one obvious affordance for reordering and stops the divider from swallowing scroll gestures, which was the original bug. I'll update the PR to that approach.

mvanhorn:
Thanks for the steer, @dani-zilla - went with your suggestion:

- Every action row now carries a drag handle; only the handle initiates a reorder, so the row and the list scroll freely on small screens (the original #11191 bug).
- The divider bar is a static line now, with no drag gesture.

I also merged `main`, which had migrated this screen to `BoltTheme` in the meantime.

Two heads-ups: (1) it's a fork PR, so the CI build needs your approval to run - I haven't been able to compile it on my end. (2) I don't have an Android emulator, so I couldn't check the drag feel visually; a sanity-check of the interaction would be welcome once CI is green.

dani-zilla (approved):
Thanks for updating that

Issue #11218, opened 2026-06-29
uplift(release): fix(webview): prevent crash by disabling webview hardware layer


wmontwe:
Uplift to release: #11149

The issue #11148 was tested on beta 21.0b1 and is fixed there.

PR #11219, opened 2026-06-29
uplift(release): fix(webview): prevent crash by disabling webview hardware layer (#11149)

wmontwe:
Resovles #11218 

The fix works for beta 21.0b1 and is ready to be uplifted to release.


rafaeltonholo (approved):
LGTM

PR #11223, opened 2026-06-29
docs: add RFC for new feature flag architecture

### LINKED ISSUE #11222: Feature Flag - Create an RFC for the new Feature Flag architecture
Create an RFC for the new Feature Flag architecture. This RFC must be focused on addressing:

1. The multiple feature flag definitions
2. Creating a centralized JSON catalog

rafaeltonholo:
## Contribution Summary

Closes #11222 

#### Description

- Add an RFC proposing a new feature flag architecture and file schema

## AI Disclosure

   **Disclaimer**: Use AI for proofreading.

### DIFF
--- a/docs/engineering/rfcs/0004-feature-flag-new-architecture.md
+++ b/docs/engineering/rfcs/0004-feature-flag-new-architecture.md
@@ -1,0 +1,242 @@
+# RFC 0004: Adopt OpenFeature for the Feature Flag System
+
+- Issue: [#11222](https://github.com/thunderbird/thunderbird-android/issues/11222)
+- Technical design: [OpenFeature-based Feature Flag Provider]() TBD
+- Status: **Proposed**
+
+## Summary
+
+Adopt the [OpenFeature](https://openfeature.dev) standard as the feature-flag *evaluation layer* for both apps (
+Thunderbird and K-9 Mail) across all build_types. Replace the six per-app/per-built type `FeatureFlagFactory` classes
+with
+a single declarative **flag catalog**, and introduce a remotely-updatable, cacheable provider that supports staleness
+checks and (later) gradual rollout and targeting.
+
+Crucially, the existing `FeatureFlagProvider` / `FeatureFlagResult` API is **kept as-is**, so the ~30 call sites do not
+change; OpenFeature sits behind that facade as an implementation detail.
+
+## Motivation
+
+Feature flags today are resolved **at build time**: each app per build_type combination ships its own
+`FeatureFlagFactory` (`TbFeatureFlagFactory` for debug/daily/beta/release, `K9FeatureFlagFactory` for debug/release)
+that hardcodes a list of booleans. This has real limitations:
+
+- **No remote control.** A flag cannot be changed without shipping a new build; there is no remote definition, no cache,
+  and no way to check whether an update is needed.
+- **Duplication and drift.** A single flag is declared in up to six files; values diverge by accident, and only 3 of 12
+  flags actually need to differ across variants.
+- **Stringly-typed keys.** Some keys are raw strings; the registry is scattered, with a "do not add flags here" comment
+  steering contributors toward per-feature `:api` modules.
+- **No path to rollout/targeting.** There is no mechanism for percentage rollouts, cohorts, or A/B (split-testing)
+  experiments.
+
+We need a system that, across two apps and their build_types, can ship a default flag definition, fetch and cache a
+remote
+definition, detect when an update is needed, and leave room for gradual rollout and targeting, **without** locking into
+a vendor or rewriting every call site.
+
+## Proposal
+
+### The Schema
+
+```json
+{
+    "$schema": "featureflag.schema.lean.json",
+    "version": "2026-06-18.1",
+    "context": {
+        "targeting_key": {
+            "source": "random_install_id",
+            "description": "Stable random per-install identifier for rollout/experiment bucketing. NOT a user identifier."
+        },
+        "attributes": [
+            {
+                "name": "app",
+                "type": "string",
+                "source": "BuildConfig (app module)",
+                "values": [
+                    "thunderbird",
+                    "k9"
+                ]
+            },
+            {
+                "name": "build_type",
+                "type": "string",
+                "source": "BuildConfig.BUILD_TYPE",
+                "values": [
+                    "debug",
+                    "daily",
+                    "beta",
+                    "release"
+                ]
+            },
+            {
+                "name": "variant",
+                "type": "string",
+                "derived": "{app}-{build_type}"
+            },
+            {
+                "name": "app_version",
+                "type": "string",
+                "source": "BuildConfig.VERSION_NAME"
+            },
+            {
+                "name": "os_sdk_int",
+                "type": "int",
+                "source": "Build.VERSION.SDK_INT"
+            }
+        ]
+    },
+    "flags": [
+        {
+            "key": "archive_marks_as_read",
+            "default": true,
+            "description": "Marks message as read when archived."
+        },
+        {
+            "key": "disable_font_size_config",
+            "default": true,
+            "description": "Disables legacy font-size configuration UI."
+        },
+        {
+            "key": "email_notification_default",
+            "default": true,
+            "description": "Default email-notification behaviour for new accounts."
+        },
+        ...
+    ],
+    "overrides": {
+        "thunderbird": {
+            "debug": {
+                "display_in_app_notifications": true,
+                "use_notification_sender_for_system_notifications": true,
+                "message_view_action_export_eml": true
+            },
+            "daily": {
+                ...
+            },
+            "beta": {
+                ...
+            },
+            "release": {
+                ...
+            }
+        },
+        "k9": {
+            "debug": {
+                "message_view_action_export_eml": true
+            },
+            "release": {
+                ...
+            }
+        }
+    }
+}
+```
+
+Where:
+
+- `context`: the **evaluation-context contract**: the inputs the app must supply on every flag evaluation. It is both
+  documentation and the codegen source for a typed context builder, so call sites cannot pass an undeclared attribute.
+  - `targeting_key`: the bucketing key for gradual rollout and experiments. It is sourced from a stable, random
+    per-install identifier; it is **pseudonymous, not a user identifier**, and under on-device evaluation it never
+    leaves the device.
+  - `attributes`: the targeting dimensions available to overrides and (later) rollout rules. Each declares a `name`, a
+    `type`, and either where it comes from (`source`) or how it is computed (`derived`):
+    - `app` / `build_type`: the former build-time matrix, now supplied as runtime context from `BuildConfig`. These
+      two replace the six source sets.
+    - `variant`: derived as `{app}-{build_type}`; a convenience key for override lookup and rule authoring.
+    - `app_version` / `os_sdk_int`: version and OS dimensions reserved for future targeting (e.g. min-SDK gating or
+      a staged rollout by app version).
+- `flags`: the **flag registry**: the single source of truth for which flags exist. Each entry carries a `key` (the
+  source for typed-key codegen), a `default` (the baseline value shipped with the app and returned when no override or
+  remote value applies), and a `description`. Flags are boolean for now; the shape is designed to extend to multivariate
+  and rollout later.
+- `overrides`: per-`app` -> per-`build_type` deviations from a flag's `default`. Only the few flags that genuinely
+  differ
+  across variants appear here (today, 3 of 12), which removes the duplication where a single flag was declared in up to
+  six files. A flag absent from an override block simply inherits its `default`. This is the offline equivalent of
+  today's per-factory overrides, and it is the layer the bundled-catalog provider resolves against.
+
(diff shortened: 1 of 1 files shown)


dani-zilla (dismissed):

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
```diff
@@ -0,0 +1,242 @@
...
+}
+```
+
+Where:
+
+- `context`: the **evaluation-context contract**: the inputs the app must supply on every flag evaluation. It is both
+  documentation and the codegen source for a typed context builder, so call sites cannot pass an undeclared attribute.
+  - `targeting_key`: the bucketing key for gradual rollout and experiments. It is sourced from a stable, random
+    per-install identifier; it is **pseudonymous, not a user identifier**, and under on-device evaluation it never
+    leaves the device.
```
wmontwe: The rollout key is a locally generated random per-install value. Although pseudonymous, it can become identifying when combined with app version, OS version, IP address, request timing, or other service-side metadata. The implementation must define how the key is generated, stored, reset, rotated, transmitted, logged, and deleted. If any remote service receives or stores the key or derived bucket identifiers, that service must follow equivalent privacy requirements, including retention limits, access controls, metadata minimization, and a prohibition on logging raw identifiers.

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
```diff
@@ -0,0 +1,242 @@
...
+  remote layer is our code regardless of in-house vs. vendor.
+- **Remote configuration can change shipped behaviour.** Requires integrity verification (signing) and a security review
+  before the remote layer ships.
+- **`targeting_key` is a pseudonymous identifier.** Rollouts carry privacy implications; the posture depends on whether
+  evaluation is local or remote.
+- **Added dependency and concepts.** One more `commonMain` dependency and new concepts (providers, evaluation context,
+  events) for contributors.
+- **`Unavailable` is not native** to the simple value API; it must be recovered from evaluation details.
+
+## Open Questions
```
wmontwe, line 191: Other question:

Should the RFC be split into two decisions: offline declarative catalog first, remote provider later after security/privacy design?

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
```diff
@@ -0,0 +1,242 @@
...
+
+- **Online layer:** in-house flagd-compatible evaluator, or a managed vendor (wrapped in a custom Kotlin provider)?
```
wmontwe: Start with an in-house local evaluator and defer any managed vendor. Vendor choice needs separate privacy/security review.

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
```diff
@@ -0,0 +1,242 @@
...
+- **Hosting & integrity:** where is the remote catalog served, and must it be signed before it can flip behaviour?
```
wmontwe: Remote catalogs must be signed before they can change behavior. Unsigned, invalid, expired, or malformed catalogs fall back to bundled defaults.

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
```diff
@@ -0,0 +1,242 @@
...
+- **Evaluation location:** local/in-process (privacy-preserving) vs. remote evaluation (transmits the bucketing key)?
```
wmontwe: Local/in-process only for initial scope. Do not send `targeting_key` or full evaluation context off-device until a separate privacy design is accepted (user consent).

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
```diff
@@ -0,0 +1,242 @@
...
+- **Catalog shape:** a single catalog with `app` targeting, or one catalog per app?
```
wmontwe: Use one catalog with app and build_type targeting. This reduces duplication.

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
```diff
@@ -0,0 +1,242 @@
...
+- **Scope:** is multivariate (non-boolean) needed in the near term, or deferred?
```
wmontwe: Boolean flags only for now. Defer multivariate flags until there is a concrete use case.

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
```diff
@@ -0,0 +1,242 @@
...
+- **Context contract placement:** the `context` block declares the *shape* of the evaluation context (attribute names,
+  types, sources) — a build-time concern consumed by codegen, rule validation, and documentation. In OpenFeature the
+  evaluation context itself is runtime data supplied per evaluation (global/client/invocation) and is never part of a
+  flag definition; the contract is also tied to the app version (the app must be rebuilt to source a new attribute), so
+  a remotely-fetched catalog cannot act on it. Should the contract therefore move out of the runtime catalog — into the
+  JSON Schema (`featureflag.schema.lean.json`) or a clearly-labelled build-time section — and be renamed off `context`
+  (e.g. `evaluationContextContract`) to avoid colliding with OpenFeature's `EvaluationContext`? Or does keeping it
+  inline as the codegen source outweigh shipping it in the resolved payload?
```
wmontwe: Move it out of the runtime catalog or rename it clearly to `evaluationContextContract`. Do not call it `context`, because that collides with OpenFeature runtime `EvaluationContext`.

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
```diff
@@ -0,0 +1,242 @@
...
+      a staged rollout by app version).
+- `flags`: the **flag registry**: the single source of truth for which flags exist. Each entry carries a `key` (the
+  source for typed-key codegen), a `default` (the baseline value shipped with the app and returned when no override or
+  remote value applies), and a `description`. Flags are boolean for now; the shape is designed to extend to multivariate
+  and rollout later.
+- `overrides`: per-`app` -> per-`build_type` deviations from a flag's `default`. Only the few flags that genuinely
+  differ
+  across variants appear here (today, 3 of 12), which removes the duplication where a single flag was declared in up to
+  six files. A flag absent from an override block simply inherits its `default`. This is the offline equivalent of
+  today's per-factory overrides, and it is the layer the bundled-catalog provider resolves against.
```
wmontwe: In this case they are `variantOverrides` to distinguish app/build defaults from local debug overrides. Define precedence explicitly: local debug override, remote value, bundled variant override, bundled default, and Unavailable only when the key is not present in the bundled catalog.

wmontwe (commented):
This feels like two decisions bundled into one RFC: deduplicating local feature flag definitions, and introducing a remotely updatable OpenFeature-based rollout system.

The concrete problems described here are relatively focused:
- six duplicated factory files
- drift between app/build variants
- some stringly typed keys
- hardcoded defaults

Those seem solvable with a smaller first step:
- one bundled declarative catalog
- generated/validated keys
- variant defaults per app/build type
- existing FeatureFlagProvider facade unchanged
- existing debug overrides unchanged

That would reduce current maintenance cost while keeping the implementation aligned with the future remote model. The remote/OpenFeature layer can then build on the same catalog in a second phase, once signing, hosting, staleness handling, targeting identifiers, privacy review, and service-side data handling rules are defined.

I’d suggest the RFC explicitly separate this into phases:
1. Local declarative catalog and validation as the quick win.
2. Remote catalog/evaluation after privacy, security, and operational requirements are agreed.

rafaeltonholo:
> This feels like two decisions bundled into one RFC: deduplicating local feature flag definitions, and introducing a remotely updatable OpenFeature-based rollout system.
> 
> The concrete problems described here are relatively focused:
> 
>     * six duplicated factory files
> 
>     * drift between app/build variants
> 
>     * some stringly typed keys
> 
>     * hardcoded defaults
> 
> 
> Those seem solvable with a smaller first step:
> 
>     * one bundled declarative catalog
> 
>     * generated/validated keys
> 
>     * variant defaults per app/build type
> 
>     * existing FeatureFlagProvider facade unchanged
> 
>     * existing debug overrides unchanged
> 
> 
> That would reduce current maintenance cost while keeping the implementation aligned with the future remote model. The remote/OpenFeature layer can then build on the same catalog in a second phase, once signing, hosting, staleness handling, targeting identifiers, privacy review, and service-side data handling rules are defined.
> 
> I’d suggest the RFC explicitly separate this into phases:
> 
>     1. Local declarative catalog and validation as the quick win.
> 
>     2. Remote catalog/evaluation after privacy, security, and operational requirements are agreed.

Sure, let's split it then. I will remove the part of the OpenFeature/remote integration from this RFC

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
rafaeltonholo: As this is more related to the remote/service implementation of the feature flag, let's keep this for the second phase of the feature flag.

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
rafaeltonholo: That is actually a good idea. What do you think, @dani-zilla?

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
rafaeltonholo: Good idea. However, if we plan to split this RFC into two parts, the in-house local evaluator would be part of the second RFC

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
rafaeltonholo: Noted. Deferred to the second phase of the feature flag, though.

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
rafaeltonholo: I lean towards removing it from the catalog. It is good to have it documented, as it will be used inside the app's logic to build it, but we don't need to transmit it over the network since the evaluation will run in-local process.

docs/engineering/rfcs/0004-feature-flag-new-architecture.md
rafaeltonholo: Not sure if I got what you meant. What would be the benefit of having both `override` and `variantOverrides` in the same JSON catalog?

The override precedence is defined inside the app, and I don't see a need to have all the possible overrides in the catalog.

Thinking about how the app will understand what flag to pick: OpenFeature provide a `MultiProvider` class that we can use to define many providers. Then, we will have:

1. RuntimeDebugOverrideFeatureFlagProvider - Keeps track of the local debug overrides at runtime
2. RemoteFeatureFlagProvider - Keeps track of the remote feature flag catalog and its overrides
3. BundledCatalogFeatureFlagProvider - Keeps track of the bundled feature flag catalog and its overrides

Then, the evaluation would run somewhat like:

```mermaid
flowchart TD
    FFP["FeatureFlag.provide(flag_key)"]
    EV[Evaluator]
    BFF(BundledCatalogFeatureFlagProvider)
    RFF(RemoteCatalogFeatureFlagProvider)
    DFF(DebugOverrideFeatureFlagProvider)
    END["feature_flag[flag_key]"]
    MSS["Missing flag"]

    FFP --> |Get 'flag_key' value| EV
    EV --> |Start evaluation| DFF

        DFF --> DFF_EVAL_1{'flag_key' present in overrides}
        DFF_EVAL_1 --> |Yes| END
        DFF_EVAL_1 --> |No| RFF


        RFF --> RFF_EVAL_1{'flag_key' present in overrides}
        RFF_EVAL_1 --> |Yes| END
        RFF_EVAL_1 --> |No| RFF_EVAL_2{'flag_key' present in defaults}
        RFF_EVAL_2 --> |Yes| END
        RFF_EVAL_2 --> |No| BFF

        BFF --> BFF_EVAL_1{'flag_key' present in overrides}
        BFF_EVAL_1 --> |Yes| END
        BFF_EVAL_1 --> |No| BFF_EVAL_2{'flag_key' present in defaults}
        BFF_EVAL_2 --> |Yes| END
        BFF_EVAL_2 --> |No| MSS

```

With that, we always assume that:
1. The local `overrides` will take the first precedence
2. Then, the remote overrides/defaults
3. Then the bundled overrides/defaults.

Am I following your thoughts?

wmontwe: Maybe, I've read it wrongly, I didn't want them in the catalog, just define the hirarchy and this sounds good:

> 1. The local overrides will take the first precedence
> 2. Then, the remote overrides/defaults
> 3. Then the bundled overrides/defaults.

rafaeltonholo:
@wmontwe and @dani-zilla, I've addressed the comments and removed the OpenFeature/remote integration part - still mentioning it, but leaving it for a follow-up RFC.

Please let me know your thoughts.

dani-zilla (approved):

wmontwe (approved):
LGTM, thank you!

PR #11225, opened 2026-06-30
fix: prevent horizontal scrolling in the message viewport

### LINKED ISSUE #11038: Viewport always allows horizontal scrolling
### Checklist

### App

K-9 Mail

### App version

18

F-Droid

### Android version

16

### Device model

_No response_

### Steps to reproduce

When I load a message, K9 allows scrolling light to the left/right.  Thus the view port is slightly wider, than the screen.  This started few months ago, let’s say approximately two major versions ago, before that scrolling left/right was not possible. (There might have been such cases with very long URLs which were not wrapped on new line, but these minor cases are a different story.)

This is a regression, as it started few months ago.

### Expected behavior

K9 should not allow horizontal scrolling, when it is not required to render the message.

### Actual behavior

K9 allows horizontal scrolling, even for this message:
```
Content-Type: text/plain; charset="UTF-8"

***** Nagios 1.0 *****

Notification Type: PROBLEM

Service: LOAD
Host: A (B)
Address: a.b.c
State: CRITICAL

Date/Time: 23-05-2026 11:30:12

Additional Info:

CRITICAL - load average: 14.52, 9.16, 5.81
```
Below the horizontal and vertical scrollbars can be seen.  I have hard times to create a screenshot with the scrollbars visible.

Image

### Logs

_No response_

mvanhorn:
## Contribution Summary

Linked Issue/Ticket: Fixes #11038
RFC / Technical Design (if applicable): N/A

#### Description

The message view allows a few pixels of horizontal scrolling: the rendered content ends up slightly wider than the screen and can be dragged left/right. As #11038 notes, ordinary messages were not horizontally scrollable before this regressed.

The cause is in the new message reader stylesheet. The main content rule sets `width: 100%` together with `padding: 0 8px`, but without `box-sizing: border-box`:

```css
.root.main-content {
  width: 100%;
  overflow-wrap: break-word;
  padding: 0 8px;
}
```

Under the default `content-box` model the 8px left/right padding is added on top of the full 100% width, so the body becomes 16px wider than the viewport, which is the small horizontal overflow the issue describes. (This also rules out the font-size theory raised on the thread: the reporter confirmed font size is at 100% and the overflow persists.)

Adding `box-sizing: border-box` to that rule makes the padding count inside the 100% width instead of being added to it, so the content matches the viewport width again. The change is confined to `DefaultGlobalCssStyleProvider` (the current default path); the legacy provider does not emit this rule and is left untouched. The existing `width: 100%`, `overflow-wrap`, and `padding: 0 8px` declarations are preserved, so the 8px gutter is still rendered, now within the viewport.

#### Screen Shots

This is a CSS box-model fix that removes a few pixels of horizontal overscroll; the difference is the absence of left/right drag rather than a visible layout change, so it is not meaningfully captured in a static screenshot. The behavior is covered instead by a unit test in `DefaultGlobalCssStyleProviderTest`.

## AI Disclosure

## Contribution Checklist

- [ ] This contribution adheres to our [Engineering process](https://github.com/thunderbird/thunderbird-android/tree/main/docs/engineering) (RFC/Technical Design/ADR)

### DIFF
--- a/feature/mail/message/reader/impl/src/androidHostTest/kotlin/net/thunderbird/feature/mail/message/reader/impl/css/DefaultGlobalCssStyleProviderTest.kt
+++ b/feature/mail/message/reader/impl/src/androidHostTest/kotlin/net/thunderbird/feature/mail/message/reader/impl/css/DefaultGlobalCssStyleProviderTest.kt
@@ -1,0 +1,80 @@
+package net.thunderbird.feature.mail.message.reader.impl.css
+
+import assertk.assertThat
+import assertk.assertions.contains
+import kotlin.test.Test
+import net.thunderbird.core.common.mail.html.HtmlSettings
+import net.thunderbird.feature.mail.message.reader.api.css.CssClassNameProvider
+import net.thunderbird.feature.mail.message.reader.api.css.CssVariableNameProvider
+
+class DefaultGlobalCssStyleProviderTest {
+
+    @Test
+    fun `style should set main content box sizing to border box`() {
+        // Arrange
+        val testSubject = createTestSubject()
+
+        // Act
+        val style = testSubject.create(createHtmlSettings()).style
+
+        // Assert
+        assertThat(style).contains("box-sizing: border-box")
+    }
+
+    @Test
+    fun `style should preserve main content width and padding`() {
+        // Arrange
+        val testSubject = createTestSubject()
+
+        // Act
+        val style = testSubject.create(createHtmlSettings()).style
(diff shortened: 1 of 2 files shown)


wmontwe (approved):
Thank you for fixing this!

dilyanpalauzov:
With K9 22.0 the problem is still present.  As shown in the image at https://github.com/thunderbird/thunderbird-android/issues/11038 there is horizontal and vertical scrollbar.

PR #11263, opened 2026-07-16
fix: prevent TransactionTooLargeException when attaching a file while replying to a message with a large HTML quote

### LINKED ISSUE #9425: Error when attaching file or image when we reply
### Checklist

- [ ] I have used the search function to see if someone else has already submitted the same bug report.

### App

Thunderbird for Android

### App version

latest version 10.0

Google Play

### Android version

15

### Device model

Many device, Xiaomi 13, Xiaomi 13T Pro

### Steps to reproduce

Hi,

When we try to attach some file or image when we reply to a message,
This is a old error that occurs in K9-email:

The erro ccours whe we select the image or file.

######################################################
java.lang.RuntimeException:
android.os.TransactionTooLargeException: data
parcel size 1164924 bytes
Bundle stats:
android:viewHierarchyState [size=25820]
android:views [size=25716]
0x7f0902d8 [size=16960]
0x7f0903b1 [size=1780]
androidx.lifecycle.BundlableSavedStateRegistry.key
[size=1056]
state:htmlQuote [size=1134364]
PersistableBundle stats:
[null]
at
android.app.servertransaction.PendingTransactionAct
ions$Stoplnfo.run(PendingTransactionActions.java:14
6)
at
android.os.Handler.handleCallback(Handler.java:959)
at
android.os.Handler.dispatchMessage(Handler.java:10
at android.os.Looper.loopOnce(Looper.java:249)
at android.os.Looper.loop(Looper.java:337)
at
android.app.ActivityThread.main(ActivityThread.java:9
403)
at
com.android.internal.os.Runtimelnit$MethodAndArgs
Caller.run(Runtimelnit.java:614)
at
com.android.internal.os.Zygotelnit.main(Zygotelnit.jav
a:1004)
Caused by: android.os.TransactionTooLargeException:
############################################

This happens only when the message that we reply have sequence of messages .
Can you please help and solve this issue?

Many thanks,

### Expected behavior

Attach file or image and continue to reply to message

### Actual behavior

App crash, and edit message close.

### Logs

java.lang.RuntimeException:
android.os.TransactionTooLargeException: data
parcel size 1164924 bytes
Bundle stats:
android:viewHierarchyState [size=25820]
android:views [size=25716]
0x7f0902d8 [size=16960]
0x7f0903b1 [size=1780]
androidx.lifecycle.BundlableSavedStateRegistry.key
[size=1056]
state:htmlQuote [size=1134364]
PersistableBundle stats:
[null]
at
android.app.servertransaction.PendingTransactionAct
ions$Stoplnfo.run(PendingTransactionActions.java:14
6)
at
android.os.Handler.handleCallback(Handler.java:959)
at
android.os.Handler.dispatchMessage(Handler.java:10
at android.os.Looper.loopOnce(Looper.java:249)
at android.os.Looper.loop(Looper.java:337)
at
android.app.ActivityThread.main(ActivityThread.java:9
403)
at
com.android.internal.os.Runtimelnit$MethodAndArgs
Caller.run(Runtimelnit.java:614)
at
com.android.internal.os.Zygotelnit.main(Zygotelnit.jav
a:1004)
Caused by: android.os.TransactionTooLargeException:

mvanhorn:
## Contribution Summary

Linked Issue/Ticket: Fixes #9425

#### Description

In `QuotedMessagePresenter.onSaveInstanceState` the offending line is `outState.putSerializable(STATE_KEY_HTML_QUOTE, quotedHtmlContent)`. Guard this put by the character length of the quoted content (`quotedHtmlContent.getQuotedContent()`): only persist the HTML quote into the saved-state `Bundle` when it is below a safe threshold well under the Binder limit (e.g. a ~256 KB / ~262144-char cap defined as a constant), and skip the put when it is larger. This eliminates the oversized parcel and therefore the crash. To keep the size check unit-testable without standing up the full presenter (whose constructor pulls DI singletons and a `MessageCompose`/view), extract the decision into a small pure helper (e.g. a package-visible static `shouldPersistQuotedHtml(InsertableHtmlContent, int maxChars)` / size computation) that `onSaveInstanceState` calls. `onRestoreInstanceState` already null-checks `quotedHtmlContent`, so when the quote was skipped the composer degrades gracefully rather than crashing: the user's typed text (saved separately by the `EditText`) is preserved, and only the rarely-hit "process killed while the picker is open, then recreated" path loses the re-displayable quote - an acceptable trade versus a guaranteed crash on every attach. `MessageCompose.java` is the caller of `onSaveInstanceState` but requires no change, so it is intentionally excluded from the scope list.

#### Screen Shots

N/A - no UI change; the fix guards saved-state serialization of the quoted-HTML in `QuotedMessagePresenter`.

## AI Disclosure

## Contribution Checklist

- [x] I have read and affirm that my contribution adheres to Mozilla's Community Participation Guidelines
- [x] This contribution adheres to the existing codestyle (spotless)
- [x] This contribution does not break existing unit tests
- [x] This contribution includes tests for any new functionality, and maintains tests for any updated functionality
- [ ] This contribution adheres to our Engineering process (RFC/Technical Design/ADR)
- [x] This PR has a descriptive title and body that accurately outlines all changes made, and references the issue it fixes (Fixes #9425).

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
@@ -55,6 +55,7 @@ import androidx.core.os.BundleCompat;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentManager;
 import app.k9mail.core.ui.legacy.designsystem.atom.icon.Icons;
+import app.k9mail.legacy.mailstore.MessageStoreManager;
 import com.bumptech.glide.Glide;
 import com.bumptech.glide.load.engine.DiskCacheStrategy;
 import com.fsck.k9.activity.compose.MessageComposeInAppNotificationFragment;
@@ -186,6 +187,8 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
     private static final String STATE_KEY_SOURCE_MESSAGE_PROCED =
             "com.fsck.k9.activity.MessageCompose.stateKeySourceMessageProced";
     private static final String STATE_KEY_DRAFT_ID = "com.fsck.k9.activity.MessageCompose.draftId";
+    private static final String STATE_KEY_RESTORE_FROM_DRAFT =
+            "com.fsck.k9.activity.MessageCompose.restoreFromDraft";
     private static final String STATE_IDENTITY_CHANGED =
             "com.fsck.k9.activity.MessageCompose.identityChanged";
     private static final String STATE_IDENTITY =
@@ -224,6 +227,7 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
     private final MessageLoaderHelperFactory messageLoaderHelperFactory = DI.get(MessageLoaderHelperFactory.class);
     private final DefaultFolderProvider defaultFolderProvider = DI.get(DefaultFolderProvider.class);
     private final MessagingController messagingController = DI.get(MessagingController.class);
+    private final MessageStoreManager messageStoreManager = DI.get(MessageStoreManager.class);
     private final Preferences preferences = DI.get(Preferences.class);
     private final GeneralSettingsManager generalSettingsManager = DI.get(GeneralSettingsManager.class);
 
@@ -286,6 +290,7 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
     private EditText signatureView;
     private EditText messageContentView;
     private LinearLayout attachmentsView;
+    private View composerContent;
 
     private String referencedMessageIds;
     private String repliedToMessageId;
@@ -299,6 +304,8 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
 
     private boolean sendMessageHasBeenTriggered = false;
     private boolean ignoreSentFolderNotAssigned = false;
+    private Integer pendingAttachmentPickerRequestCode;
+    private boolean restoringFromDraft;
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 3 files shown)


killpt:
Hi @mvanhorn,

Any news when this issue will be solved?

Thanks,

wmontwe:
@mvanhorn Thanks for finding the source of the issue and proposing a quick fix. This prevents the crash, but does so by silently omitting oversized quoted HTML from the saved state. After activity recreation, this could alter the composed message or cause part of its state to be lost.
Large composer content shouldn’t be stored in savedInstanceState. The bundle should contain only small restoration metadata, such as a draft identifier. The composer’s working state should instead be persisted as a draft and restored from there.
I think this requires a broader rework of composer-state persistence rather than an arbitrary size cutoff.

mvanhorn:
You were right that the size cutoff was the wrong shape. Trading a crash for silent content loss is not a fix, and an arbitrary threshold would have been a permanent source of "my reply lost the quoted part" reports. Pushed 8ebaeff0 taking the approach you described instead.

- `savedInstanceState` now carries only a draft reference, `STATE_KEY_RESTORE_FROM_DRAFT`, rather than composer content.
- When the activity is recreated with that reference present, the composer restores its working state from the draft.
- In that path the composer view hierarchy has saving disabled via `setSaveFromParentEnabled(false)`, so the view layer cannot write the content back into the bundle behind our backs. That was the part I would have missed: removing our own explicit put is not enough while the `EditText`s are still saving themselves.
- The size-threshold tests are replaced by one asserting the saved state carries only quote restoration metadata.

On verification, so you know exactly what has and has not been checked: `:legacy:ui:legacy:compileDebugJavaWithJavac` builds and `QuotedMessagePresenterTest` passes locally. That is the limit of it. I have not exercised real activity recreation, low-memory process death, or the interaction with an existing draft on a device, and those are precisely where this could still be wrong. Given this sits on the draft-persistence path, I would rather you look at the approach before trusting it than have me claim more confidence than I have.

Two things I am unsure about and would value your view on: whether restoring from draft should also cover the case where no draft has been saved yet at recreation time, and whether disabling parent save on the whole composer container is too broad a hammer compared with targeting the specific views.

PR #11268, opened 2026-07-18
fix: keep attachment file extension when saving

### LINKED ISSUE #7968: Saving an attachment (appointment.ics ) adds an extra file extension
### Checklist

### App version

6.901

F-Droid

### Android version

Android 13; Custom Rom: /e/ OS Version 2.1

### Device model

Fairphone 3

### Steps to reproduce

1. Receive a mail with an attached calendar file (MIME: text/calendar, name for example: appointment.ics)
2. Save the attachment to the filesystem. The filename in the save dialog is correctly shown as appointment.ics

### Expected behavior

Search for the file in the file browser and find, that it has been saved as apointment.ics

### Actual behavior

Search for the file in the file browser and find, that it has been saved as apointment.ics.txt

### Logs

--------- beginning of main
06-21 20:45:24.112  9043  9067 I MessagingController: Running command 'clearNotifications', seq = 841 (foreground priority)
06-21 20:45:24.114  9043  9067 V NotificationController: Removing some notifications for account da64a12f-9663-49f2-9097-d79d92f07dd9
06-21 20:45:24.115  9043  9067 V NotificationController: Removing some notifications for account 582425d0-6f4b-4e9a-ac4b-2a5b2ac58bc5
06-21 20:45:24.115  9043  9067 I MessagingController:  Command 'clearNotifications' completed
06-21 20:45:24.127  9043 10368 E OpenGLRenderer: Unable to match the desired swap behavior.
06-21 20:45:28.981  9043  9043 D MessageViewFragment: MessageViewFragment displaying message MessageReference(accountUuid=da64a12f-9663-49f2-9097-d79d92f07dd9, folderId=2, uid=7027)
06-21 20:45:28.987  9043  9043 D MessageLoaderHelper: Creating new local message loader
06-21 20:45:29.014  9043  9043 D MessageLoaderHelper: Creating new decode message loader
06-21 20:45:29.015  9043  8367 V NotificationController: Removing notification for message MessageReference(accountUuid=da64a12f-9663-49f2-9097-d79d92f07dd9, folderId=2, uid=7027)
06-21 20:45:29.187  9043  9043 D Utility : External images found
06-21 20:45:38.135  9043 10368 E OpenGLRenderer: Unable to match the desired swap behavior.
k9-log.txt (END)

mvanhorn:
## Contribution Summary

Linked Issue/Ticket: Closes #7968
RFC / Technical Design (if applicable): N/A (small bug fix)

#### Description

When saving a mail attachment to the filesystem, an extra file extension was appended to the file on disk: `appointment.ics` was written as `appointment.ics.txt` (and a corroborating report showed `something.pkpass` written with a spurious trailing extension). The save dialog showed the correct name, but the saved file gained an extra extension.

**Root cause**

`MessageViewFragment.onSaveAttachment` launches `Intent.ACTION_CREATE_DOCUMENT` (via `CreateDocumentResultContract`) passing the attachment's *declared* MIME type, which comes from the mail `Content-Type` header (e.g. `text/plain`), together with the display name (`appointment.ics`). Android's Storage Access Framework (`FileSystemProvider.splitFileName`) checks whether the display name's extension round-trips to the supplied MIME type. Because `.ics` does not map back to `text/plain`, the provider appends the MIME type's preferred extension (`text/plain` -> `.txt`), producing `appointment.ics.txt`.

**Fix**

Pass a MIME type that is consistent with the display name's extension, so the SAF provider leaves the filename unchanged. This mirrors the pattern the VIEW flow already uses (`ViewIntentFinder.getBestViewIntent` derives `MimeTypeUtil.getMimeTypeByExtension(displayName)`).

- Added a small, unit-testable helper `MimeTypeUtil.getMimeTypeForFilename(displayName, declaredMimeType)` that returns the extension-derived MIME type when the display name has an extension, and otherwise falls back to the declared MIME type (or `application/octet-stream` when that is `null`).
- `onSaveAttachment` now builds the `CreateDocumentResultContract.Input` with `mimeType = MimeTypeUtil.getMimeTypeForFilename(attachment.displayName, attachment.mimeType)`.

Because the derived MIME type round-trips with the extension Android computes, the provider no longer appends an extra extension. The actual byte copy in `AttachmentController.writeAttachment` is unaffected, and the unrelated EML-export path (which passes its own explicit `message/rfc822`) is untouched.

Notes on edge cases (all covered by the new unit test):

- Known extension (`appointment.ics`, `boardingpass.pkpass`): resolves to the extension-consistent type (`text/calendar`, `application/vnd.apple.pkpass`) so the filename is preserved.
- No extension (`noname`): falls back to the declared type, and to `application/octet-stream` when the declared type is `null` — the provider may then add a suitable extension, which is the desired behaviour.
- Unknown extension (`mystery.unknownext`): resolves to `application/octet-stream`. This is deliberate: SAF treats `octet-stream` as matching any unrecognised extension, so it keeps the filename as-is rather than appending a spurious extension. (Returning the declared type here would reintroduce the double-extension bug.)

This also removes a latent `requireNotNull(attachment.mimeType)` that could throw for an attachment with a null declared MIME type; the new helper handles that case gracefully.

#### Screen Shots

No UI changes (the fix affects the filename passed to the create-document intent, not any visible screen).

## AI Disclosure

## Contribution Checklist

- [x] This contribution is in Kotlin where possible (the caller change and the new test are Kotlin; the helper is added to the existing `MimeTypeUtil` Java utility class alongside the closely related `getMimeTypeByExtension` it builds on, to keep the change minimal and consistent)
- [x] This contribution adheres to the existing codestyle (`./gradlew :legacy:core:spotlessCheck :legacy:ui:legacy:spotlessCheck` passes; `./gradlew :legacy:core:detekt :legacy:ui:legacy:detekt` passes)
- [x] This contribution does not break existing unit tests (`./gradlew :legacy:core:testDebugUnitTest :legacy:ui:legacy:testDebugUnitTest` passes: 751 and 245 tests respectively, 0 failures)
- [x] This contribution includes tests for any new functionality, and maintains tests for any updated functionality (new `MimeTypeUtilTest` with 6 cases covering known/unknown/missing extensions)
- [x] This contribution adheres to our [Engineering process](https://github.com/thunderbird/thunderbird-android/tree/main/docs/engineering) (RFC/Technical Design/ADR) — N/A for a self-contained bug fix
- [x] This PR has a descriptive title and body that accurately outlines all changes made, and contains a reference to any issues that it fixes (Closes #7968)

### Testing

Run locally with JDK 21 and Android SDK 36:

- `./gradlew :legacy:core:testDebugUnitTest --tests "com.fsck.k9.helper.MimeTypeUtilTest"` — 6 passed, 0 failed
- `./gradlew :legacy:core:testDebugUnitTest :legacy:ui:legacy:testDebugUnitTest` — 996 passed (6 skipped, pre-existing), 0 failed
- `./gradlew :legacy:core:spotlessCheck :legacy:ui:legacy:spotlessCheck` — passed
- `./gradlew :legacy:core:detekt :legacy:ui:legacy:detekt` — passed

`connectedAndroidTest` (instrumented) was not run; there is no device/emulator in this environment. The SAF filename behaviour itself is provided by the Android platform and is exercised via the new unit test on the MIME-resolution helper that drives it.

AI was used for assistance.

### DIFF
--- a/legacy/core/src/main/java/com/fsck/k9/helper/MimeTypeUtil.java
+++ b/legacy/core/src/main/java/com/fsck/k9/helper/MimeTypeUtil.java
@@ -898,6 +898,26 @@ public class MimeTypeUtil {
         return DEFAULT_ATTACHMENT_MIME_TYPE;
     }
 
+    /**
+     * Returns a MIME type that is consistent with the extension of {@code displayName}.
+     *
+     * <p>This is used when creating a document via the Storage Access Framework. If the declared MIME type does not
+     * round-trip to the display name's extension, the provider appends the declared type's preferred extension (e.g.
+     * saving {@code appointment.ics} declared as {@code text/plain} would produce {@code appointment.ics.txt}). By
+     * passing a MIME type derived from the extension, the provider keeps the filename unchanged.
+     *
+     * <p>When the display name has no extension, there is nothing to preserve, so the declared MIME type (or
+     * {@link #DEFAULT_ATTACHMENT_MIME_TYPE} when it is {@code null}) is returned and the provider may add a suitable
+     * extension.
+     */
+    public static String getMimeTypeForFilename(String displayName, String declaredMimeType) {
+        if (displayName != null && displayName.lastIndexOf('.') != -1) {
+            return getMimeTypeByExtension(displayName);
+        }
+
+        return declaredMimeType != null ? declaredMimeType : DEFAULT_ATTACHMENT_MIME_TYPE;
+    }
+
     public static String getExtensionByMimeType(@NotNull String mimeType) {
         String lowerCaseMimeType = mimeType.toLowerCase(Locale.US);
         for (String[] contentTypeMapEntry : MIME_TYPE_BY_EXTENSION_MAP) {
--- a/legacy/core/src/main/java/com/fsck/k9/helper/MimeTypeUtil.java
+++ b/legacy/core/src/main/java/com/fsck/k9/helper/MimeTypeUtil.java
@@ -4,6 +4,7 @@ package com.fsck.k9.helper;
 import java.util.Locale;
 
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 
 public class MimeTypeUtil {
@@ -910,6 +911,7 @@ public class MimeTypeUtil {
      * {@link #DEFAULT_ATTACHMENT_MIME_TYPE} when it is {@code null}) is returned and the provider may add a suitable
      * extension.
      */
+    @Nullable
     public static String getMimeTypeForFilename(String displayName, String declaredMimeType) {
         if (displayName != null && displayName.lastIndexOf('.') != -1) {
             return getMimeTypeByExtension(displayName);
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessageViewFragment.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessageViewFragment.kt
@@ -48,6 +48,7 @@ import com.fsck.k9.fragment.ConfirmationDialogFragment
 import com.fsck.k9.fragment.ConfirmationDialogFragment.ConfirmationDialogFragmentListener
 import com.fsck.k9.helper.HttpsUnsubscribeUri
 import com.fsck.k9.helper.MailtoUnsubscribeUri
+import com.fsck.k9.helper.MimeTypeUtil
 import com.fsck.k9.helper.UnsubscribeUri
 import com.fsck.k9.mail.Part
 import com.fsck.k9.mailstore.AttachmentViewInfo
@@ -1197,9 +1198,7 @@ class MessageViewFragment :
             createDocumentLauncher.launch(
                 input = CreateDocumentResultContract.Input(
                     title = attachment.displayName ?: getString(MessageReaderR.string.unnamed_attachment_title),
-                    mimeType = requireNotNull(attachment.mimeType) {
-                        "Invalid attachment type. The mimeType is null. Attachment = $attachment"
-                    },
+                    mimeType = MimeTypeUtil.getMimeTypeForFilename(attachment.displayName, attachment.mimeType),
                 ),
             )
         } catch (_: ActivityNotFoundException) {
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessageViewFragment.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessageViewFragment.kt
@@ -51,7 +51,6 @@ import com.fsck.k9.fragment.ConfirmationDialogFragment
 import com.fsck.k9.fragment.ConfirmationDialogFragment.ConfirmationDialogFragmentListener
 import com.fsck.k9.helper.HttpsUnsubscribeUri
 import com.fsck.k9.helper.MailtoUnsubscribeUri
-import com.fsck.k9.helper.MimeTypeUtil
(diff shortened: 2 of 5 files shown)


legacy/core/src/main/java/com/fsck/k9/helper/MimeTypeUtil.java
```diff
@@ -898,6 +898,26 @@ public static String getMimeTypeByExtension(String filename) {
...
+     * <p>This is used when creating a document via the Storage Access Framework. If the declared MIME type does not
+     * round-trip to the display name's extension, the provider appends the declared type's preferred extension (e.g.
+     * saving {@code appointment.ics} declared as {@code text/plain} would produce {@code appointment.ics.txt}). By
+     * passing a MIME type derived from the extension, the provider keeps the filename unchanged.
+     *
+     * <p>When the display name has no extension, there is nothing to preserve, so the declared MIME type (or
+     * {@link #DEFAULT_ATTACHMENT_MIME_TYPE} when it is {@code null}) is returned and the provider may add a suitable
+     * extension.
+     */
+    public static String getMimeTypeForFilename(String displayName, String declaredMimeType) {
```
rafaeltonholo, line 915: If the return can be nullable, the method on java must always be annotated with `@Nullable`.

```suggestion
    @Nullable
    public static String getMimeTypeForFilename(String displayName, String declaredMimeType) {
```

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messageview/MessageViewFragment.kt
```diff
@@ -1197,9 +1198,7 @@ class MessageViewFragment :
             createDocumentLauncher.launch(
                 input = CreateDocumentResultContract.Input(
                     title = attachment.displayName ?: getString(MessageReaderR.string.unnamed_attachment_title),
-                    mimeType = requireNotNull(attachment.mimeType) {
-                        "Invalid attachment type. The mimeType is null. Attachment = $attachment"
-                    },
+                    mimeType = MimeTypeUtil.getMimeTypeForFilename(attachment.displayName, attachment.mimeType),
```
rafaeltonholo: There are two issues with this code:

1. We should not transform data in a view. The view should only consume the data provided to it and perform validations, in case of a need to display something to the user. This should be moved to where the `attachment` is created.
2. This code can potentially cause a NullPointerException without an explanation why. Since the mimetype can return null at this point. The best approach here would be to wrap it using a `requireNotNull`, so we can add a proper messaging that would help us to understand what actually caused the NullPointerException.

rafaeltonholo (changes requested):
Thanks for adding this contribution; however, there are a few points we need to address before being able to merge it.

mvanhorn:
Updated in 678ada5.

The extension is now resolved from the attachment's MIME type when the supplied name has no suffix, in `MimeTypeUtil`, and applied where the display name is built in `AttachmentInfoExtractor`, so the message view and the save path agree rather than diverging at the point of saving.

That second file is beyond what I first touched. Keeping the change inside `MimeTypeUtil` alone left the name correct in one place and stale in the other, which is what made the earlier version incomplete.

`MimeTypeUtilTest` and `AttachmentInfoExtractorTest` both pass locally.

rafaeltonholo:
@mvanhorn, your branch has conflicts. Could you please rebase and fix them?

mvanhorn:
Rebased onto current main in a8c8e93, @wmontwe.

The only conflict was an import: main moved `Log` from `net.thunderbird.core.logging.legacy` to `net.thunderbird.legacy.logging`, so I took main's package. No behavior change — the diff against main is the same 5 files and +95/-7 it was before the rebase.

`:legacy:core:compileDebugJavaWithJavac` is green locally.

PR #11309, opened 2026-07-26
Fix crash when camera result arrives after process death (#11296)

### LINKED ISSUE #11296: [Crash] UninitializedPropertyAccessException in CameraCaptureHandler when attaching photo via camera
### Summary

Thunderbird crashes when adding an attachment to a message using the camera. The crash occurs when `MessageCompose` receives the successful camera activity result after the application process has been recreated.

The issue is reproducible on the `main` branch and was found during the target SDK 36 migration.

### Environment

* **Product:** `lineage_gta4xl`
* **Model:** `SM_P615`
* **Device:** `gta4xl`
* **Android version:** Android 16
* **Application ID:** `net.thunderbird.android.debug`
* **WebView version:** `145.0.7632.109`

### Steps to reproduce

1. Open the message composer or edit an existing draft.
2. Select the option to add an attachment.
3. Choose the camera as the attachment source.
4. Capture and accept an image.
5. Return to Thunderbird.

### Actual behaviour

The application crashes while resuming `MessageCompose` and processing the camera result:

```text
java.lang.RuntimeException: Unable to resume activity
    {net.thunderbird.android.debug/com.fsck.k9.activity.MessageCompose}

Caused by: java.lang.RuntimeException:
    Failure delivering result ResultInfo{who=null, request=6, result=-1, data=null}

Caused by: kotlin.UninitializedPropertyAccessException:
    Image Uri not initialized
    at app.k9mail.core.android.common.camera.CameraCaptureHandler.getCapturedImageUri(
        CameraCaptureHandler.kt:30
    )
    at com.fsck.k9.activity.MessageCompose.onActivityResult(
        MessageCompose.java:984
    )
```

The log shows that the original application process is terminated before the camera result is handled. Android then starts a new process, restores `MessageCompose` with the `EDIT_DRAFT` intent, and delivers a successful activity result with `data=null`. The recreated `CameraCaptureHandler` no longer has its captured-image URI initialized, so `getCapturedImageUri()` throws `UninitializedPropertyAccessException`. 

### Expected behaviour

The captured image should be attached to the message after returning from the camera, including when the activity or application process has been recreated. A missing or unrestored camera output URI should be handled without crashing.

### Suspected cause

The camera output URI is held only in the in-memory state of `CameraCaptureHandler` and is not restored after process recreation. `MessageCompose.onActivityResult()` assumes that the URI is always initialized when it receives a successful camera result.

We need to make `CameraCaptureHandler` survive process death.

ibrahim-iqbal:
## Contribution Summary

Linked Issue/Ticket: Fixes #11296

#### Description

Thunderbird crashes when `MessageCompose` receives a successful camera activity result after the app process has been recreated. The captured-image URI lived only in the in-memory state of the `CameraCaptureHandler` Koin singleton, so on process death the fresh singleton had no URI and `getCapturedImageUri()` threw:

```
kotlin.UninitializedPropertyAccessException: Image Uri not initialized
  at app.k9mail.core.android.common.camera.CameraCaptureHandler.getCapturedImageUri(CameraCaptureHandler.kt:30)
  at com.fsck.k9.activity.MessageCompose.onActivityResult(MessageCompose.java:984)
```

This PR makes the handler recoverable across process death.

**`CameraCaptureHandler`:**
- `capturedImageUri` is now `Uri?` instead of `lateinit var`.
- `getCapturedImageUri()` returns `Uri?` instead of throwing.
- New `saveInstanceState(Bundle)` / `restoreInstanceState(Bundle?)` helpers round-trip the URI through a namespaced key (`STATE_KEY_CAPTURED_IMAGE_URI`).

**`MessageCompose`:**
- `onCreate(savedInstanceState)` calls `cameraCaptureHandler.restoreInstanceState(savedInstanceState)`.
- `onSaveInstanceState(outState)` calls `cameraCaptureHandler.saveInstanceState(outState)`.
- `onActivityResult` null-checks the URI. If it can't be recovered, the user sees a Toast (`camera_capture_lost`: "Captured image was lost because the app was closed. Please try again.") instead of a crash. This matches the issue's expected behaviour ("A missing or unrestored camera output URI should be handled without crashing").

**Tests:** `CameraCaptureHandlerTest` (Robolectric) covers four regressions:
- `getCapturedImageUri` returns null before `openCamera` has been called
- `getCapturedImageUri` returns null after `restoreInstanceState(null)`
- `restoreInstanceState` rehydrates the URI a previous handler wrote out
- `saveInstanceState` omits the URI when nothing has been captured

Locally verified:
- `./gradlew :core:android:common:testDebugUnitTest --tests "...CameraCaptureHandlerTest"` — passes
- `./gradlew :legacy:ui:legacy:compileDebugJavaWithJavac` — clean

**Notes for reviewers:**
- `getCapturedImageUri()` used to throw; call sites should now null-check. `MessageCompose` is the only in-tree caller and is updated in this PR.
- The Toast string is English-only in `legacy/ui/legacy/src/main/res/values/strings.xml` — happy to add translations if that's the convention.
- The Robolectric test uses `Bundle.getParcelable(String)` which is deprecated on API 33+. Kept for minimal-diff; can switch to `BundleCompat.getParcelable(bundle, key, Uri.class)` in a follow-up.

#### Screen Shots

_No UI changes — this is a crash fix. The only user-visible change on the recovery path is a Toast message on an edge case that previously crashed the app._

## AI Disclosure

## Contribution Checklist

- [x] This contribution is in Kotlin where possible (handler + tests are Kotlin; `MessageCompose` is pre-existing Java)
- [x] This contribution adheres to the existing codestyle (verified locally)
- [x] This contribution does not break existing unit tests (`testDebugUnitTest` on the affected module passes)
- [x] This contribution includes tests for any new functionality, and maintains tests for any updated functionality
- [x] This contribution adheres to your Engineering process — this is a bugfix scoped to the referenced issue, no new architecture

---

_Note: `require-report-label` CI check needs a maintainer to apply one of `report: include` / `report: highlight` / `report: exclude`. This is a user-visible bug fix (crash on returning from the camera after process death), so `report: include` looks right — happy to adjust if that doesn't match your release-notes conventions._

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
@@ -305,11 +305,6 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-
-        // Rehydrate the camera-capture handler's URI so the pending
-        // camera result can still be attached after process death. See
-        // issue #11296: without this the handler is a fresh singleton and
-        // getCapturedImageUri() used to throw.
         cameraCaptureHandler.restoreInstanceState(savedInstanceState);
 
         if (databaseUpgradeInterceptor.checkAndHandleUpgrade(this, getIntent())) {
@@ -987,12 +982,8 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
             }
 
             if (requestCode == REQUEST_IMAGE_CAPTURE && resultCode == Activity.RESULT_OK) {
-                android.net.Uri capturedImageUri = cameraCaptureHandler.getCapturedImageUri();
+                final Uri capturedImageUri = cameraCaptureHandler.getCapturedImageUri();
                 if (capturedImageUri == null) {
-                    // The URI wasn't restored after process death (issue
-                    // #11296). We can't attach an image we no longer know
-                    // the location of; surface a message instead of
-                    // crashing.
                     Toast.makeText(this, R.string.camera_capture_lost, Toast.LENGTH_LONG).show();
                     return;
                 }
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
@@ -306,6 +306,12 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+        // Rehydrate the camera-capture handler's URI so the pending
+        // camera result can still be attached after process death. See
+        // issue #11296: without this the handler is a fresh singleton and
+        // getCapturedImageUri() used to throw.
+        cameraCaptureHandler.restoreInstanceState(savedInstanceState);
+
         if (databaseUpgradeInterceptor.checkAndHandleUpgrade(this, getIntent())) {
             finish();
             return;
@@ -723,6 +729,7 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
         recipientPresenter.onSaveInstanceState(outState);
         quotedMessagePresenter.onSaveInstanceState(outState);
         attachmentPresenter.onSaveInstanceState(outState);
+        cameraCaptureHandler.saveInstanceState(outState);
     }
 
     @Override
@@ -980,8 +987,17 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
             }
 
             if (requestCode == REQUEST_IMAGE_CAPTURE && resultCode == Activity.RESULT_OK) {
+                android.net.Uri capturedImageUri = cameraCaptureHandler.getCapturedImageUri();
+                if (capturedImageUri == null) {
+                    // The URI wasn't restored after process death (issue
+                    // #11296). We can't attach an image we no longer know
+                    // the location of; surface a message instead of
+                    // crashing.
+                    Toast.makeText(this, R.string.camera_capture_lost, Toast.LENGTH_LONG).show();
+                    return;
+                }
                 Intent intent = new Intent();
-                intent.setData(cameraCaptureHandler.getCapturedImageUri());
+                intent.setData(capturedImageUri);
                 attachmentPresenter.onActivityResult(resultCode, REQUEST_CODE_ATTACHMENT_URI, intent);
                 return;
             }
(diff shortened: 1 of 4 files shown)


legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
```diff
@@ -980,8 +987,17 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
             }
 
             if (requestCode == REQUEST_IMAGE_CAPTURE && resultCode == Activity.RESULT_OK) {
+                android.net.Uri capturedImageUri = cameraCaptureHandler.getCapturedImageUri();
```
rafaeltonholo: Please, avoid full qualifiers when they are not needed. Additionally, the `capturedImageUri` is never changed after it's assignment, thus it should be final.

```suggestion
final Uri capturedImageUri = cameraCaptureHandler.getCapturedImageUri();
```

legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
```diff
@@ -980,8 +987,17 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+                if (capturedImageUri == null) {
+                    // The URI wasn't restored after process death (issue
+                    // #11296). We can't attach an image we no longer know
+                    // the location of; surface a message instead of
+                    // crashing.
+                    Toast.makeText(this, R.string.camera_capture_lost, Toast.LENGTH_LONG).show();
```
rafaeltonholo: Please, remove the unnecessary comments. This just bloat the code.
```suggestion
                    Toast.makeText(this, R.string.camera_capture_lost, Toast.LENGTH_LONG).show();
```

legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageCompose.java
```diff
@@ -306,6 +306,12 @@ public class MessageCompose extends BaseActivity implements OnClickListener,
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+        // Rehydrate the camera-capture handler's URI so the pending
+        // camera result can still be attached after process death. See
+        // issue #11296: without this the handler is a fresh singleton and
+        // getCapturedImageUri() used to throw.
```
rafaeltonholo: Unnecessary comment block. Please remove.
```suggestion
```

rafaeltonholo (commented):
Thanks for this contribution. LGTM, however, I would like to not bloat the code with unnecessary comments.

rafaeltonholo:
@ibrahim-iqbal, you also need to address the quality issues. Please run `./gradlew spotlessApply detekt --rerun-tasks`, address the issues (in case they were not auto-fixed), and commit the fixes.

rafaeltonholo (changes requested):

rafaeltonholo (approved):
LGTM, thanks for addressing the comments

PR #11338, opened 2026-07-29
fix: use filtered counts for unified account

### LINKED ISSUE #9773: Unread message count wrong for filtered Unified Inbox
### Checklist

### App

Thunderbird for Android

### App version

12.1

F-Droid

### Android version

16

### Device model

Google Pixel 6 Pro

### Steps to reproduce

1. Click the hamburger menu to see the account list
2. Check the unread email count for each account vs the Unified Inbox

### Expected behavior

UN icon for Unified Inbox count reports unread count for only “opted-in” accounts.

### Actual behavior

UN icon shows unread messages for all accounts. 

I have 4 accounts set up, and have removed two from the unified inbox. However, the new Unified Inbox pseudo-account icon “UN” seems to show the sum of all account unread inbox emails, not just those attached to the unified inbox.

I have all emails read apart from several in one of the accounts that is _not_ attached to the unified inbox. The unread count on the UN icon shows the same number that is also on the icon for said unread account. 

[Unrelated to this issue, largely, is that I'm not quite sure what the use case is for the new UN icon, it seems to just make it necessary to have one more click to access non-Inbox folders in my “main” account, which previously sat “under” the unified inbox. It's not needed to return to the unified inbox as _back_ already (and still) does that.)l]

### Logs

_No response_

arpitagarwal1301:
Fixes #9773

Use the existing filtered unified-inbox counts for the Unified Account badge instead of summing all account totals. This prevents folders excluded from the unified inbox from inflating the unread and starred counts.

## Verification

- `:feature:navigation:drawer:dropdown:testDebugUnitTest --tests 'net.thunderbird.feature.navigation.drawer.dropdown.domain.usecase.GetDisplayAccountsTest'`
- `spotlessCheck`
- `git diff --check`

### DIFF
--- a/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/NavigationDrawerModule.kt
+++ b/feature/navigation/drawer/dropdown/src/main/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/NavigationDrawerModule.kt
@@ -42,6 +42,7 @@ val navigationDropDownDrawerModule: Module = module {
             notificationStream = get(),
             featureFlagProvider = get(),
             avatarMapper = get(),
+            unifiedFolderRepository = get(),
         )
     }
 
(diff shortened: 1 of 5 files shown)


rafaeltonholo (changes requested):
Hi @arpitagarwal1301, thanks for this contribution!

The changes work partially. If we change the folder `Unify` option, the changes only get applied when reopening the app, when they should be applied when getting back to the Message List/Drawer screen.

See the screen recording for reference:

https://github.com/user-attachments/assets/8ace060a-764d-4ec5-923d-0d486816dbfe

rafaeltonholo:
Hi @arpitagarwal1301, are you able to address my last comment?

PR #11350, opened 2026-08-02
fix: enable multi-line input for Composition Defaults signature field

### LINKED ISSUE #11343: Composition Defaults: Signature Field Does Not Accept Multi-Line Input
## Summary 
After the modernization of the Composition defaults screen (issue #10724 / PR #11012), the signature text field appears to accept only a single line of text. Multi-line signatures (with line breaks) are not possible, even though they work in the Manage Identities menu.This may be a regression introduced during the XML → Jetpack Compose migration, or it may have been a pre-existing limitation of the old screen that was never noticed.

## Steps to reproduce
- Open Thunderbird for Android (version 21.0 or later).
- Go to Settings -> Account -> Sending mail -> Composition defaults.
- Enable “Use signature” (or equivalent).
- Try to enter a multi-line signature (e.g. press Enter / Return to create a new line, or paste text containing newlines).

## Expected behavior
The signature field should accept and preserve multi-line text, consistent with the behavior in Manage Identities.

## Actual behavior
Only a single line of text can be entered. Line breaks are either ignored or the field behaves as a single-line TextField.

shamim-emon:
- Fixes #11343


rafaeltonholo (approved):
LGTM

PR #11369, opened 2026-08-05
fix: show feedback when search results are empty or remote search ends

JustNotesa:
## Description

A manual search whose local search finds nothing currently shows a completely blank screen: no hint that nothing was found, no hint that only locally synced messages were searched, and no visible way to discover the remote search action (small toolbar icon / pull-to-refresh). A failed remote search only flashes a short toast; a remote search with zero hits ends on the same blank screen.

This PR adds lasting feedback in the (previously always empty) search footer:

- No local results + remote search available: **"No messages found on this device. Tap to search on the server."** - tapping it starts the remote search directly
- No results otherwise: **"No messages found"**
- After a remote search with no hits: **"No messages found on the server"**
- After a failed remote search: **"Remote search failed"** stays visible (the existing toast is kept)

The search footer is now computed in one place (`getSearchResultsFooterText()`) so these states stay consistent, and the footer layout can wrap to two lines.

Related to #4513.

## Reason for the change

Users interpret the blank search screen as "search is broken" because nothing communicates that only local messages were searched or that a server search exists. See reproduction below.

## Reproduction

1. IMAP account, open a folder, tap the search icon
2. Search for a term that only matches messages that are on the server but not synced locally
3. Before: completely blank result screen. After: prompt shown; tapping it runs the server search

## Testing

```
./gradlew :legacy:ui:legacy:testDebugUnitTest :legacy:ui:legacy:detekt :legacy:ui:legacy:spotlessCheck
./gradlew :app-thunderbird:assembleFossDebug
```

All pass. Manually verified in the emulator against a local GreenMail IMAP server (empty result prompt, tap-to-search, zero-hit result line, failure line with server stopped). `connectedAndroidTest` was not run (no CI device available locally).

## Risks / trade-offs

- UI-only change in the legacy message list fragment; no controller/protocol changes
- The footer strings are new; translations will follow via Weblate

## Disclosure

This change was developed with AI assistance (Claude). It was reviewed, built and manually tested by a human before submission.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11370, opened 2026-08-05
fix: keep remote search progress messages visible in the list footer

JustNotesa:
## Description

While a remote search is running, the `MessagingListener` callbacks show progress in the list footer ("Sending query to server", "Fetching N results"). As soon as downloaded results start arriving, every resulting message list update erases those messages again in two ways:

1. `setMessageList()` rebuilds `adapter.viewItems` without the footer item, so the footer vanishes until the next progress callback happens to fire
2. `updateFooterText()` (called from `setMessageList()`) recomputes the footer from folder state and sets it to `null` for manual searches, overwriting the progress message

In practice the progress text flickers or is not visible at all during a remote search, which makes the search look stuck or broken.

This PR keeps the existing footer item when the list is rebuilt and skips the footer recomputation while a remote search is in progress (`remoteSearchFuture != null`). When the search finishes, the footer is recomputed as before.

Related to #4513.

## Reproduction

1. IMAP account, search in a folder for a term with several server-only matches, trigger "Search messages on server"
2. Before: the progress footer disappears/flickers as soon as the first results arrive. After: it stays visible until the search completes

(Easiest to observe against a throttled/slow IMAP server.)

## Testing

```
./gradlew :legacy:ui:legacy:testDebugUnitTest :legacy:ui:legacy:detekt :legacy:ui:legacy:spotlessCheck
./gradlew :app-thunderbird:assembleFossDebug
```

All pass. Manually verified in the emulator against a local GreenMail IMAP server behind a delay proxy: progress messages now stay visible while results stream in. `connectedAndroidTest` was not run (no CI device available locally).

## Risks / trade-offs

- 7-line UI-only change in the legacy message list fragment
- Independent of, but complementary to, #11369 (empty-result feedback); trivial to rebase whichever lands second

## Disclosure

This change was developed with AI assistance (Claude). It was reviewed, built and manually tested by a human before submission.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/LegacyMessageListFragment.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagelist/LegacyMessageListFragment.kt
@@ -1391,6 +1391,10 @@ class LegacyMessageListFragment :
     }
 
     private fun updateFooterText() {
+        // While a remote search is running, the MessagingListener callbacks supply the footer text
+        // (progress messages). Don't overwrite them here.
+        if (remoteSearchFuture != null) return
+
         val currentFolder = this.currentFolder
         val account = this.account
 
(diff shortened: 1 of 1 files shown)


jbott-tbird:
This doesn't follow our guidelines or template.  The comments don't make clear what is being done or why.    Please refer to our community guidlines on contributing [here](https://github.com/thunderbird/thunderbird-android/blob/main/docs/contributing/contribution-workflow.md#discussing-your-plan)

JustNotesa:
Thanks for the pointer, and apologies for not following the workflow here as well. The underlying bug is now filed as #11378 with a proposed approach; I'll wait for triage/assignment there before opening a new PR using the template.

Issue #11320, opened 2026-07-28
Notification Settings — Create RFC For Default-On IMAP Idle (Push)


dani-zilla:
Because [IMAP IDLE](https://datatracker.ietf.org/doc/html/rfc2177) could potentially result in more battery usage, track what it would be on a device or two and report on what the change entails in a small RFC document before we go ahead with [this change](https://github.com/thunderbird/thunderbird-android/issues/11299). Include a rough description of how IMAP IDLE is used currently (which we label as "Push" in most locations in the app) as well.

wmontwe:
I think the RFC currently documents the existing implementation more than it proposes a change. I suggest splitting it into two documents: one describing the current push behavior and another RFC focused specifically on the proposed changes.

See also: https://github.com/thunderbird/thunderbird-android/pull/11403#issuecomment-5395896706

PR #11403, opened 2026-08-12
docs(RFC): Add RFC for Defaulting IMAP IDLE (Push) On for New Accounts

dani-zilla:
## Contribution Summary

Linked Issue/Ticket: [#11320](https://github.com/thunderbird/thunderbird-android/issues/11320)

#### Description

Creates the RFC describing the general idea of how we can implement push syncing for new users' inbox folders. 

## AI Disclosure

## Contribution Checklist

- [X] This contribution includes tests for any new functionality, and maintains tests for any updated functionality.

### DIFF
--- a/docs/engineering/rfcs/0005-imap-idle-push-on-new-account-inboxes.md
+++ b/docs/engineering/rfcs/0005-imap-idle-push-on-new-account-inboxes.md
@@ -1,37 +1,86 @@
-# RFC: Add "Push" Notifications to New Users' Inboxes With IMAP IDLE
+# RFC 0005: Enable IMAP IDLE ("Push") for the Inbox by Default on Newly Added IMAP Accounts
 
 - Issue: [#11320](https://github.com/thunderbird/thunderbird-android/issues/11320)
 - Status: **Proposed**
 
 ## Summary
 
-When a new user creates an account, if it is an IMAP account and IMAP IDLE is supported on the server, we will turn on "push" syncing for the inbox folder using IMAP IDLE. If they choose to enable notifications, they'll receive notifications when they receive new emails in as close to instant as we can get it without a true push service.
+When a user adds a new account, if it is an IMAP account and IMAP IDLE is supported on the server, we will turn on "
+push" syncing for the inbox folder using IMAP IDLE. If they choose to enable notifications, they'll receive
+notifications when they receive new emails in as close to instant as we can get it without a true push service.
 
 ## Motivation
 
-Users complain that their emails do not show up on time. When they learn they can activate IMAP IDLE, which we call "push" to the end user, they find that the setting is buried in an unexpected place. They may also misconfigure it, causing their email to stop working altogether. While IMAP IDLE works on a per-folder basis, the end user doesn't want to have to think about this level of detail. They don't want to have to figure out server capacities for open sockets and pick their most important folders. They just want their new mail when it comes in. Therefore, we'll turn on IMAP IDLE on their Inbox, what they'll see as "push," which will give them a notification or instant syncing of their inbox when changes appear in that folder on the server. Most users will never feel the need to change this setting, and will believe it's working as it was always intended.
+Users complain that their emails do not show up on time. When they learn they can activate IMAP IDLE, which we call "
+push" to the end user, they find that the setting is buried in an unexpected place. They may also misconfigure it,
+causing their email to stop working altogether. While IMAP IDLE works on a per-folder basis, the end user doesn't want
+to have to think about this level of detail. They don't want to have to figure out server capacities for open sockets
+and pick their most important folders. They just want their new mail when it comes in. Therefore, we'll turn on IMAP
+IDLE on their Inbox, what they'll see as "push," which will give them a notification or instant syncing of their inbox
+when changes appear in that folder on the server. Most users will never feel the need to change this setting, and will
+believe it's working as it was always intended.
 
 ## Proposal
 
-Whether or not a folder has push enabled is a boolean stored locally in `FolderDetails.isPushEnabled`, with those details saved to the device's storage in `FolderSettingsDataStore`. We'd update the folder's details to contain `isPushEnabled = true` when creating a folder details instance for the account's inbox folder.
+Whether or not a folder has push enabled is a boolean stored locally in `FolderDetails.isPushEnabled`, with those
+details saved to the device's storage in `FolderSettingsDataStore`. We'd update the folder's details to contain
+`isPushEnabled = true` when creating a folder details instance for the account's inbox folder.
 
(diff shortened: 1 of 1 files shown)


docs/engineering/rfcs/0005-imap-idle-push-on-new-account-inboxes.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Risks & Drawbacks
+
+IMAP IDLE keeps a socket open to receive information from the server when there are changes to the folders they're watching. This will be the Inbox folder. If a user receives many email messages throughout the day, it could potentially lead to more fetches, and more battery usage. In my personal testing, there seemed to be no difference in battery consumption between fetching all folders every 15 minutes and turning on IMAP IDLE for multiple accounts, and push on for multiple folders as well, more than our end users would get with the default settings proposed here. Users can turn the feature off if it becomes an issue, but IMAP IDLE was made specifically to reduce the need to sync with a server frequently and reduce both battery usage and network bandwidth. As a result, it may reduce battery consumption for our users. Those who see an increase would likely have an account where they receive many emails, but not all at once, such as automated emails every few minutes. Someone with such a setup could turn off the setting where it exists today. In a future project, we'll make these settings even easier to access.
```
rafaeltonholo: I would soften the following claim a bit:

> As a result, it may reduce battery consumption for our users.

Maybe:

```suggestion
IMAP IDLE keeps a socket open to receive information from the server when there are changes to the folders they're watching. This will be the Inbox folder. If a user receives many email messages throughout the day, it could potentially lead to more fetches, and more battery usage. In my personal testing, there seemed to be no difference in battery consumption between fetching all folders every 15 minutes and turning on IMAP IDLE for multiple accounts, and push on for multiple folders as well, more than our end users would get with the default settings proposed here. Users can turn the feature off if it becomes an issue, but IMAP IDLE was made specifically to reduce the need to sync with a server frequently and reduce both battery usage and network bandwidth. 

As a result, it may reduce battery consumption for our users. IMAP IDLE can reduce periodic polling, but its net battery impact depends on connection stability, network conditions, server behaviour, account count, and incoming-mail volume.

Those who see an increase would likely have an account where they receive many emails, but not all at once, such as automated emails every few minutes. Someone with such a setup could turn off the setting where it exists today. In a future project, we'll make these settings even easier to access.
```

docs/engineering/rfcs/0005-imap-idle-push-on-new-account-inboxes.md
```diff
@@ -0,0 +1,37 @@
+# RFC: Add "Push" Notifications to New Users' Inboxes With IMAP IDLE
```
rafaeltonholo: Since this RFC is more related to how we sync rather than how we notify, I would like to suggest changing its title to:

```suggestion
# RFC 0005: Enable IMAP IDLE ("Push") for the Inbox by Default on Newly Added IMAP Accounts
```

rafaeltonholo (commented):
I have a few points that we should consider, either in this RFC or later in the Tech Design:

1. Thunderbird's current IMAP IDLE implementation depends on the **Alarms & reminders** special access so it can schedule exact alarms used to refresh or recover Push connections. When `canScheduleExactAlarms()` is false, the active per-account Push controllers aren't started; however, the persisted folder Push setting isn't disabled, and `PushService` is still started in the `ALARM_PERMISSION_MISSING` state. We should define when this access is requested, what the user sees when it is unavailable, and whether periodic synchronization is used as a fallback. See [PushController.kt#L166-L182](https://github.com/thunderbird/thunderbird-android/blob/20f083a2b7504944e4cf6ad889bedd497d159c92/legacy/core/src/main/java/com/fsck/k9/controller/push/PushController.kt#L166-L182).

2. Since Push uses a `specialUse` Foreground Service and will be enabled by default, we should review the user-stoppability flow. The current ongoing notification doesn't provide a direct Stop action; tapping it opens another screen from which the user can choose **Disable Push**. Disabling Push persistently changes the user's configuration, which is different from stopping the currently running `PushService` and closing its IMAP IDLE connections while leaving Push enabled so it can run again when the user next opens the app. We should consider a separate **Stop for now** action and define its restart semantics, including preventing alarms or other background events from immediately restarting the service.

3. Before shipping this change, we will need to update the Google Play Console FGS declaration and demonstration video. The current video only shows Push being enabled through **Manage folders > Folder > Enable Push**. After this change, creating an eligible IMAP account will also trigger `PushService` automatically. The declaration and video should show account creation, the ongoing notification, and the available Stop and Disable flows. Otherwise, the declaration may no longer match the released behaviour and could create a Play review or rejection risk.

4. We should explicitly define the relationship between Push, new-message notifications, and periodic synchronization. In particular, should Push still be enabled—and Alarms & reminders access requested—when the user declines new-message notifications? We should also clarify whether enabling Push replaces periodic synchronization for the Inbox, leaves it enabled as a fallback, or intentionally allows both mechanisms to run.

5. The scope should be described as **newly added accounts** rather than **new users**, unless the intention is specifically limited to first-run onboarding. We should also define how imported or restored accounts are handled and ensure that an explicit user decision to disable Push is never overwritten by folder recreation, capability rediscovery, migration, or another account-initialization path.

6. The manifest's `PROPERTY_SPECIAL_USE_FGS_SUBTYPE` description and the user documentation will need to be updated. The current subtype describes the service only as providing instant notifications, while Push also provides near-real-time mailbox synchronization when new-message notifications are disabled. The current documentation also states that Push is disabled by default.

docs/engineering/rfcs/0005-imap-idle-push-on-new-account-inboxes.md
```diff
@@ -24,15 +24,50 @@ believe it's working as it was always intended.
...
+syncing method by default, not fix existing fallback issues. Still, it's something worth examining and ensuring
+that it works as expected, as we'll be rolling this feature out to far more users than before. This feature was not
+easy to find, and therefore only a small percentage of users may have been using it. Now, every new account with
+IMAP IDLE capabilities will have it on by default. That could increase the number of affected users to a level that 
+will require our attention, rather than simply offering support staff or help articles to affected users.
+
+#### User Flow, Google Play, and Documentation
+
+In Google Play, we need to make a declaration for Foreground Services (FGS) we use. "Push" uses a `specialUse` FGS. 
+We define this in the manifest under the name `com.fsck.k9.controller.push.PushService`, a sspecific type of
```
rafaeltonholo: Small typo:

```suggestion
We define this in the manifest under the name `com.fsck.k9.controller.push.PushService`, as a specific type of 
```

rafaeltonholo (commented):
LGTM! Thanks for addressing the comments. There is only a small typo, but much better now!

rafaeltonholo (approved):
LGTM

PR #11413, opened 2026-08-17
fix(TokenCompleteTextView): let the tab key move focus out of the field

### LINKED ISSUE #8231: External keyboard gets stuck on the address fields on Compose screen

snowykte0426:
## Contribution Summary

Linked Issue/Ticket: Fixes #8231
RFC / Technical Design (if applicable): n/a, behaviour agreed in [this comment](https://github.com/thunderbird/thunderbird-android/issues/8231#issuecomment-5316227853)

#### Description

With a hardware keyboard the focus could not leave To, Cc or Bcc.

`TokenCompleteTextView.onKeyDown` consumed the tab key along with enter and dpad centre. `ViewRootImpl` only moves focus for a key that comes back unhandled, so consuming it trapped the focus. Shift+tab kept working, because `hasNoModifiers()` is false for it. That is the "only way out is backwards" from the first comment on the issue.

`AutoCompleteTextView` already gets tab right. It consumes the key while the completion list is open and leaves it alone otherwise, and `TextView` deliberately returns it unhandled so that focus can move. So tab is no longer touched in `onKeyDown`.

It is handled in `onKeyUp` instead. Left alone, `AutoCompleteTextView` completes on tab immediately, which picks a suggestion for the user as soon as there is more than one to pick from. Now: with the list closed the key falls through and the framework moves the focus on. With a single suggestion tab completes it and the focus stays in the field. With several, tab selects into the list, and you walk it with tab or the arrow keys and confirm with enter.

Enter and dpad centre behave as they did.

#### Screen Shots

Verified on an Android 16 emulator with a real tab key (`adb shell input keyevent 61`), and two contacts so the number of suggestions could be controlled. Before, tab did nothing at all. After, it walks To, the Cc/Bcc expander, Subject and then Message text; a single suggestion gets completed; and tab, tab, enter commits the second of two.

Tests are in `TokenCompleteTextViewBehaviorTest`, on the Robolectric harness that was already there. Five of them fail against the current code.

Two things worth knowing. Tab no longer hides the soft keyboard, which used to happen through `handleDone()`. With an external keyboard there is nothing to hide, and once tab lands on Subject you want the keyboard up anyway, so I left it that way; say the word if you would rather have the old behaviour back. The other is that the selected suggestion still is not highlighted. That turned out to be a separate problem in the drop down's styling, and #11416 fixes it.

## AI Disclosure

## Contribution Checklist

- [ ] This contribution is in Kotlin where possible

The Kotlin box is unticked because this changes the vendored `token-auto-complete` widget, which is Java, and its tests sit alongside it.

### DIFF
--- a/library/token-auto-complete/src/main/java/com/tokenautocomplete/TokenCompleteTextView.java
+++ b/library/token-auto-complete/src/main/java/com/tokenautocomplete/TokenCompleteTextView.java
@@ -34,6 +34,7 @@ import android.view.inputmethod.InputConnection;
 import android.view.inputmethod.InputConnectionWrapper;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.Filter;
+import android.widget.ListAdapter;
 import android.widget.ListView;
 import android.widget.TextView;
 
@@ -68,7 +69,7 @@ public abstract class TokenCompleteTextView<T> extends AppCompatAutoCompleteText
     private boolean initialized = false;
     private boolean performBestGuess = true;
     private boolean savingState = false;
-    private boolean shouldFocusNext = false;
+    private boolean completeOnKeyUp = false;
     private boolean allowCollapse = true;
     private boolean internalEditInProgress = false;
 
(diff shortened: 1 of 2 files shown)


rafaeltonholo (approved):
LGTM

PR #11416, opened 2026-08-18
fix(recipient-completion): show which suggestion the keyboard has selected

snowykte0426:
## Contribution Summary

Linked Issue/Ticket: follow-up to #11413 (issue #8231)

#### Description

Now that #11413 has landed, tab selects a suggestion in the recipient completion list. Walking that list with a keyboard barely shows up on screen, though. The first key press leaves no mark at all, and later presses only produce a faint grey fill, so it is hard to tell which suggestion is about to be picked. #11413 makes the list navigable. This makes the selection visible.

The list selector cannot do the job. `setListSelection()` reaches `ListView.layoutChildren`, which positions the selector with `positionSelector(INVALID_POSITION, sel)`, and `AbsListView.positionSelector` then skips the block that refreshes the selector's drawable state, because `position != mSelectorPosition` is false. Its state is the list's own state too, and `DropDownListView` reports itself as focused for as long as the popup exists, so a stateful selector cannot tell a selected row from a cleared one.

The row can. `ListView.setupChild` calls `child.setSelected(selected && shouldShowSelector())`, and that is right from the first selection onwards. So the row background draws the selection, and the list selector is left alone so touch feedback stays as it is.

A word on the colours. `colorSecondaryContainer` is the Material token for a selected list item, but it does not work here: the rows take their text colour from `?attr/textColorPrimaryRecipientDropdown`, which is pinned to black in the light theme and only reaches 3.3:1 on it. `colorSurfaceContainerHighest` keeps the text at 16:1, but on its own it is a 3% lightness step, 1.22:1 against the popup background. So the selected row gets a `colorPrimary` outline as well: 7.7:1 in the light theme and 14.1:1 in the dark one, which meets WCAG 1.4.11 for a state indicator.

#### Screen Shots

The first suggestion selected with one tab press, on an Android 16 emulator.

Before:
11416-before

After, light theme:
11416-after-light

After, dark theme:
11416-after-dark

Verified in both themes. The selected row picks up the fill and the outline on the first key press, follows further presses onto the next row, clears when the list is touched, and tab, tab, enter still commits the second suggestion.

## AI Disclosure

## Contribution Checklist

Two things to flag:

- **This does nothing without #11413.** With only this change, the arrow keys move focus out of the field rather than into the list, so nothing is ever selected. Please merge it after #11413, or tell me to fold it in there.
- **A screen reader still gets nothing.** `setSelected(true)` fires `TYPE_VIEW_SELECTED`, which TalkBack ignores for an unfocused view in a non-focusable window, and focus never enters the popup. An `announceForAccessibility()` of the selected suggestion would fix it, and belongs in #11413. Happy to add it if you want.

One thing to flag. A screen reader still gets nothing, because `setSelected(true)` fires `TYPE_VIEW_SELECTED`, which TalkBack ignores for an unfocused view in a non-focusable window, and focus never enters the popup. An `announceForAccessibility()` of the selected suggestion would fix that, and it belongs in #11413; happy to add it if you want it.

This is a drawable and a layout attribute, so there is nothing a unit test would pin that `ListView` does not already guarantee. I verified it on device instead.

### DIFF
--- a/legacy/ui/legacy/src/main/res/drawable/recipient_dropdown_item_background.xml
+++ b/legacy/ui/legacy/src/main/res/drawable/recipient_dropdown_item_background.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+    Marks the recipient completion suggestion that the keyboard has selected.
+
+    The selection has to be drawn by the row rather than by the list selector. ListView.setupChild() calls
+    child.setSelected() on the row, which is correct from the first selection onwards, whereas the selector's
+    drawable state is the list's own state (AbsListView.getDrawableStateForSelector) and DropDownListView
+    reports itself as focused for as long as the popup exists, so a stateful selector cannot tell a selected
+    row from a cleared one. The list selector is deliberately left alone, so touch feedback is unchanged.
+
+    The fill on its own is only 1.22:1 against the popup background in the light theme, so the outline carries
+    the contrast: colorPrimary is 7.7:1 there and 14.1:1 in the dark theme. colorSecondaryContainer, the
+    Material token for a selected list item, is not usable here because the rows take their text colour from
+    textColorPrimaryRecipientDropdown, which is pinned to black in the light theme and reaches only 3.3:1 on it.
+
+    Note that state_selected also propagates to the row's TextViews. It does not change their colour today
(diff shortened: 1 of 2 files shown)


rafaeltonholo (approved):
LGTM! Thank you for this contribution!

Issue #11348, opened 2026-07-31
"Edit as new mail" after send fail always sets the identity to default identity?


studyingegret:
### Checklist

- [x] I have used the search function to see if someone else has already submitted the same bug report.
- [x] I will describe the problem with as much detail as possible.

### App

Thunderbird for Android

### App version

18.0b2

### Where did you get the app from?

F-Droid

### Android version

12

### Steps to reproduce

1. Make sure you have 2 identities, one of which is the default.
2. Turn off your device's network.
3. Compose and send a mail with an identity that's not the default.
4. Find the mail in the sending box, and choose "edit as new mail".
5. Hit send button.
6. View mail in sending box.

### Expected behavior

The mail uses the identity that's not the default.

### Actual behavior

The mail uses the default identity.

mgw2013:
I guess that's a design choice but I too feel it should assume the original ("copied") mail's addressees.

studyingegret:
@mgw2013 For context I asked on the forum if it's designed behavior but no one seems to know. No answers.

dani-zilla:
I was able to confirm this. Adding a second identity to my account, filling out the email, and finding it in the inbox later does seem to produce the default identity instead of the selected non-default identity. It does seem to pull over the reply to field correctly, but would still send from the initial identity. Good catch!

danielcoderman:
I found this bug in the sent mail folder too. If you click "Edit as new message" for an email you sent with a secondary identity, the default identity is shown instead. The same issue occurs when clicking the message header of an email you sent with a second identity because a bottom sheet is shown that displays the initial identity as the sender.

danielcoderman:
I would like to get permission to work on this issue. I would use a debugger to understand the issue in more detail, and to modify the relevant code.

From my findings so far, the `findIdentity` method in `LegacyAccountDto.kt` is at least part of the problem because it isn't taking into account the `personal` field of the passed in `Address` object, but only the `address` field. What ends up happening is that the first identity that matches the `address` of the `Address` object gets returned, and that's the default identity. There's no chance for a non-default identity to be returned when it has the same email address as the default identity.
```kotlin
  @Synchronized
  fun findIdentity(address: Address): Identity? {
      return identities.find { identity ->
          identity.email.equals(address.address, ignoreCase = true)
          identity.name.equals(address.personal, ignoreCase = true) // new line: takes into account the personal field of the passed in Address
      }
  }
```

rafaeltonholo:
[...]

PR #11432, opened 2026-08-19
fix(identity): use correct sender identity for non-draft messages edited as new messages

danielcoderman:
## Contribution Summary

Linked Issue/Ticket: #11348
RFC / Technical Design (if applicable):

#### Description

This PR fixes issue #11348 by getting the correct sender identity for non-draft messages edited as new messages. It: 
- Adds an else block in the `processDraftMessage` method of `MessageCompose` which assigns the correct sender identity
- Creates a new `getSenderIdentityFromMessage` function in `IdentityHelper`
- Modifies the existing `findIdentity` method in `LegacyAccountDto` to first match an identity by email and name before matching by only email
- Adds unit tests for the new `IdentityHelper` function

Problem:
Messages edited as new messages go through the same code path as draft messages opened for editing, starting from the call to `MessageActions.actionEditDraft` they both make (see `openMessage` and `onEditAsNewMessage` in `MessageHomeActivity`). At some point after this call they both go through the `processDraftMessage` method in `MessageCompose`. Here is where the problem is. In `processDraftMessage` the identity metadata is extracted from the K9 identity header if it's attached to the message being processed. This K9 identity header is only attached to messages saved as drafts which means that since this was the only way of getting the identity metadata, in `processDraftMessage`, non-draft messages ended up using the default identity set in the `onCreate` method of `MessageCompose` (line 497).

Approach:
I added an else block to the if statement in `processDraftMessage` (line 1561) that basically checks if there's an available k9 identity header. In this else block a call to a new function I wrote, `IdentityHelper.getSenderIdentityFromMessage`, is made. This function takes in the account the message belongs to and the message itself as arguments. It then gets the `Address` object corresponding to the sender of the message and passes it to the updated `findIdentity` method which returns the matching identity. The `identity` field in `MessageCompose` is set to this returned matching identity in the else block and from there the rest of the code takes care of updating the screen so that the right identity metadata is displayed.

I noticed the `findIdentity` method didn't take into account the `personal` field of the passed in `Address` object (corresponds to the `name` field of an `Identity` object), so I updated it so that it first tries to match an identity to both email and name, and if that doesn't work then it finds an identity that just matches the email. At first I removed matching by email only but that failed existing tests, and I later found that names for identities aren't required, and that identity names can be edited. That's why I decided to leave the email only matching because it's necessary. For the case where identity names can be edited, the default identity is still returned if the edited identity shares the same email as the default identity. I don't know if that matters, but the only way to get around that with my current approach would be if the `Address` object corresponding to the sender of the message contained some sort of ID that doesn't change. Of course the `Identity` objects would also need this same kind of ID in order for matching to work.

Tests:
I created 3 new tests and a private helper function, in the existing `IdentityHelperTest` file, for the new `getSenderIdentityFromMessage` function which also happen to indirectly test the updated `findIdentity` method. These tests follow the same style of tests in the existing test file, but I can update them to follow the AAA pattern mentioned in the test guide.

I couldn't find an existing test file for `LegacyAccountDto` to directly test `findIdentity` after updating it. I can also add tests for it, but since I couldn't find them I first want to make sure I'm not doing anything wrong. Similarly, I couldn't find existing tests for the `processDraftMessage` method of `MessageCompose`, but I'm not exactly sure how the testing would work there.

Notes:
- The update to `findIdentity` also fixed a bug where clicking the message header of an email you sent with a second identity (sharing the same email as the default identity) showed a bottom sheet displaying the default identity
- The new code in `MessageCompose` solves this issue #11348 which happens to occur for any message edited as a new message, not just for failed messages in the Outbox which are edited as new messages.

## AI Disclosure

## Contribution Checklist

### DIFF
--- a/core/android/account/src/main/kotlin/net/thunderbird/core/android/account/LegacyAccountDto.kt
+++ b/core/android/account/src/main/kotlin/net/thunderbird/core/android/account/LegacyAccountDto.kt
@@ -548,7 +548,12 @@ open class LegacyAccountDto(
 
     @Synchronized
     fun findIdentity(address: Address): Identity? {
+        // Prefer to match by email and name.
+        // Fall back to email only if no identity's name matches (e.g. a renamed identity).
         return identities.find { identity ->
+            identity.email.equals(address.address, ignoreCase = true) &&
+                identity.name.equals(address.personal, ignoreCase = true)
+        } ?: identities.find { identity ->
             identity.email.equals(address.address, ignoreCase = true)
         }
     }
--- a/legacy/core/src/main/java/com/fsck/k9/helper/IdentityHelper.kt
+++ b/legacy/core/src/main/java/com/fsck/k9/helper/IdentityHelper.kt
@@ -37,4 +37,22 @@ object IdentityHelper {
 
         return recipient ?: account.getIdentity(0)
     }
+
+    /**
+     * Find the identity a message was sent from.
+     *
+     * @param account
+     * The account the message belongs to.
+     * @param message
+     * The message to get the sender from.
+     *
+     * @return The identity the message was sent from, or `null` if it couldn't be determined which
+     * identity sent this message.
+     *
+     * @see LegacyAccountDto.findIdentity
+     */
+    @JvmStatic
+    fun getSenderIdentityFromMessage(account: LegacyAccountDto, message: Message): Identity? {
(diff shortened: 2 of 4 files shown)


dani-zilla (approved):
Tested locally and it does fix the issue. Thanks for the contribution!

PR #11435, opened 2026-08-21
fix: keep dark-mode message text readable on light sender backgrounds

### LINKED ISSUE #11381: Text hidden/invisible on emails from Amazon Pay
### Checklist

### App

Thunderbird for Android

### App version

21.1

Google Play

### Android version

Android 15

### Device model

Asus ROG Phone 7 Ultimate

### Steps to reproduce

1. Open an email from Amazon Pay.
2. Notice most of it blank:
Image

As a workaround, select any visible text and drag:
Image
Unformated message option in Thunderbird for Android:
Image

### Expected behavior

Emails should be visible like it is visible in any other client.
Here's how it looks in **Mozilla Thunderbird Desktop**:
Image
Here's how it looks in **Yahoo Webmail**:
Image

### Actual behavior
Some text is invisible/hidden unless selected.

### Logs

_No response_

mvanhorn:
## Contribution Summary

Amazon Pay messages can render white default text over an explicitly white message container when Thunderbird for Android uses dark mode, leaving most content visible only when selected. The message advertises both light and dark color-scheme support while retaining light inline backgrounds, so WebView treats its default foreground as dark-mode content without consistently transforming the sender-defined background.

This updates `DefaultGlobalCssStyleProvider` to read `HtmlSettings.useDarkMode` and, only in dark mode, emit a document-level rule advertising a light-only author color scheme, so the existing WebView algorithmic-darkening path transforms foregrounds and backgrounds as one unit. The rule is document-scoped and carries no Amazon-specific selectors, colors, or markup, so any message that claims dark-mode support while keeping light containers gets the same treatment. Light mode emits no color-scheme override, and the existing global layout, selection, blockquote, and wrapping rules are unchanged.

Closes #11381

#### Description

Three added lines in `DefaultGlobalCssStyleProvider` (`commonMain`), plus tests covering the dark-mode and light-mode branches in `DefaultGlobalCssStyleProviderTest`.

#### Screen Shots

N/A: I could not capture device screenshots for this change. The effect is a document-level `color-scheme` declaration in the generated reader CSS; the added tests assert the emitted CSS for both the dark-mode and light-mode branches.

## AI Disclosure

## Contribution Checklist

- [ ] This contribution adheres to the existing codestyle (run `gradlew spotlessCheck` to check and `gradlew spotlessApply` to format your source code; will be checked by CI). Not verified locally: `spotlessCheck` was not run in my environment, so I am relying on CI for this.
- [ ] This contribution does not break existing unit tests (run `gradlew testDebugUnitTest`; will be checked by CI). Not verified locally: `testDebugUnitTest` was not run in my environment, so I am relying on CI for this.
- [ ] This contribution adheres to our [Engineering process](https://github.com/thunderbird/thunderbird-android/tree/main/docs/engineering) (RFC/Technical Design/ADR). Not claimed: no RFC/ADR was filed for this bug fix; happy to follow up if one is expected.

AI was used for assistance.

### DIFF
--- a/feature/mail/message/reader/impl/src/androidHostTest/kotlin/net/thunderbird/feature/mail/message/reader/impl/css/DefaultGlobalCssStyleProviderTest.kt
+++ b/feature/mail/message/reader/impl/src/androidHostTest/kotlin/net/thunderbird/feature/mail/message/reader/impl/css/DefaultGlobalCssStyleProviderTest.kt
@@ -2,6 +2,7 @@ package net.thunderbird.feature.mail.message.reader.impl.css
 
 import assertk.assertThat
 import assertk.assertions.contains
+import assertk.assertions.doesNotContain
 import kotlin.test.Test
 import net.thunderbird.core.common.mail.html.HtmlSettings
 import net.thunderbird.feature.mail.message.reader.api.css.CssClassNameProvider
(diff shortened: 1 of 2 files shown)


rafaeltonholo (changes requested):
This is not the way to fix the issue.

Although this "fixes" the issue for the Amazon Pay email, it breaks all other emails on dark mode, making them always render in light mode.

| Amazon Pay email | text/plain email |
| :---: | :---: |
|  |  | 
| **text/html email** | **Email with complex HTML** |
| Image |  | 

This is the reason we ask for screenshots: to make sure you have properly tested the change you have made.

PR #11452, opened 2026-08-25
docs(database): add ADR 0010 for project-wide Repository pattern adoption

### LINKED ISSUE #11440: Global Database 0.1.6 - Exploration - Find and list all the DB entrypoints to create new agnostic Repositories
We need to decouple the existing code from the database so we can migrate to a new database with minimal trouble.

This task is meant to investigate how we retrieve and mutate data from the database and define how database access should be handled; ideally using the Repository pattern.

The result of this task will be an ADR that will define the standard for how we fetch and mutate data within a database.

Points we need to ensure:

- Define the Repository API
- List the files that are written in Java to migrate to Kotlin that would be necessary for the migration
- Make sure that the new API is pagination-friendly

rafaeltonholo:
## Contribution Summary

Linked Issue/Ticket: Closes #11440

#### Description

Add a new ADR to consolidate the Repository pattern to decouple the business logic from the data layer

## AI Disclosure

## Contribution Checklist


docs/architecture/repository-pattern.md
```diff
@@ -0,0 +1,262 @@
...
+
+A repository is a domain-facing contract that hides how data is read, stored, or synchronized. Its implementation may
+coordinate local storage and remote services, but UI state and UI-specific business logic belong elsewhere.
+
+Split contracts by responsibility before grouping them by aggregate. Do not create one interface per database table or
+one interface per SQL operation. An optional aggregate-named facade may compose focused contracts, but it must not add
+methods. A caller depends on the narrowest contract it needs.
+
+An aggregate is a group of related domain data that changes together. A facade is an interface that combines focused
+contracts for callers that need all of them.
```
dani-zilla: I do wonder if the more precise definition of aggregate and facade should appear before the instructions on how to properly use them. 

Some definition of contracts and how we use them between the API and Implementation could also be useful. Not everyone who may wish to contribute to this project will come from years of software engineering experience, we'll likely get a few curious beginners eager to learn.

docs/architecture/repository-pattern.md
```diff
@@ -0,0 +1,262 @@
...
+This makes the scope visible to callers and supports a consolidated database with one connection for all accounts.
+
+## Asynchronous work and results
+
+Repository I/O uses coroutines:
+
+- A one-shot operation is a `suspend fun`.
+- A reactive operation returns a `Flow`.
+
+Choose the return type based on the contract rather than the current implementation. An expected recoverable failure is
```
dani-zilla: This seems to be about a specific case where modifying the api to change the contract, making the new implementation match, rather than basing it off of existing repo functions, right?

docs/architecture/repository-pattern.md
```diff
@@ -0,0 +1,262 @@
...
+a normal condition that the caller can handle, such as unavailable storage, a network failure, a rejected operation, or
+a missing required entity. Return a value directly only when the contract can always provide a meaningful result,
+possibly by using a default.
```
dani-zilla: Would we want a default over an error, even a generic one? Unless the return type specifically only uses a default on error, it could lead to misleading returned data. Unless I'm misunderstanding the intention here.

dani-zilla: Seeing the default example below, with it loading default settings if none are available, does make sense. I suppose that, with the below context, this may be clear enough.

docs/architecture/repository-pattern.md
```diff
@@ -0,0 +1,262 @@
...
+
+Use `create()`, `update()`, and `delete()` for aggregate lifecycle operations. Do not add a generic `save()` method to
+new contracts. Use `clear<Scope>()` only for explicitly scoped bulk removal, such as `clearCache()`.
+
+Within a focused repository contract, method names describe the operation and must not repeat the aggregate named by the
+repository. For example, use `DraftRepository.create()`, `FolderQueryRepository.findById()`, and
+`FolderQueryRepository.findByServerId()`, not `createDraft()`, `findByFolderId()`, or `findFolderByServerId()`. Use
+`FolderQueryRepository.getById()` or `FolderQueryRepository.getByServerId()` for required lookups.
+
+Avoid repository methods whose names start with `set`. Multiple independent setters expose invalid intermediate state
```
dani-zilla, line 146: Definitely a good point to make 👍

docs/engineering/adr/0010-adopt-project-wide-repository-pattern.md
```diff
@@ -0,0 +1,53 @@
+# ADR 0010: Adopt project-wide Repository pattern
+
+- Issue: [#11440](https://github.com/thunderbird/thunderbird-android/issues/11440)
+- Status: **Proposed**
+
+## Context
+
+Data access is spread across the codebase and is often coupled to business logic. The current per-account database
+model makes cross-account features, including unified inbox pagination and unified folders, difficult to implement and
+prevents a gradual migration to a consolidated database.
```
dani-zilla, line 10: This really does lay out why this is such an important guideline succinctly 👍

docs/architecture/repository-pattern.md
wmontwe: I added a core concepts section before this and moved the definitions there.

docs/architecture/repository-pattern.md
wmontwe: Not specifically. The intended rule is broader: design the repository contract around domain semantics and caller needs, not around what the current database, network API, or legacy repository conveniently returns.

I rewrote the paragraph.

docs/architecture/repository-pattern.md
```diff
@@ -0,0 +1,278 @@
...
+    suspend fun disable(
+        accountId: AccountId,
+    ): Outcome<Unit, FolderError>
+}
+
+interface FolderRepository :
+    FolderQueryRepository,
+    RemoteFolderRepository,
+    FolderPushTrackingRepository,
+    FolderSettingsRepository
```
rafaeltonholo: There is one problem with the method naming here:

If we want to enable the aggregate, we can't simply use `find<Identifier>` or `observer<Identifier>`. The reason is that it will potentially create implementation conflicts since we could easily end up with two interfaces with the same method name but different returns.

A concrete example is the current `FolderRepository`. If we split it into smaller repositories, we would end up with the following:

```kotlin
interface RemoteFolderRepository {
    suspend fun findAllByAccountId(accountId: AccountId): Outcome<List<RemoteFolder>, FolderError>
}

interface RemoteFolderDetailsRepository {
    suspend fun findAllByAccountId(accountId: AccountId): Outcome<List<RemoteFolder>, FolderError>
}

// ...other repositories

interface FolderRepository: 
    RemoteFolderRepository, 
    RemoteFolderDetailsRepository,
    ...other repositories

```

For the API definition, it isn't a problem, and Kotlin will not complain at the interface level. However, now we have two problems:

1. When using the aggregate directly (`FolderRepository`), only one of the two methods defined above will be visible
2. We can't have a concrete implementation of `FolderRepository` since we have two methods with the same name, but different return types; Kotlin won't be able to understand that and will show the error: `Class 'DefaultFolderRepository' must override 'findAllByAccountId' because it inherits multiple implementations for it.`

We could address that issue by allowing `findAll<concern>By<identifier>`, but that could create a verbose solution.

dani-zilla: It may be a bit verbose, but verbose function names isn't something I'd shy away from. Verbose in boilerplate code, that sort of thing, can be problematic just for the sheer amount of knowledge and changes required for basic changes. But I don't see a longer name as introducing complexity, so, while it's more verbose, I don't mind it.

rafaeltonholo: Yes, but the question we need to answer is: Do we need an aggregate repository, or is it only required for a slow migration?

I lean towards the second, so I would say no to an aggregate repository and keep the name without the repetition.

dani-zilla: Sorry, it seemed like there were two options here: Use the aggregate repository (`FolderRepository`, in this example), which would potentially have two functions, one for local and one for remote folders, that both have the same function name, such as `FindAllById`, that could apply to remote or local folder repositories. The solution here being something like `findAllRemoteFoldersById` and `findAllLocalFoldersById`? I could also see, in this case, the FolderRepository returning both Folder types, with some indication if they are remote or local in the folder objects themselves.

So, I see it as the aggregate could call both and merge them, or the naming could be more complex. I think either should be permitted, depending on use case. 

Unless I've misunderstood where the problem is here.

wmontwe: I think the aggregate is not needed. I added it to bridge legacy code, but looking at it again, it's establishing a pattern that isn't great.

rafaeltonholo: > I think the aggregate is not needed. I added it to bridge legacy code, but looking at it again, it's establishing a pattern that isn't great.

I second that. The stack PR #11478 exercises this ADR and the aggregate repository is not needed at the end of the changes. I think we would be fine without it.

wmontwe: ADR is updated

rafaeltonholo (commented):
I can't approve since I created the PR, but the changes LGTM

dani-zilla (approved):

Issue #11104, opened 2026-06-04
Global Database 0.1.2 - Write RFC


wmontwe:
Prepare an RFC for the Global Database.

Turn the spike results into a production architecture.

PR #11471, opened 2026-08-31
docs(rfc): global database

wmontwe:
## Contribution Summary

Linked Issue/Ticket: resolves #11104

#### Description

This adds the RFC for the global database milestone, proposing to use Room3 as the database backend to replace the existing per account databases.

## AI Disclosure

Select **one** of the following (mandatory)

- [ ] This contribution does not include any changes created or assisted by AI.
- [x] This contribution includes changes assisted by AI. -> proof read the RFC and suggested edits
- [ ] This contribution includes changes created by AI.

Issue #11468, opened 2026-08-31
Global Database 0.1.11 - Cut over legacy mail storage to the global database


wmontwe:
Implement the production transition from legacy per-account mail storage to the global mail database. Import durable mail data, validate the result, and establish the durable cutover state that makes global storage authoritative.

#### Acceptance criteria

- Import covers every mapping in the accepted legacy inventory.
- Derived data, such as search indexes, is rebuilt before validation runs.
- Validation covers imported data, relationships, queued operations, attachments, restart integrity, and search.
- An attachment that fails to copy or validate is recorded in the migration result and fails the migration.
- Before cutover, normal mail access cannot observe imported global data.
- Failed migration leaves legacy storage authoritative.
- Cutover removes legacy database and attachment artifacts.
- After cutover, all mail access uses global storage and cleanup failure does not restore legacy authority.
- If the migration fails due to insufficient storage, the system notifies the user to disable auto-updates and postpone manual version upgrades until sufficient space is cleared.

PR #11472, opened 2026-08-31
docs(technical-design): global database

### LINKED ISSUE #11293: Global Database 0.1.3 - Write technical design
Translate the accepted RFC into a technical design defining the detailed database foundation and first mail persistence slice.

The RFC delegates these questions to the technical design:

- The table-by-table migration inventory, recording which legacy tables are durable, derived, or excluded
- Module placement and driver configuration
- Global identifiers
- The migration screen
- Source-schema support
- The cleanup retry strategy

wmontwe:
## Contribution Summary

Linked Issue/Ticket: resolves #11293 
RFC / Technical Design (if applicable): #11104 

#### Description

Technical design for the Global Database

## AI Disclosure

- [x] This contribution includes changes assisted by AI. -> proof read the technical design and suggested edits


docs/architecture/database.md
```diff
@@ -0,0 +1,91 @@
...
+
+- **Mail repository contracts** are the only storage boundary visible to mail-domain callers. They follow the
+  [Repository pattern ADR](../engineering/adr/0010-adopt-project-wide-repository-pattern.md): they are focused,
+  explicitly account-scoped, and do not expose Room, SQL, cursors, files, or legacy `LocalStore` types.
+- **Repository implementations and local data sources** belong to the mail feature's internal persistence
+  implementation. They own mappings between domain data and both global and legacy representations.
+- **The global mail database** owns the legacy-compatible mail schema, migration metadata, and derived data.
+  It is one physical database for the application, not one database per account.
+- **The attachment store** remains file-backed for content that legacy already held on disk, which is every body part
+  above a size threshold. Parts at or below it stay `message_parts` BLOBs, as in legacy. Database records and attachment
```
rafaeltonholo, line 45: > Parts at or below it stay `message_parts` BLOBs, as in legacy.

I wonder what the benefits are of having two sources for attachments. Why not store everything as files on disk instead of storing them as BLOBs in the database?

docs/engineering/rfcs/0007-global-database.md
```diff
@@ -0,0 +1,142 @@
+# RFC 0007: Global Database
+
+- Issue: [#11104](https://github.com/thunderbird/thunderbird-android/issues/11104)
+- Technical design: [Global Database](../technical-designs/0003-global-database.md)
+- Repository pattern: [ADR 0010 proposal](https://github.com/thunderbird/thunderbird-android/pull/11452)
+- Portable data format: [RFC 0008: Portable Profile Data Format](0008-portable-profile-data-format.md)
+- Status: **Proposed**
```
rafaeltonholo: Since this is a stacked PR, and this PR will only be merged once the #11471 is reviewed and accepted, you could change the Status here to accepted.

docs/engineering/technical-designs/0003-global-database.md
```diff
@@ -0,0 +1,100 @@
...
+## Global mail store
+
+The global database keeps the legacy mail-store schema and behavior as its starting point. It stores all accounts in one
+database, keeps account-qualified legacy references, and adds global identifiers only where callers need them.
+
+The [legacy table inventory](0003-global-database/legacy-table-inventory.md) defines the exact mapping. It includes
+downloaded attachments. Legacy keeps a body part on disk above a size threshold and as a `message_parts` BLOB at or
+below it. The migration preserves that split rather than changing where content lives.
+
+## Migration
```
rafaeltonholo, line 88: We need to ensure that the migration process can take place even when the app is running in the background. Users with large email accounts may require minutes or even hours to migrate all their data.

It's important to recognize that we cannot rely on users to keep the app in the foreground while the migration is occurring.

I recommend starting the migration process as a data synchronization foreground service. This will notify the user when the migration is complete, and we can keep the progress updated through a persistent notification.

If the user remains in the app while the migration is ongoing, the screen should continuously update to reflect the migration's progress. However, if the user switches to the background or closes the app, we must ensure that the migration continues to run until it is fully completed.

docs/architecture/database.md
wmontwe: I wanted to keep the existing behavior, if it's not too much work, the migration could change that too, but adds complexity.

docs/engineering/technical-designs/0003-global-database.md
wmontwe: I added a section for that.

dani-zilla: Along with that, I'd say we want to figure out _when_ we'll tell users to update. Perhaps we can run a few tests of our own and figure out, in general, how long a transfer might take on a midrange device. Obviously it would be hard to predict this with any accuracy unless we test on a great deal of different hardware and DB sizes, but we may be able to tell them something like, "This migration could take an hour, would you prefer to be reminded tonight?" 
The goal would be that users never notice the down time. They tap the notification we may send them at a chosen time, go to bed, and wake up to their app using the new database with the report and success message. 
I don't think this needs to be detailed here, necessarily, but we should consider a possible stretch goal of scheduling the upgrade so it can happen when the user isn't going to be using the app.

docs/engineering/technical-designs/0003-global-database.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Pre-flight storage check
+
+The pre-flight check runs as the first foreground migration phase, before creating the unpublished database. It
+inventories the legacy databases and file-backed attachments, then calculates the additional space needed while legacy
+and global storage coexist. Required headroom includes the estimated global database, copied attachments, rebuilt search
+data, SQLite transaction and temporary-file growth, and a safety margin. If an RFC 0008 archive is written to the same
+volume, its estimated size is also included.
+
+Migration starts only when available space on the target volume meets the calculated headroom. If it does not, the
```
dani-zilla, line 125: We may want to inform users who can't do the update not to upgrade to a newer version of the app and turn off auto-update until they can. I doubt we'd have many users who could not transfer their database over but also have auto update turned on, a feature that could increase the storage space used, but it's still perhaps worth mentioning to users who can't update the DB due to storage space.

dani-zilla (approved):

docs/engineering/technical-designs/0003-global-database.md
wmontwe: I added this to #11468

PR #11491, opened 2026-09-03
fix(ui) changed drawer to follow expected behavior for auto closing

### LINKED ISSUE #11461: Sidebar account selection issue
### Checklist

### App

Thunderbird for Android

### App version

22.0

Google Play

### Android version

Android 12, rooted (not that it matters)

### Device model

Moto XT2215-4

### Steps to reproduce

Over time, the sidebar mechanism has greatly improved and is now (v22.0) nearly perfect. However I cannot help but notice 2 problems:

Problem 1:
1. Open sidebar, tap account drop down, tap the currently selected account.

2. The sidebar disappears as expected and you see the account's inbox as expected.

3. The problem is seen when sidebar is next opened. It shows account drop down, not account's folders - until you tap to close the drop down.

This differs from the much better/logical behavior seen if any other account is selected in 1st step.

Problem 2:
One even smaller matter is that the icon in the account drop down is unresponsive to tap. in order to select you must tap the text. 

### Expected behavior

Problem 1: Selecting an account is selecting an account - no matter, in this case, if it has already been selected. The same behavior should prevail - close the drop down and close the the sidebar so that when the sidebar is next opened, the account's folders are shown.

Problem 2: the account drop down is a list of entries to select an account. As tapping the icon rather than the text has no special semantics/behavior, treat tapping the icon in the same way as tapping the text.

Thanks for all!

### Actual behavior

I think I put this above in the context.

### Logs

_No response_

itsSiide:
-Added a state change to selection of an account
-Changed the AvatarLayout click event to make the whole box clickable for account selection 
-Added a new test to confirm drawer close behavior

## Contribution Summary

Linked Issue/Ticket:  Fixes #11461

#### Description

The drawer previously properly closed itself after selecting a user account but the expanded account selection was still expanded even after the drawer has already been closed. I fixed this by setting the state correctly before emitting the drawer close event. This only applies when the auto select folders account setting is set to anything but "none". If "none" is selected the account selection closes and the drawer still remains open.

I also adjusted the second stated Problem with the avatar icon not being clickable while selecting an account. This was the case due to the AvatarLayout always setting an onClick event even though it may be initialized without one by the caller. This caused the icon to just take up the click and doing nothing with it. I changed the modifier to only be set when an onClick is explicitly given.

A Test to confirm the changed account selection drawer behavior was added as well.

#### Screen Shots

Nothing visual was changed, just ui behavior.

## AI Disclosure

## Contribution Checklist

### DIFF
--- a/feature/navigation/drawer/dropdown/src/test/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/DrawerViewModelTest.kt
+++ b/feature/navigation/drawer/dropdown/src/test/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/DrawerViewModelTest.kt
@@ -289,6 +289,34 @@ internal class DrawerViewModelTest {
             isEqualTo(Effect.CloseDrawer)
         }
     }
+    @Test
+    fun `should not emit CloseDrawer effect after emitting OpenAccount if AutoExpandFolder is None`() = runMviTest {
+        val displayAccounts = createDisplayAccountList(1) + createDisplayAccount(
+            id = "uuid-1",
+            hasAutoExpandFolder = false,
+        )
+        val getDisplayAccountsFlow = MutableStateFlow(displayAccounts)
+        val testSubject = createTestSubject(
+            displayAccountsFlow = getDisplayAccountsFlow,
+        )
+        val turbines = turbinesWithInitialStateCheck(
+            testSubject,
+            State(
+                accounts = displayAccounts.toImmutableList(),
+                selectedAccountId = displayAccounts.first().id,
+            ),
+        )
+        advanceUntilIdle()
+
+        testSubject.event(Event.OnAccountClick(displayAccounts[1]))
+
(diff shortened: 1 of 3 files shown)


feature/navigation/drawer/dropdown/src/test/kotlin/net/thunderbird/feature/navigation/drawer/dropdown/ui/DrawerViewModelTest.kt
```diff
@@ -290,6 +290,35 @@ internal class DrawerViewModelTest {
         }
     }
 
+    @Test
```
wmontwe, line 293: The initial state leaves `showAccountSelection` at its default `false`, and the test only verifies that `CloseDrawer` is not emitted. But it doesn't verify that `showAccountSelection = true` is changed to `false`.

wmontwe (commented):
@itsSiide Thank you! Just a tiny test improvement left.

wmontwe (approved):
Thank you for your fix!

PR #11499, opened 2026-09-03
chore(ci): update merge script to match b0 beta suffix coming from main

### LINKED ISSUE #11500: merge_gradle.py script always flags app-thunderbird/build.gradle.kts as a conflict due to versionNameSuffix mismatch during main <> beta merges
`merge_gradle.py` searches main's `app-thunderbird/build.gradle.kts` for `versionNameSuffix = "b[1-9]"` ([here](https://github.com/thunderbird/thunderbird-android/blob/main/scripts/ci/merges/merge_gradle.py#L68-L75)), but main sends `"b0"` ([here](https://github.com/thunderbird/thunderbird-android/blob/main/app-thunderbird/build.gradle.kts#L120)), so the search never matches and the beta path raises `SystemExit`, which causes git to treat it as a conflict.

The "Bump version suffix" step in `shippable_builds.yml` increments the suffix per beta release, so the merge result has to start at "b0".

kryoseu:
Fixes: https://github.com/thunderbird/thunderbird-android/issues/11500

## Problem

The beta path of `merge_gradle.py` searched `theirs` (main's `app-thunderbird/build.gradle.kts`) for `versionNameSuffix = "b[1-9]"`.  Since `main` carries `b0`, which `b[1-9]` never matches, so the search failed and the path raised `SystemExit`, causing git to mark the file as conflicted, so every `main -> beta` merge needed manual fixing.

## Fix

Search for `b\d+` (e.g b1, b2, b11, etc) so main's value matches whatever it is, and normalize the merge result to `b0` for both apps. The "Bump version suffix" step in `shippable_builds.yml` bumps the suffix per Thunderbird beta release, so the merge result has to start at `b0` for the first beta release to produce `b1`. We don't ship betas for k9, and the beta suffix is dropped on a release. 

I've created a `set_version_name_suffix()` helper to be used for both branch logic, and two new constants: `BETA_SUFFIX` and `MAIN_SUFFIX`.

Three details:

- The search now targets `ours` rather than `theirs`. `shutil.copyfile(theirs, ours)` runs first, so the two are identical in content, but `ours` is the file being patched and the one whose final state matters.
- `SystemExit` is now reserved for the case where the searched suffix line doesn't exist at all: `a1` on the app-k9mail path, `b\d+` on app-thunderbird.
- The release path uses `b\d+` too, replacing `b[1-9]`, so it no longer breaks once a cycle reaches `b10`.

## Testing

Ran the driver directly against the pre-merge branch tips:

| | ours (beta) | theirs (main) | result |
|---|---|---|---|
| app-thunderbird | 23.0 / code 58 / `b1` | 24.0 / code 4 / `b0` | 24.0 / code 58 / `b0` |
| app-k9mail | 23.0 / code 39021 / `b1` | 24.0 / code 39004 / `a1` | 24.0 / code 39021 / `b0` |

### DIFF
--- a/scripts/ci/merges/merge_gradle.py
+++ b/scripts/ci/merges/merge_gradle.py
@@ -8,6 +8,9 @@ import sys
 ours = sys.argv[1]
 theirs = sys.argv[2]
 
+BETA_SUFFIX = r"versionNameSuffix = \"b\d+\""
+MAIN_SUFFIX = r"versionNameSuffix = \"a1\""
+
 
 def get_current_branch():
     result = subprocess.run(
@@ -43,6 +46,24 @@ def replace_matching_line(file_path, search_term, new_line):
                 file.write(line)
 
 
+def set_version_name_suffix(file_path, search_term, suffix):
+    """Rewrites the versionNameSuffix line matching search_term.
+
+    Sets the suffix to `suffix`, or drops the line entirely when `suffix` is
+    None. Raises if the line is missing.
+    """
+    found_line = find_matching_line(file_path, search_term)
+    if not found_line:
+        raise SystemExit(f"Search term '{search_term}' not found in merge result.")
+    if suffix is None:
+        replace_matching_line(file_path, search_term, "")
+        return
+    if f'"{suffix}"' in found_line:
+        return
+    new_line = '{}= "{}"\n'.format(found_line.split("=")[0], suffix)
+    replace_matching_line(file_path, search_term, new_line)
+
+
 branch = get_current_branch()
 
 search_term = "com.fsck.k9"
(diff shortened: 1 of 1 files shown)


scripts/ci/merges/merge_gradle.py
```diff
@@ -63,16 +63,23 @@ def replace_matching_line(file_path, search_term, new_line):
 
 if branch == "beta":
     if is_k9:
```
coreycb: We might actually be able to remove the is_k9 code block since we don't release beta for k9. Or if you want to keep it you can probably re-use more of the code across k9/thunderbird paths below.

scripts/ci/merges/merge_gradle.py
```diff
@@ -63,16 +63,23 @@ def replace_matching_line(file_path, search_term, new_line):
...
+        replace_matching_line(ours, search_term, new_line)
     else:
-        search_term = r"versionNameSuffix = \"b[1-9]\""
-    found_line = find_matching_line(theirs, search_term)
-    if found_line:
-        if "b1" not in found_line:
-            new_line = "{}{}\n".format(found_line.split("=")[0], '= "b1"')
+        # beta always starts at "b0"; the shippable build workflow
+        # bumps the suffix per beta release.
+        search_term = r"versionNameSuffix = \"b\d+\""
```
coreycb: We might be able to get away with `b\d` since I think we generally don't get past b3 or b4. Or maybe `b\d{1,2}(?!\d)` if we want to include 2 digits.

scripts/ci/merges/merge_gradle.py
```diff
@@ -63,16 +63,23 @@ def replace_matching_line(file_path, search_term, new_line):
...
+        found_line = find_matching_line(ours, search_term)
```
coreycb: Should this be `theirs`?

scripts/ci/merges/merge_gradle.py
```diff
@@ -63,16 +63,23 @@ def replace_matching_line(file_path, search_term, new_line):
...
+        if not found_line:
+            raise SystemExit(f"Search term '{search_term}' not found in merge result.")
```
coreycb: Maybe 'not found in theirs file'?

scripts/ci/merges/merge_gradle.py
kryoseu: At line 57 `theirs` gets brought over to `ours`:

```
shutil.copyfile(theirs, ours)
```

So either works, but I think it reads better checking against the post merge results (beta - ours) instead of main's (theirs).

scripts/ci/merges/merge_gradle.py
kryoseu: By the time `find_matching_line()` is called at like 77, `ours` is the merge result (since `shutil.copyfile(theirs, ours)` is called at line 57. I think the current message correctly reflects that.

scripts/ci/merges/merge_gradle.py
kryoseu: The current regex (with `d+`) does that:

```
>>> import re
>>> pat = r'versionNameSuffix = \"b\d+\"'
>>> for line in ['versionNameSuffix = "b1"', 'versionNameSuffix = "b2"', 'versionNameSuffix = "b11"', 'versionNameSuffix = "a1"']:
...  print(line, '->', bool(re.search(pat, line)))
...
versionNameSuffix = "b1" -> True
versionNameSuffix = "b2" -> True
versionNameSuffix = "b11" -> True
versionNameSuffix = "a1" -> False
```

scripts/ci/merges/merge_gradle.py
kryoseu: If we remove the `is_k9` code block, beta will inherit `a1` from main. While that's fine for `beta`, it will flow into the next `beta` -> `release` merge and break the release path in the script.

kryoseu: I will update the script to re-use more of the code though.

scripts/ci/merges/merge_gradle.py
coreycb: That's fine. I was just thinking it might be too much because we'll never have a b000009. Should be fine though.

scripts/ci/merges/merge_gradle.py
coreycb: Yeah that makes sense

coreycb (approved):
Looks good, thank you

PR #11514, opened 2026-09-08
fix(funding): center last row of contribution items

### LINKED ISSUE #8355: The lowest value contribution is displayed as a single big button on Tablet devices while in landscape
### Checklist

### App

Thunderbird for Android

### App version

8.0b3

Google Play

### Android version

14

### Device model

Galaxy Tab A9+

### Steps to reproduce

Preconditions:
Have at least one account added.
Have auto-rotate activated

Steps:

Launch TB.
1. Go to Hamburger Menu
2. Tap Show accounts
3. Tap the Settings cogwheel.
4. Tap Support Thunderbird Beta
5. Flip device in landscape mode

### Expected behavior

All contribution values are displayed symmetrically as displayed in [Figma design documentation](https://www.figma.com/design/BWHQ2Dd3s0Ejrsb9d25EEx/TBAndroid_2024_Q3_Appeal_v01.0.0?node-id=2135-98&node-type=frame&t=xii3dnFhgREk2Gxa-0) while device is in Landscape mode.

### Actual behavior

The lowest contribution value is displayed as a long button while device is in landscape mode.
This issue is also present in the small screen of the Gagaxy Fold 4 device(landscape)

### Logs

Screenshot_20241017_161506_Thunderbird Beta

khaledhaltam:
## Contribution Summary

Linked Issue/Ticket: Fixes #8355

#### Description

The contribution items are laid out in a `FlowRow` where each item carries `Modifier.weight(1f)`. Weights are distributed inside a row, so the items shrink to share whatever width the row is given instead of wrapping at their natural size. That produces two visible defects:

- The leftover item of an incomplete last row receives the full remaining width and stretches across the screen. This is what the issue reports.
- Before wrapping happens at all, the items are squeezed below the width their label needs, so a price breaks mid-number and `250 $` renders on two lines. This is not mentioned in the issue but has the same cause.

This replaces the `FlowRow` with a `Layout` that:

- derives one column width from the intrinsic width of the widest item, so items are never squeezed below what their label needs,
- uses that same column width for every row, so an incomplete last row keeps the width of the rows above it,
- centers a row that is not completely filled,
- hands the remainder of the width division to the leftmost columns, so a completely filled row still uses the available width exactly and keeps the pixel layout it had before.

**On the issue title.** The report names a tablet in landscape and the Fold's cover display, but neither the device nor the orientation decides. What decides is the available width relative to the number of items and the length of the price labels. That is why the same layout appears on a phone at certain widths, and why the reporter hit it on a tablet where the shorter sample amounts would not trigger it.

**On the earlier suggestion.** @kewisch proposed capping the items per row in 2024 and was unsure whether that would hold up with different currencies and longer numbers. On its own it does not: with four amounts, a cap of three produces 3 + 1 and the one stretches again. Centering the incomplete row fixes it independently of how many items fit, which is the direction @wmontwe gave when assigning the issue, so no cap is added here. If a cap is wanted for design reasons, that would be a separate change.

**Known limitation.** The screen needs Google Play Billing for real amounts, which requires maintainer level access, so I could not verify it with live data. The layout code does not depend on where the amounts come from, and the preview renders the same production composable with sample data.

#### Screen Shots

Compose preview on a Pixel 10 Pro, widths set with `adb shell wm size`.

| Width | Before | After |
|---|---|---|
| 320 dp | 3 + 3, correct | unchanged |
| 390 dp | 4 + 2, the amounts above break mid-number, the two below are twice as wide | 3 + 3, equal width, no line break |
| 500 dp | all six squeezed into one row, every amount wrapped onto two lines | 5 + 1, the last item at normal width and centered |

##### Width: 320 dp
Before: 
before-320dp
After: 
after-320dp

##### Width: 390 dp
Before: 
before-390dp
After: 
after-390dp

##### Width: 500 dp
Before: 
before-500dp
After: 
after-500dp

#### Testing

To reproduce and verify manually:

1. Run `ContributionListOneTimePreview` from `ContributionListPreview.kt` on a device or emulator via Run Preview
2. Set the width with `adb shell wm size 1500x1100` (500 dp at 3x density)
3. Restart the preview so Compose re-lays out, otherwise a stale layout is shown
4. Before the change: all six amounts sit in one row, each label wrapped onto two lines. After the change: five amounts in the first row, the sixth at the same width and centered below them

`ContributionListKtTest` covers ten cases: the two defect layouts (4 + 2 and 5 + 1) for both equal width and centering, a set of longer price labels that reproduces the reporter's situation, a completely filled row as evidence that the unaffected case does not change, a single amount, two amounts, the recurring amounts on the second tab, and a narrow screen where every item has to stay above the minimum touch target size.

Four of these ten fail on the parent commit and pass with the change:

```
should not stretch the items of a partially filled last row
  expected <140.0f> to be close to <66.0f> with delta of <1.0f>
should not stretch a single item in the last row
  expected <328.0f> to be close to <60.0f> with delta of <1.0f>
should center the last row of the recurring contributions as well
  expected <140.0f> to be close to <66.0f> with delta of <1.0f>
should center the last row when longer price labels reduce the items per row
  expected [size]:<[2]> but was:<[1]>
```

The two pure centering assertions pass on the parent commit as well, which is expected: an item stretched across the full width is trivially centered. They are kept as regression guards.

For side effects I ran a temporary sweep over 204 combinations, 51 widths from 200 dp to 1200 dp against four label lengths from `5 $` to `1.234.567,89 CHF`, asserting per combination that no item is taller than a single line, that all items share one width, that every row is centered in the content area, and that completely filled rows are flush with the edges. The change reports 0 findings, the parent commit reports 25. On a device I checked twelve widths from 280 dp to 952 dp, landscape, and font scale at 200%, all correct. The change can produce one row more than before at some widths, so the screen gets taller; `ContributionContent` already wraps the content in `verticalScroll`, so this is absorbed.

Commands run, all passing:

```
./gradlew :feature:funding:googleplay:testDebugUnitTest --tests "*ContributionListKtTest*"
./gradlew test lint detekt spotlessCheck testsOnCi dependencyGuard :quality:konsist:test
./gradlew check
```

`./gradlew check` was re-run after rebasing onto `upstream/main`. `connectedAndroidTest` was not run because the repository contains no `androidTest` source set, so the task has no tests to execute.

## AI Disclosure

- [] This contribution includes changes assisted by AI.

## Contribution Checklist

### DIFF
--- a/feature/funding/googleplay/src/main/kotlin/net/thunderbird/feature/funding/googleplay/ui/contribution/list/ContributionList.kt
+++ b/feature/funding/googleplay/src/main/kotlin/net/thunderbird/feature/funding/googleplay/ui/contribution/list/ContributionList.kt
@@ -4,8 +4,6 @@ import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.ExperimentalLayoutApi
-import androidx.compose.foundation.layout.FlowRow
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
@@ -14,6 +12,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.layout.Layout
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.LinkAnnotation
 import androidx.compose.ui.text.SpanStyle
@@ -22,6 +21,7 @@ import androidx.compose.ui.text.font.FontStyle
 import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.text.withLink
 import androidx.compose.ui.text.withStyle
+import androidx.compose.ui.unit.Constraints
 import app.k9mail.core.ui.compose.common.resources.annotatedStringResource
 import kotlinx.collections.immutable.ImmutableList
 import kotlinx.collections.immutable.toImmutableList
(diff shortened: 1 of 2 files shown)


wmontwe (approved):
Thank you, works great and is tested!

PR #11522, opened 2026-09-08
chore(ci): Build app only on push to main instead of on PR updates

### LINKED ISSUE #11521: Build app only on push to main instead of on each PR update
### Checklist

### App

Thunderbird for Android

### App version

all

None

### Android version

n/a

### Device model

_No response_

### Steps to reproduce

Currently we build the app on every push to a PR to ensure the PR doesn't break the build. That results in a lot of builds. While we want to ensure patches don't break the build, we don't necessarily have to do that on every push to a PR.

The caveat is that there doesn't appear to be a good way to block a PR from landing if we only build once per PR. I'm just planning to run a single build after a PR is pushed to main, and we will at least have a history that we can review after the fact if a patch breaks the build.

I think we can keep non-beta branches building on PR updates since that is limited to uplifts where there is not a lot of activity.

### Expected behavior

One build per PR.

### Actual behavior

One build per PR update.

### Logs

_No response_

coreycb:
Fixes #11521

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -5,7 +5,18 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - '.idea/**'
+      - '.gitattributes'
+      - '.github/**.json'
+      - '.gitignore'
+      - '.gitmodules'
+      - '**.md'
+      - 'LICENSE'
+      - 'NOTICE'
   pull_request:
+    branches-ignore:
+      - main
     paths-ignore:
       - '.idea/**'
       - '.gitattributes'


wmontwe (changes requested):
This removes the only pre-merge protection against breaking `main`.   

The proposed `push` build is not an equivalent replacement. It runs only after the change has landed on `main`. At that point, CI can tell us that `main` is broken, but it cannot prevent the breakage. Every maintainer who updates their branch, starts new work, or opens another pull request can then encounter the failure until someone investigates and merges a fix or reverts the broken change.

This also transfers the burden from the author of the breaking change to the rest of the team. The failure may be discovered when the author is unavailable, and unrelated pull requests may become blocked or show failures caused by the broken base branch. That creates more disruption and investigation work than allowing CI to validate each update before merge. 

The workflow already has:
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
```yaml                                                                                                                                                                                                                                                         
    cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}                                                                                                                                                                                                    
```

so superseded runs for a pull request are cancelled.

In my opinion, reducing the number of runs does not justify removing the only pre-merge gate that assembles the applications and runs the unit tests and Android lint.

PR #11535, opened 2026-09-10
feat(settings) default push for inbox

### LINKED ISSUE #11299: Notification Settings — Default IMAP IDLE "Push" Notifications on Newly Added Accounts by Default
Users expect push-style notifications in their email apps. IMAP has had the IDLE feature, which we make use of, for years. This allows us to receive a notification when a new email is received and update the user accordingly. We currently make use of this feature to create push notifications for users on IMAP accounts. However, it's off by default and difficult to find. This has lead users to receive their notifications slowly, and become frustrated when trying to toggle the feature on. 

With this in mind, we'll enable IMAP IDLE updates, or, as users know them, "push notifications" for all IMAP accounts that support the IDLE feature. This will only change for accounts when they're added, not for existing accounts, as part of the onboarding flow. We will be making this feature more accessible in future updates, but, for now, will improve this experience for new users. 

Existing users can find this setting (subject to change) by selecting an account in the side menu if they have unified inbox turned on. Then, re-opening the side menu, selecting Manage Folders, and selecting the mailboxes they wish to turn push notifications on for. This will be improved in a future ticket and is out of scope for this one. 

AC:

- Push notifications through IMAP IDLE are enabled by default for new users who have: 1) added an account that supports IMAP's IDLE feature, and 2) have turned on push notifications and granted those permissions

dani-zilla:
Closes: #11299 
RFC / Technical Design (if applicable): [RFC 0005](https://github.com/thunderbird/thunderbird-android/blob/main/docs/engineering/rfcs/0005-imap-idle-push-on-new-account-inboxes.md)
feature-flag: `push_enabled_on_inbox_by_default`

#### Description

Enables push on the inbox by default for accounts created through the new account creation process (email and OAuth/password)

## AI Disclosure

## Contribution Checklist

### DIFF
--- a/legacy/mailstore/src/main/java/app/k9mail/legacy/mailstore/domain/GetFolderIdsForTypeUseCase.kt
+++ b/legacy/mailstore/src/main/java/app/k9mail/legacy/mailstore/domain/GetFolderIdsForTypeUseCase.kt
@@ -1,0 +1,22 @@
+package app.k9mail.legacy.mailstore.domain
+
+import app.k9mail.legacy.mailstore.MessageStoreManager
+import com.fsck.k9.mail.FolderType
+
+class GetFolderIdsForTypeUseCase(
+    private val messageStoreManager: MessageStoreManager,
+) {
+    operator fun invoke(
+        accountUuid: String,
+        folderType: FolderType,
+    ): List<Long?> {
+        return messageStoreManager.getMessageStore(accountUuid)
+            .getFolders(true) { folderDetails ->
+                if (folderDetails.type == folderType) {
+                    folderDetails.id
+                } else {
+                    null
+                }
+            }
+    }
+}
--- a/legacy/mailstore/src/main/java/app/k9mail/legacy/mailstore/domain/SetPushForFolderUseCase.kt
+++ b/legacy/mailstore/src/main/java/app/k9mail/legacy/mailstore/domain/SetPushForFolderUseCase.kt
@@ -1,0 +1,11 @@
+package app.k9mail.legacy.mailstore.domain
+
+import app.k9mail.legacy.mailstore.MessageStoreManager
+
+class SetPushForFolderUseCase(
+    private val messageStoreManager: MessageStoreManager,
+) {
+    operator fun invoke(accountUuid: String, folderId: Long, enabled: Boolean) {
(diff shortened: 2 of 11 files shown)


legacy/mailstore/src/main/java/app/k9mail/legacy/mailstore/domain/GetFolderIdsForTypeUseCase.kt
wmontwe, line 1: A UseCase should be tested.

legacy/mailstore/src/main/java/app/k9mail/legacy/mailstore/domain/SetPushForFolderUseCase.kt
wmontwe, line 1: A UseCase should be tested.

dani-zilla: The closest I have for this right now is the `SetPushForFolderUseCase should call setPushEnabled` test. Which does test that the push enabled change function gets called when the function that uses the usecase is called.

I had wondered how to do the testing for this. I had looked at the `CreateAccountTest` and had initially thought about doing it there, but the created account here is mocked, and the function being tested is mocked as well. Basically it tests that the function can take a lambda and that the lambda spits out an account, but the test is designed to basically always do that. I wasn't sure what was being tested. 

Obviously there's something more I could do here, but I'm a bit stumped on how we can write a meaningful test for this, given how mocked the account creation process and test is already. It seems like I could create a mock account, with mock folders, and mock folder settings, and make sure that the function changes that when the use case is called directly.

dani-zilla: Sorry to bump this, but do you have any suggestions for testing this? I could wrap it similar to the account creation use case, but doing so won't test the actual functionality. The other option is completely mocking or faking the data it accesses. The `SetPushForFolderUseCase should call setPushEnabled` test does ensure that it's called, but doesn't ensure it's changing the setting. I think we should probably do this with a large mockup of the user's settings and see that it's changed, but the way this is set up, that involves doing a lot of fake data object creation. I assume that's why we did the account creation use case testing the way we did. What would be the best way forward on it?

PR #11536, opened 2026-09-10
fix: distinguish “from me” and “to me” in message details

### LINKED ISSUE #7917: Grammatical error in German localization
### Checklist

### App version

6.901

F-Droid

### Android version

fp2_sibon-userdebug 1 23.02.0-rel0

### Device model

Fairphone 2

### Steps to reproduce

1. Select any mail in "Sent"

### Expected behavior

The correct German translation of "me" (indicating the mail was written by the current user) uses the dative case: "Mir" (or longer: "von mir") and not the the accusative case "Mich" that is currently used. Please adjust.

Yes, this means that it is not possible to use the same word for sender and recipient as is possible in English ("me"): (from me/to me). Sorry, but that is grammar, it's just not correct... Correct is "von mir/an mich".

So "mir" (dative) (longer: "von mir") would be correct. 

### Actual behavior

K-9 says "mich" (accusative). For received mail that would be correct ("an mich"). But not for mails written. "von mich" is really bad grammar.,,,

### Logs

_No response_

Friese2001:
## Contribution Summary

Linked Issue/Ticket: Fixes #7917
RFC / Technical Design (if applicable):

#### Description

I introduced a second string message_view_from_me_text to the english strings.xml and renamed the existing message_view_me_text to message_view_to_me_text to avoid confusion. Then I adjusted the existing logic around message_view_me_text to include both strings and display the correct values at their respective positions. For this I also added the boolean isSender. Finally I modified the relevant test to also include both strings.

Important note: This does not fully fix the reported issue. The result of these changes is that the translation for message_view_from_me_text is missing for all other languages and has to be added through Weblate after the merge. This change only introduces the required second variable so the translation can be added as different from message_view_to_me_text in all languages where the two instances of "me" are not translated as the same word.

#### Screen Shots

Screenshot_me_text

## AI Disclosure

## Contribution Checklist

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagedetails/MessageDetailsViewModel.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagedetails/MessageDetailsViewModel.kt
@@ -83,12 +83,12 @@ internal class MessageDetailsViewModel(
                 val messageDetailsUi = MessageDetailsUi(
                     date = buildDisplayDate(messageDetails.date),
                     cryptoDetails = cryptoResult?.toCryptoDetails(),
-                    from = messageDetails.from.toParticipants(account),
-                    sender = senderList.toParticipants(account),
-                    replyTo = messageDetails.replyTo.toParticipants(account),
-                    to = messageDetails.to.toParticipants(account),
-                    cc = messageDetails.cc.toParticipants(account),
-                    bcc = messageDetails.bcc.toParticipants(account),
+                    from = messageDetails.from.toParticipants(account, isSender = true),
+                    sender = senderList.toParticipants(account, isSender = true),
+                    replyTo = messageDetails.replyTo.toParticipants(account, isSender = true),
+                    to = messageDetails.to.toParticipants(account, isSender = false),
+                    cc = messageDetails.cc.toParticipants(account, isSender = false),
+                    bcc = messageDetails.bcc.toParticipants(account, isSender = false),
                     folder = folder?.toFolderInfo(),
                 )
 
@@ -125,9 +125,9 @@ internal class MessageDetailsViewModel(
         )
     }
 
-    private fun List<Address>.toParticipants(account: LegacyAccountDto): List<Participant> {
+    private fun List<Address>.toParticipants(account: LegacyAccountDto, isSender: Boolean): List<Participant> {
         return this.map { address ->
-            val displayName = participantFormatter.getDisplayName(address, account)
+            val displayName = participantFormatter.getDisplayName(address, account, isSender)
             val emailAddress = address.address
 
             Participant(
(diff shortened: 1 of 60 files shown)


legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagedetails/MessageDetailsViewModel.kt
```diff
@@ -83,12 +83,12 @@ internal class MessageDetailsViewModel(
...
                     cryptoDetails = cryptoResult?.toCryptoDetails(),
-                    from = messageDetails.from.toParticipants(account),
-                    sender = senderList.toParticipants(account),
-                    replyTo = messageDetails.replyTo.toParticipants(account),
-                    to = messageDetails.to.toParticipants(account),
-                    cc = messageDetails.cc.toParticipants(account),
-                    bcc = messageDetails.bcc.toParticipants(account),
+                    from = messageDetails.from.toParticipants(account, isSender = true),
+                    sender = senderList.toParticipants(account, isSender = true),
+                    replyTo = messageDetails.replyTo.toParticipants(account, isSender = true),
```
wmontwe, line 88: Shouldn't this be `false`, otherwise this renders currently as `Antworten von mir` instead of `Antworten an mich`

legacy/ui/legacy/src/main/res/values/strings.xml
```diff
@@ -255,7 +255,9 @@
     <!-- Displayed in the message view screen in front of the number of additional recipients (the ones that are not displayed). -->
     <string name="message_view_additional_recipient_prefix">+</string>
     <!-- Displayed in the message view screen if the message was addressed to the user. Make sure this works with 'message_view_recipient_prefix'. In English it's "to me". -->
-    <string name="message_view_me_text">me</string>
+    <string name="message_view_to_me_text">me</string>
+    <!-- Displayed in the message view screen if the message was addressed to the user. Make sure this works with 'message_view_recipient_prefix'. In English it's "to me". -->
```
wmontwe, line 259:
```suggestion
    <!-- Displayed in the message details screen when the message was sent by the user. This is shown after the "From" label. Languages may require a different grammatical form than message_view_to_me_text. In English, both are "me". -->
```

wmontwe (commented):
@Friese2001 Thank you for the fix. Could you check the reply to case and the string comment?

legacy/ui/legacy/src/main/java/com/fsck/k9/ui/messagedetails/MessageDetailsViewModel.kt
Friese2001: Hi, thank you for catching that! Yes, isSender should be false for replyTo so it renders as "Antworten an mich" and not "Antworten an mir". I'm a little unsure about the workflow at this point. Should I fix this locally and just push the changes?

wmontwe: You could either write a fix-up commit or patch the original commit and force push the changed branch.

As there is already another commit, it's better to write a fix-up commit.

wmontwe (approved):
Thank you for fixing this issue!

PR #11538, opened 2026-09-10
fix(settings): prevent HTML signature preview reloads on rotation

### LINKED ISSUE #11528: 24.0b1 - HTML signature entry on composition defaults page doesn't handle screen rotation well.
### Checklist

### App

Thunderbird for Android

### App version

24.0b1

Other

### Android version

16

### Device model

Oneplus 10 Pro

### Steps to reproduce

1. Navigate to settings > account > sending mail > composition defaults
2. Tap use signature
3. Tap save signature as HTML
4. Type some text into the HTML signature entry box
5. Rotate the screen horizontal, then vertical while focus is still within the signature entry box.

### Expected behavior

Text entered within the signature entry box is visible, with a wider view when horizontal

### Actual behavior

No text is visible. A large blank empty space is displayed above the signature position header.
Issue occurs about 80% of the time I try so may require a few attempts to reproduce.
Issue occurs with light or dark theme selected.

The same issue does NOT occur when adding a signature from the manage identifies page.

Image

Image

### Logs

_No response_

wmontwe:
## Contribution Summary

Linked Issue/Ticket: Fixes #11528 

#### Description

Prevents the HTML signature preview WebView from reloading during unrelated recompositions, which could produce an incorrect preview height after screen rotation. The preview now reloads only when its HTML content changes. Adds unit tests for the reload behavior.

## AI Disclosure

- [x] This contribution includes changes assisted by AI. -> researched the cause

### DIFF
--- a/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/setup/signature/SignatureContent.kt
+++ b/legacy/ui/legacy/src/main/java/com/fsck/k9/activity/setup/signature/SignatureContent.kt
@@ -189,6 +189,10 @@ private fun SignatureHtmlPreview(
     var contentHeight by remember { mutableStateOf(0.dp) }
     val animatedContentHeight by animateDpAsState(contentHeight)
 
+    // AndroidView's update block also runs for unrelated recompositions, e.g. height and window inset changes.
+    // Reloading the document in those cases can produce a new, incorrect content height while rotating the device.
+    val loadState = remember { SignaturePreviewLoadState() }
+
     AndroidView(
         factory = { context ->
             MessageWebView(context).apply {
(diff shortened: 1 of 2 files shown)


legacy/ui/legacy/src/main/java/com/fsck/k9/activity/setup/signature/SignatureContent.kt
```diff
@@ -189,17 +189,35 @@ private fun SignatureHtmlPreview(
     var contentHeight by remember { mutableStateOf(0.dp) }
     val animatedContentHeight by animateDpAsState(contentHeight)
 
+    // AndroidView's update block also runs for unrelated recompositions, e.g. height and window inset changes.
+    // Reloading the document in those cases can produce a new, incorrect content height while rotating the device.
+    val loadState = remember { SignaturePreviewLoadState() }
```
rafaeltonholo, line 194: Any reasons for not using `mutableStateOf` here?

```suggestion
    val loadState = remember { mutableStateOf(SignaturePreviewLoadState()) }
```

wmontwe: `loadedSignature` only tracks which HTML was last sent to the current WebView. The Compose UI doesn’t depend on it, so changes shouldn’t trigger recomposition. It only needs to survive recompositions, which `remember` provides. If the WebView is recreated, the state should reset so the new WebView loads the HTML once.

rafaeltonholo (approved):
LGTM

Issue #9232, opened 2025-05-28
Replace direct use of K9 class


shamim-emon:
Goal of this  issue is to replace all direct usage of K9 class with appropriate implementations of `PreferenceDataStore `

PR #11556, opened 2026-09-14
feat(donation-appeal): added preferences and accessors for donation appeal

### LINKED ISSUE #11546: Donation Appeal — New Database Entries
The new donation appeal will not replace the existing one, but will instead display a set amount of time after the initial introduction donation appeal has been displayed. That means we have to track the time that it has been displayed separately to ensure that we're only showing it once and only after the other one has displayed. 

AC: 
- Add a database entry for the new funding reminder shown timestamp, likely `secondaryFundingReminderShownTimestamp`, or something less verbose
- Add all accessors and modifiers to the `FundingSettings` classes that use this
- Ensure we can read/write the value to shared preferences

dani-zilla:
## Contribution Summary

Closes: #11546 

#### Description

Adds the preferences and accessors for `lastFundingReminderShownTimestamp` and `fundingReminderCount`

This will be part of future work to change the logic of how we display donation appeals, including the new one we're adding as well as the potential to add new ones in the future without as much effort. 

These won't replace the existing logic, only add to it and support the new logic for the new funding appeals display

## AI Disclosure

## Contribution Checklist

- [X] This contribution includes tests for any new functionality, and maintains tests for any updated functionality (new functionality added in future ticket)


rafaeltonholo (changes requested):
Thank you for your contribution! Although the current changes follow the same architecture as the previous implementation used for funding reminders, we will no longer be using that approach.

We should avoid adding new properties to the `K9` object, as this is not a good practice. 

The funding feature (in the `:feature:funding:googleplay` module) is already utilizing the `GeneralSettings` replacement, `ConfigStore`, so let's adopt that method instead of continuing with the `GeneralSettings` approach.

You can find instructions on how to create a `ConfigStore` in the following locations:
- `feature/funding/googleplay/src/main/kotlin/net/thunderbird/feature/funding/googleplay/data/local/configstore`
- `core/featureflag/src/commonMain/kotlin/net/thunderbird/core/featureflag/data/configstore`

You do not need to migrate the other three funding preferences to use `ConfigStore`, but feel free to do so if you are comfortable with it.

Additionally, if these changes resolve the issue, please make sure to use the auto-closing keywords when mentioning the issue number in the PR description.

Please let me know if you need any assistance with this matter.

dani-zilla:
You know, it's always hard to tell what's the "we just do it this way" bad practice and the "we have to do it a different way next time" bad practice. This is, actually, my version of the latter as well, as I'd prefer to build out a larger campaign system, but we're under some tight deadlines with this one. Can look into moving this today though. Thanks.

wmontwe:
We should avoid adding new code to the legacy modules, this is only acceptable when there is no other way or the cost is very high. The K9 class is subject to being replaced and no new code should be added. See #9232 for reference.

app-common/src/main/kotlin/net/thunderbird/app/common/feature/funding/configstore/FundingConfigStore.kt
```diff
@@ -0,0 +1,36 @@
...
+import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.runBlocking
+import net.thunderbird.core.configstore.BaseConfigStore
+import net.thunderbird.core.configstore.ConfigId
+import net.thunderbird.core.configstore.ConfigStore
+import net.thunderbird.core.configstore.backend.ConfigBackendProvider
+
+interface FundingConfigStore : ConfigStore<FundingConfig> {
+    fun dataStateFlow(): StateFlow<FundingConfig>
```
rafaeltonholo: nit: I would name it as:

```suggestion
    fun configAsStateFlow(): StateFlow<FundingConfig>
```

rafaeltonholo (dismissed):
LGTM! I would just move `FundingConfig`, `FundingConfigDefinition`, `FundingConfigKeys` and the `FundingConfigStore` interface to the `:feature:funding:api` module since they are api definitions. However, I won't block this PR because of that. You can address that in a follow-up PR if you prefer.

Also, make sure to address spotless issues.

Thanks for this contribution!

rafaeltonholo (approved):
LGTM

PR #11571, opened 2026-09-16
fix(pii-safe): detect PII in nested type arguments

### LINKED ISSUE #11570: [BUG] Pii-Safe Compiler Plugin — FIR checker not considering Type Arguments annotated with `@PiiSafe.HasPii`
## Description

When a data class is **NOT annotated** with `@PiiSafe.HasPii`, the FIR checker `HasPiiRequiresAnnotatedPropertyChecker` verifies if any of its properties are annotated with `@PiiSafe.HasPii`. However, it is missing checking the Type Arguments as well, leaving the following scenario prone to leaking PII:

```
@PiiSafe.HasPii
data class MessageAddress(
    @get:PiiSafe.Mask
    val value: String,
    val label: String,
)

data class MessageEnvelope(
    val subject: String,
    val from: List<MessageAddress>,
    ...
)
```

## Expected behaviour
1. Build should fail with the following message:
    ```
     > Task :feature:mail:message:api:compileKotlinJvm FAILED
     e: file:///<file-path>/MessageEnvelope.kt:22:5 This class holds a property of a type annotated @PiiSafe.HasPii; this class must also be annotated @PiiSafe.HasPii.
    ```
2. (In case of enabled) IDE should mark the properties whose type argument is annotated with `@PiiSafe.HasPii` as an error, showing the above-mentioned message

## Actual behaviour
1. The build succeeds
2. No `toString` method is overridden if no other property type is annotated with `@PiiSafe.HasPii`
3. (In case of enabled) No IDE error

Build output:
```
> BUILD SUCCESSFUL
```

## Instructions

* Although this is labelled as `good first issue`, it involves a complex topic with not much documentation. You may need to check the [Kotlin Compiler source code](https://github.com/JetBrains/kotlin/tree/v2.4.20/compiler) to succeed in this task.
* This is a bug in the newly introduced Kotlin Compiler Plugin, located at library/pii-safe/compiler-plugin
* You will need to change the logic inside the `HasPiiRequiresAnnotatedPropertyChecker.check` method to look after type arguments.
* You MUST implement test cases to handle this scenario inside `HasPiiRequiresAnnotatedPropertyCheckerTest`

taljeon:
## Contribution Summary

Linked Issue/Ticket: Fixes #11570
RFC / Technical Design (if applicable): N/A — focused compiler-plugin bug fix

#### Description

- Recursively inspect resolved generic type arguments in `HasPiiPropagationChecker`, so an unannotated data class cannot hide a PII-bearing type inside nested containers.
- Keep `HasPiiRequiresAnnotatedPropertyChecker` strict: a `@HasPii` data class is valid only when a property is directly `@HasPii` or explicitly uses `@Mask`/`@Hide`; a nested PII type alone is insufficient because the outer generic's `toString()` may leak it.
- Add regressions for both boundaries: nested `Map<String, List<MessageAddress>>` propagation and an unmasked `LeakyEnvelope<MessageAddress>` property.

Before the initial implementation, the two nested-type regressions ran with the six existing focused tests and both failed. A review regression then demonstrated that the shared recursive predicate allowed an unmasked generic property to compile (1 of 4 annotation-required tests failed before the correction). After separating the policies, all eight focused tests pass.

#### Screen Shots

N/A — no UI changes.

## AI Disclosure

OpenAI Codex generated the implementation, tests, and pull request text. No manual human code review was performed before submission; the exact automated checks run are listed below.

## Verification

Passed:

- `./gradlew :library:pii-safe:compiler-plugin:test --tests 'net.thunderbird.piisafe.compiler.fir.checkers.HasPiiPropagationCheckerTest' --tests 'net.thunderbird.piisafe.compiler.fir.checkers.HasPiiRequiresAnnotatedPropertyCheckerTest'` — 8 tests
- `./gradlew :library:pii-safe:compiler-plugin:check` — full module tests (41), Detekt, and Spotless
- `./gradlew detekt spotlessCheck` — repository-wide
- `git diff --check`
- isolated changed-file and PR-body leak checks

Not run locally:

- `./gradlew test lint detekt spotlessCheck` could not reach `test` or `lint` because this host has no Android SDK (`ANDROID_HOME` / `sdk.dir`). Repository-wide `detekt` and `spotlessCheck` were rerun separately and passed.
- `connectedAndroidTest` was not run because no Android SDK/device is available; the changed module is a JVM compiler plugin and its focused/full module tests passed.

## Contribution Checklist

- [ ] This contribution does not break existing unit tests (run `gradlew testDebugUnitTest`; will be checked by CI). The relevant JVM module suite passes; the repository-wide Android task is blocked locally by the missing SDK as documented above.


taljeon:
Could a maintainer apply `report: exclude`? This is an internal compiler-plugin correctness fix rather than a user-facing change. I attempted to set the required label, but contributor permissions do not allow it.

wmontwe (changes requested):
Treating a nested `@HasPii` type argument as sufficient allows an unmasked generic property to bypass the existing safety check.                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                     
The generated `toString()` delegates to the outer generic type’s `toString()`, which is not guaranteed to preserve masking:                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                     
 ```kotlin                                                                                                                                                                                                                                                                                                                                                                           
   @PiiSafe.HasPii                                                                                                                                                                                                                                                                                                                                                                   
   data class MessageAddress(@get:PiiSafe.Mask val value: String)                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                     
   class LeakyEnvelope<T>(val value: T) {                                                                                                                                                                                                                                                                                                                                            
       override fun toString() = (value as MessageAddress).value                                                                                                                                                                                                                                                                                                                     
   }                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                     
   @PiiSafe.HasPii                                                                                                                                                                                                                                                                                                                                                                   
   data class MessageEnvelope(val secret: LeakyEnvelope<MessageAddress>)                                                                                                                                                                                                                                                                                                             
 ```                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                     
Before this change, `MessageEnvelope` fails compilation because secret needs `@Mask` or `@Hide`. With this change, it compiles and produces:                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                     
 ```text                                                                                                                                                                                                                                                                                                                                                                             
   MessageEnvelope(secret = private)                                                                                                                                                                                                                                                                                                                                                 
 ```                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                     
Could we keep recursive type-argument detection for `HasPiiPropagationChecker`, but only allow a directly `@HasPii` property OR an explicitly `@Mask/@Hide` property—to satisfy `HasPiiRequiresAnnotatedPropertyChecker`?

taljeon:
Thanks — you're right. Recursive type-argument detection now remains only in `HasPiiPropagationChecker`.

`HasPiiRequiresAnnotatedPropertyChecker` now resolves only the property's outer return-type symbol, so it accepts a directly `@HasPii` property or an explicit `@Mask`/`@Hide`, but not an unmasked generic that merely contains a PII-bearing type.

I added the `LeakyEnvelope<MessageAddress>` case from the review as the regression. It compiled before the correction (1 of the 4 annotation-required tests failed against the expected diagnostic) and now fails compilation with `HAS_PII_CLASS_WITHOUT_ANNOTATED_PROPERTY`. The propagation check for nested `Map<String, List<MessageAddress>>` still passes.

Verification on the refreshed branch:

- both checker suites: 8/8 tests pass;
- `:library:pii-safe:compiler-plugin:check`: full module tests, Detekt, and Spotless pass;
- repository-wide `detekt spotlessCheck`: pass.

wmontwe (approved):
Thank you for the fix!

PR #11583, opened 2026-09-18
Draft: feat(donation-appeal): Create the logic to display the donation appeal

### LINKED ISSUE #11547: Donation Appeal — New Donation Display Logic
The new donation appeal popup should display after 30 minutes of in-app usage after the first one has displayed. 
We'll need new logic to define how this works. We'll have to ensure that:

- The first popup has already displayed
- The user has used the app for 30 minutes after it has shown

We can use the existing popup UI for now to build and test this feature. 

This should be registered with the current donation popup so we can track it alongside it. 

We will also use the newly created secondary donation appeal shown shared preference entry to track if/when this has been displayed

AC:
- New display logic has been created
- Existing popup UI will display after 30 minutes of active usage after the first one has displayed
-  A future ticket will create the new UI, this is out of scope for this ticket. We're using the current UI just to create the logic layer while the new UI is being finalized. This will also make testing easier
- Create new tests to ensure we trigger this second alert at the right time.

dani-zilla:
## Contribution Summary

Closes: #11547  

#### Description

This introduces the logic and tests that define when we display a second donation appeal after the first one. 

It currently uses the same fragment and UI as the original, as the UI will be completed in another ticket. 

This also introduces tests for the second funding reminder, as well as adding new tests for displaying the first funding reminder. We don't want these to overlap. 

#### Screen Shots

<Will be provided when fragment launch issue is resolved>

## AI Disclosure

## Contribution Checklist

- [X] This contribution includes tests for any new functionality, and maintains tests for any updated functionality.


dani-zilla:
Fixed two issues, but still seems to fail to launch the donation appeal fragment, despite passing all the checks. Looking into this currently as well

dani-zilla:
There's an issue in main that this PR needs to address. While calling the initialization in the `postResume()` block allows the first donation appeal to display, we run into a problem later. When it's called multiple times, it resets the activity counter. This keeps it from being able to register the second appeal. 

We need to ensure we're only calling the lifecycle observer for this once. One way to do that will be moving it entirely to the `MessageHomeController` class, and having it only pass in the elapsed time to the funding initialization. This way, as it's recreated, it won't duplicate this.

Currently, this is the only reason the second reminder does not display, and we can fix it once we fix this.

dani-zilla:
This also now includes the fix for observing the activity lifecycle for the time in activity counter by only doing it once during initialization of the funding reminder. Testing it should work as described above, you can go to settings to cause the activity to re-load and test, then navigate away and back to the message list to show the notification. You should be able to repeat this twice. Adjust the interval in the FundingReminderContract to 0 to ensure you can see the popup immediately.

app-thunderbird/src/main/kotlin/net/thunderbird/android/feature/TbFundingSettings.kt
rafaeltonholo: **Unrelated to this change, and should be addressed in a follow-up ticket:** Both `K9FundingSettings` and `TbFundingSettings` have the exact same content and could be exported to a common implementation inside the `app-common`.

feature/funding/api/src/commonMain/kotlin/net/thunderbird/feature/funding/api/FundingConfigStore.kt
```diff
@@ -4,5 +4,6 @@ import kotlinx.coroutines.flow.StateFlow
 import net.thunderbird.core.configstore.ConfigStore
 
 interface FundingConfigStore : ConfigStore<FundingConfig> {
-    fun configAsStateFlow(): StateFlow<FundingConfig>
+    val configStateFlow: StateFlow<FundingConfig>
+    suspend fun ConfigStore<FundingConfig>.update(transform: (FundingConfig) -> FundingConfig)
```
rafaeltonholo: As the parent has the exact same function signature, but with a nullable parameter for the lambda parameter, Kotlin will never resolve to this one. I would suggest using a different name, such as:

```suggestion
    suspend fun ConfigStore<FundingConfig>.safeUpdate(transform: (FundingConfig) -> FundingConfig)
```

app-thunderbird/src/main/kotlin/net/thunderbird/android/feature/TbFundingSettings.kt
```diff
@@ -25,32 +22,38 @@ internal class TbFundingSettings(
...
-                val oldConfig = it ?: FundingConfig.DEFAULT
-                oldConfig.copy(fundingReminderCount = count)
-            }
+    override suspend fun setReminderShownCount(count: Int) {
+        fundingConfigStore.update {
+            val oldConfig = it ?: FundingConfig.DEFAULT
+            oldConfig.copy(
+                lastFundingReminderShownActivityAmount = oldConfig.lastFundingReminderShownActivityAmount,
+                fundingReminderCount = count,
+            )
```
rafaeltonholo: The `lastFundingReminderShownActivityAmount = oldConfig.lastFundingReminderShownActivityAmount` is redundant here

```suggestion
            oldConfig.copy(fundingReminderCount = count)
```

app-thunderbird/src/main/kotlin/net/thunderbird/android/feature/TbFundingSettings.kt
```diff
@@ -25,32 +22,38 @@ internal class TbFundingSettings(
...
-                val oldConfig = it ?: FundingConfig.DEFAULT
-                oldConfig.copy(lastFundingReminderShownTimestamp = timestamp)
-            }
+    override suspend fun setLastReminderShownActivityAmount(activityInMillis: Long) {
+        fundingConfigStore.update {
+            val oldConfig = it ?: FundingConfig.DEFAULT
+            oldConfig.copy(
+                lastFundingReminderShownActivityAmount = activityInMillis,
+                fundingReminderCount = oldConfig.fundingReminderCount,
+            )
```
rafaeltonholo: The `fundingReminderCount = oldConfig.fundingReminderCount,` is redundant here

```suggestion
            oldConfig.copy(
                lastFundingReminderShownActivityAmount = activityInMillis,
            )
```

feature/funding/googleplay/src/main/kotlin/net/thunderbird/feature/funding/googleplay/ui/reminder/FundingReminder.kt
```diff
@@ -48,27 +65,31 @@ constructor(
...
             fragmentObserver.register(observedFragmentManager) {
                 showFundingReminderDialog(dialogFragmentManager)
             }
+        } else if (shouldShowSecondReminder()) {
+            // TODO make this point to the new dialog: #11526
+            fragmentObserver.register(observedFragmentManager) {
+                showSecondFundingReminderDialog(dialogFragmentManager)
+            }
         }
     }
```
rafaeltonholo: nit: I would use a when instead of a if/else-if here. This can grow in case we have a third, forth reminders

```suggestion
        when {
            shouldShowReminder() -> fragmentObserver.register(observedFragmentManager) {
                showFundingReminderDialog(dialogFragmentManager)
            }
           shouldShowSecondReminder() -> fragmentObserver.register(observedFragmentManager) {
                // TODO make this point to the new dialog: #11526
                showSecondFundingReminderDialog(dialogFragmentManager)
            }
        }
```

feature/funding/googleplay/src/main/kotlin/net/thunderbird/feature/funding/googleplay/ui/reminder/FundingReminder.kt
```diff
@@ -94,9 +131,27 @@ constructor(
     private fun showFundingReminderDialog(fragmentManager: FragmentManager) {
         // We're about to show the funding reminder dialog. So mark it as being shown. This way, if there's an error,
         // we err on the side of the dialog not being shown rather than it being shown more than once.
-        @OptIn(ExperimentalTime::class)
-        settings.setReminderShownTimestamp(clock.now().toEpochMilliseconds())
+        scope.launch {
+            @OptIn(ExperimentalTime::class)
```
rafaeltonholo: This isn't necessary anymore. Clock is no longer experimental

feature/funding/googleplay/src/main/kotlin/net/thunderbird/feature/funding/googleplay/ui/reminder/FundingReminder.kt
```diff
@@ -94,9 +131,27 @@ constructor(
...
+        }
+
+        dialog.show(fragmentManager)
+    }
 
+    private fun showSecondFundingReminderDialog(fragmentManager: FragmentManager) {
+        // TODO: This implementation is currently the same as showFundingReminderDialog(),
+        //  but will differ after the new UI and logic to block the first reminder for new users is introduced
+        scope.launch {
+            @OptIn(ExperimentalTime::class)
```
rafaeltonholo: This isn't necessary anymore. Clock is no longer experimental

feature/funding/googleplay/src/main/kotlin/net/thunderbird/feature/funding/googleplay/ui/reminder/FundingReminder.kt
```diff
@@ -94,9 +131,27 @@ constructor(
...
+            settings.setReminderShownTimestamp(now)
+            settings.setLastReminderShownActivityAmount(settings.getActivityCounterInMillis())
+            settings.incrementReminderShownCount()
+        }
+
+        dialog.show(fragmentManager)
+    }
 
+    private fun showSecondFundingReminderDialog(fragmentManager: FragmentManager) {
+        // TODO: This implementation is currently the same as showFundingReminderDialog(),
```
rafaeltonholo: Please, add the ticket number to the TODO keyword, so we don't let this slip.

BTW, thanks for the comment, I was about to add a comment saying it is the same implementation, lol

feature/funding/googleplay/src/test/kotlin/net/thunderbird/feature/funding/googleplay/ui/reminder/FundingReminderTest.kt
```diff
@@ -107,6 +113,81 @@ class FundingReminderTest {
...
+            scope = CoroutineScope(mainDispatcher.testDispatcher),
+        )
+
+        testSubject.registerReminder { }
+
+        assertFalse(dialogShown)
+        assertFalse(fragmentObserver.isRegistered)
+        assertThat(settings.getReminderShownTimestamp()).isEqualTo(LAST_REMINDER_SHOWN_UNSET)
+        assertEquals(REMINDER_COUNTER_FIRST_SHOWN, settings.getReminderShownCount())
+    }
```
rafaeltonholo: You should use assertk instead.

```suggestion
        assertThat(dialogShown).isFalse()
        assertThat(fragmentObserver.isRegistered).isFalse()
        assertThat(settings.getReminderShownTimestamp()).isEqualTo(LAST_REMINDER_SHOWN_UNSET)
        assertThat(settings.getReminderShownCount()).isEqualTo(REMINDER_COUNTER_FIRST_SHOWN)
```

feature/funding/googleplay/src/test/kotlin/net/thunderbird/feature/funding/googleplay/ui/reminder/FundingReminderTest.kt
```diff
@@ -107,6 +113,81 @@ class FundingReminderTest {
...
+            scope = CoroutineScope(mainDispatcher.testDispatcher),
+        )
+
+        testSubject.registerReminder { }
+
+        assertFalse(dialogShown)
+        assertFalse(fragmentObserver.isRegistered)
+        assertThat(settings.getReminderShownTimestamp()).isEqualTo(REMINDER_SHOWN_TIMESTAMP_UNSET)
+        assertThat(settings.getLastReminderShownActivityAmount()).isEqualTo(REMINDER_REFERENCE_TIMESTAMP)
+        assertEquals(0, settings.getReminderShownCount())
```
rafaeltonholo: Please, use assertk instead

legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageHomeActivity.kt
rafaeltonholo: Please, revert the changes for this file, since at the end it is the same content

app-thunderbird/src/main/kotlin/net/thunderbird/android/feature/TbFundingSettings.kt
dani-zilla: Agreed. I didn't want to further complicate this any more as it's only getting more difficult to blend the legacy and current code, but this was annoying. I'm also going to eventually be combining a number of them, including the google play and link views, as we'll eventually expand this to foss builds too. This definitely is in need of a larger refactor.

app-thunderbird/src/main/kotlin/net/thunderbird/android/feature/TbFundingSettings.kt
dani-zilla: Ah, yeah, good catch. Was an attempted patch while I was having concurrency issues.

feature/funding/googleplay/src/main/kotlin/net/thunderbird/feature/funding/googleplay/ui/reminder/FundingReminder.kt
dani-zilla: And, while I was hoping we could get a refactor before then, I don't think we will, so, good catch

legacy/ui/legacy/src/main/java/com/fsck/k9/activity/MessageHomeActivity.kt
dani-zilla: Yeah, Wolf took it for the patch, but I had moved it down below onResume before so it would be less out of order if you're reading through the code. Just felt like it might make more sense there organizationally. I didn't realize he didn't do the same when applying it to the patch. What do you think?

rafaeltonholo: All good, makes sense

rafaeltonholo (dismissed):
LGTM

rafaeltonholo (approved):
LGTM

PR #11590, opened 2026-09-21
uplift(beta): fix(funding): avoid reminder activity race

wmontwe:
## Summary

Uplifts #11589 to `beta`.

The funding reminder was registered during `MessageHomeActivity.onCreate()`. This could race with `ActivityProvider` lifecycle updates, causing the reminder to register with the wrong activity or not register at all. As a result, eligible users might not see the donation appeal.

## Risk

Low. This is a small lifecycle timing change affecting only funding reminder registration.


wmontwe:
Thank you for your uplift request! Please add a comment with the following approval request template filled out.

**[Approval Request]**
Original Issue/Pull request: #11589                                                                                                                                                                                                                                                                                                                                                 
 Regression caused by (issue #): #10660                                                                                                                                                                                                                                                                                                                                              
 User impact if declined: Eligible users may not see the donation reminder because it can register with the wrong activity or fail to register.                                                                                                                                                                                                                                      
 Testing completed (on daily, beta, etc.): main                                                                                                                                                                                          
 Introduces or modifies localizable strings (yes/no): No                                                                                                                                                                                                                                                                                                                             
 Skips a database migration (yes/no): No                                                                                                                                                                                                                                                                                                                                             
 Risk to taking this patch (and alternatives if risky): Low, it only moves registration to a safer lifecycle callback.

rafaeltonholo (approved):
LGTM

rafaeltonholo (changes requested):
After testing again, we noticed that the listener is getting registered multiple times, but not being unregistered, causing a waste of resources.

wmontwe:
> After testing again, we noticed that the listener is getting registered multiple times, but not being unregistered, causing a waste of resources.

Yeah, this moved it from `onCreate()` to `onPostResume` which is called multiple times and doesn't check registration.

PR #11591, opened 2026-09-21
uplift(release): fix(funding): avoid reminder activity race

wmontwe:
## Summary

Uplifts #11589 to `beta`.

The funding reminder was registered during `MessageHomeActivity.onCreate()`. This could race with `ActivityProvider` lifecycle updates, causing the reminder to register with the wrong activity or not register at all. As a result, eligible users might not see the donation appeal.

## Risk

Low. This is a small lifecycle timing change affecting only funding reminder registration.


dani-zilla (approved):

dani-zilla:
Same fix I applied locally in my branch for the new donation appeal features. Have manually tested it and it works 👍 Good to merge once build completes.

dani-zilla:
I've done some further testing on this and I'm not sure this is anything but a quick fix. Can you try testing it for something like 5 minutes? During that time, go out of the message activity and come back in. You can do this by going into settings. I believe the observer is being reset when the `MessageHomeActivity` resumes.

wmontwe:
> I've done some further testing on this and I'm not sure this is anything but a quick fix. Can you try testing it for something like 5 minutes? During that time, go out of the message activity and come back in. You can do this by going into settings. I believe the observer is being reset when the `MessageHomeActivity` resumes.

I'm not sure why the value resets because there's no reset logic implemented. The observer simply adds elapsed time to the value loaded at startup. Since K9 saves asynchronously, there might be an issue, if you see it reset to 0 or an old value.

We also have a bug with this patch that switches to `onPostResume`, resulting in observers are being registered multiple times. This doesn't reset the count, but it causes the time to inflate, for instance, two observers adding 2 minutes to a 5-minute baseline results in 9 minutes total.

